PR #6, opened 2022-03-30
Restructure library, bugfixes and extension of various functionality

hikinggrass:
- Add DataTransfer callbacks
- Disable building of example binaries by default
- Add set max current callback
- Fix DateTime parsing on some central systems
- Add a part of the SmartCharging profile
- Restructure header only code into hpp and cpp files
- Enable compiler warnings
- Various bugfixes


chGoodchild (approved):
Looks like a lot of header-only restructuring :)

PR #8, opened 2022-04-25
Add some missing includes

hikinggrass:
Signed-off-by: Kai-Uwe Hermann <kai-uwe.hermann@pionix.de>

### DIFF
--- a/lib/ocpp1_6/charge_point_state_machine.cpp
+++ b/lib/ocpp1_6/charge_point_state_machine.cpp
@@ -1,6 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
 #include <ocpp1_6/charge_point_state_machine.hpp>
+#include <ocpp1_6/enums.hpp>
+
+#include <stddef.h>
+#include <cstdint>
+#include <stdexcept>
+#include <utility>
+
+#include <fsm/async.hpp>
 
 namespace ocpp1_6 {
 ChargePointStateMachine::ChargePointStateMachine(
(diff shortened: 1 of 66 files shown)


lib/ocpp1_6/charge_point_state_machine.cpp
```diff
@@ -1,6 +1,14 @@
...
 // Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
 #include <ocpp1_6/charge_point_state_machine.hpp>
+#include <ocpp1_6/enums.hpp>
+
+#include <stddef.h>
+#include <cstdint>
+#include <stdexcept>
+#include <utility>
+
+#include <fsm/async.hpp>
```
a-w50, line 11: why does this for example gets included here again, if it is already included in the charge_point_state_machine.hpp?

src/code_generator/generate_cpp.py
```diff
@@ -345,6 +369,51 @@ def parse_schemas(version: str, schema_dir: Path = Path('schemas/json/'),
...
+        message_needs_types = False
+
+        for (type_name, type_key) in (('Request', 'req'), ('Response', 'res')):
+            parsed_types.clear()
+            parsed_enums.clear()
+
+            # FIXME!
+            current_defs.clear()
+            current_defs.update(
+                type_of_action[type_key].get('definitions', {}))
```
a-w50, line 384: This should really be fixed some day :)

lib/ocpp1_6/charge_point_state_machine.cpp
hikinggrass: Good question, I tried to apply some suggestions from include-what-you-use but might have gone a little overboard with some of them :D

a-w50 (approved):

PR #31, opened 2022-07-08
Implemented missing error code for fault state

Pietfried:
When changing to state "Faulted" it was necessary to send a payload with the event to describe the error code.

### DIFF
--- a/include/ocpp1_6/charge_point.hpp
+++ b/include/ocpp1_6/charge_point.hpp
@@ -328,7 +328,7 @@ public:
 
     /// EV/EVSE indicates that an error with the given \p error_code occured
     /// \returns true if this state change was possible
-    bool error(int32_t connector, ChargePointErrorCode error_code);
+    bool error(int32_t connector, const ChargePointErrorCode& error);
 
     /// EV/EVSE indicates that a vendor specific error with the given \p vendor_error_code occured on the given \p
     /// connector \returns true if this state change was possible
--- a/include/ocpp1_6/charge_point_state_machine.hpp
+++ b/include/ocpp1_6/charge_point_state_machine.hpp
@@ -21,8 +21,7 @@
 
 namespace ocpp1_6 {
 
-enum class ChargePointStatusTransition
-{
+enum class ChargePointStatusTransition {
     BecomeAvailable,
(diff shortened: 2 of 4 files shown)


include/ocpp1_6/charge_point.hpp
```diff
@@ -328,7 +328,7 @@ class ChargePoint {
 
     /// EV/EVSE indicates that an error with the given \p error_code occured
     /// \returns true if this state change was possible
-    bool error(int32_t connector, ChargePointErrorCode error_code);
+    bool error(int32_t connector, json error_code);
```
a-w50, line 331: Why did the type of error_code change here? I would favor the more specific type of ChargePointErrorCode

a-w50 (commented):
Could you please squash the commits (interactive rebase) and be more specific in the commit message ("Extended StateMachine" is quite general ...)

include/ocpp1_6/charge_point.hpp
Pietfried: We receive different error codes from the evse manager. At some time we have to unify or map them

a-w50: so the error_code function is a handler thats called from the evse manager?

Pietfried: its called from the OCPP module in the everest-core as a callback when the evse manager sends an error

a-w50: I see, but I would argue, that the one, who is calling this error function should do the conversion of json to ChargePointErrorCode.  The libocpp should not be in charge of knowing about evse manager naming of variables.  So propbably on could do this conversion inside the ocpp module of everest (that would be the right place of conversion).  For example. libocpp should have absolutely no knowlegde about "SLAC"

Pietfried: yes definitely, ill keep this PR open untils this + the clearance of the error is fixed

a-w50: ok

Pietfried: fixed it - should be merged together with https://github.com/EVerest/everest-core/pull/73

PR #38, opened 2022-08-01
Changed Arguments of GetDiagnostics.req and UpdateFirmware.req callbacks

Pietfried:
The argument for the callbacks for GetDiagnostics.req and UpdateFirmware.req has been changed from the single (ftp) location to the whole ocpp message, so that all members of this message can be considered in the callback function.

Signed-off-by: pietfried <piet.goempel@pionix.de>


a-w50 (approved):
You could do something like
using UpdateFirmwareCallbackType = std::function<void( ....)>
and then
UpdateFirmwareCallbackType update_firmware_callback;
and
void register_update_firmware_callback(const UpdateFirmwareCallbackType& callback)
etc

Pietfried:
Was integrated in https://github.com/EVerest/libocpp/tree/pg-evse-event-auth-refactoring and will be part of PR of this branch

PR #49, opened 2022-10-11
Access Smart Charging profiles

Pietfried:
Smart charging profiles of ChargePoint can now be calculated and requested externally.


Pietfried:
Should be merged with https://github.com/EVerest/everest-core/pull/125

PR #65, opened 2022-12-12
Rework of OCPP SmartCharging

Pietfried:
- Reworked handling for SetChargingProfile.req ClearChargingProfile.req and GetCompositeSchedule.req
- Added class SmartChargingHandler which is responsible for maintaining profiles and calculating the composite schedule 
- Supporting all ChargingProfilePurposes and ChargingProfileKinds
- Storing charging profiles persistently - added table CHARGING_PROFILES to database - ChargingProfiles are loaded on boot

Signed-off-by: pietfried <piet.goempel@pionix.de>


include/ocpp1_6/smart_charging.hpp
```diff
@@ -0,0 +1,134 @@
...
+#include <ocpp1_6/transaction.hpp>
+
+namespace ocpp1_6 {
+
+static const int LOW_VOLTAGE = 230;
+static const int DEFAULT_AND_MAX_NUMBER_PHASES = 3;
+static const int HOURS_PER_DAY = 24;
+static const int SECONDS_PER_HOUR = 3600;
+static const int SECONDS_PER_DAY = 86400;
+static const int DAYS_PER_WEEK = 7;
```
a-w50, line 21: Using `static` here is not a good idea in general, because it will create an instance in every compilation unit (cpp file), that is including this header file.

hikinggrass (approved):

PR #70, opened 2023-01-31
added dc reporting for get_latest_meter_value 

Afronut:
get_latest_meter_value value does not report DC  value for Voltage and current


Pietfried (approved):
Thanks for adding this!

PR #71, opened 2023-02-03
Basic Implementation of Functional Block: Provisioning

Pietfried:
- moved ChargePointConnectionState back to 1.6
- improved ws connection and bootnotification handling
- implemented functional block Provisioning profile as draft (B01 - B08)
- renamed common ChargePoint to ChargingStationBase
- renamed ChargePointConfiguration to DeviceModelManager for OCPP2.0.1
- added method documentation
- implemented draft for state machine within connector class instead of using libfsm

Signed-off-by: pietfried <piet.goempel@pionix.de>

PR #72, opened 2023-02-16
Session Logging

Pietfried:
Extended Message Logging and API so that sessions can be logged in seperate files. This requires the parameter log_path within the on_session_started and on_session_stopped handlers and the session logging has to be enabled within the MessageLogFormats.

The session logging simply logs all ocpp messages that occur within the two handlers. So there is no filtering for messages that only belong to one or the other session. Everything is captured within these two events.

### DIFF
--- a/include/ocpp/v16/charge_point.hpp
+++ b/include/ocpp/v16/charge_point.hpp
@@ -295,12 +295,14 @@ public:
     void on_max_current_offered(int32_t connector, int32_t max_current);
 
     /// \brief Notifies chargepoint that a new session with the given \p session_id has been started at the given \p
-    /// connector with the given \p reason
-    void on_session_started(int32_t connector, const std::string& session_id, const std::string& reason);
+    /// connector with the given \p reason . The logs of the session will be written into \p session_logging_path if
+    /// present
+    void on_session_started(int32_t connector, const std::string& session_id, const std::string& reason,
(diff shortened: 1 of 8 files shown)


include/ocpp/v16/charge_point.hpp
```diff
@@ -295,12 +295,14 @@ class ChargePoint : ocpp::ChargePoint {
     void on_max_current_offered(int32_t connector, int32_t max_current);
 
     /// \brief Notifies chargepoint that a new session with the given \p session_id has been started at the given \p
-    /// connector with the given \p reason
-    void on_session_started(int32_t connector, const std::string& session_id, const std::string& reason);
+    /// connector with the given \p reason . The logs of the session will be written into \p session_logging_path if
+    /// present
+    void on_session_started(int32_t connector, const std::string& session_id, const std::string& reason,
```
hikinggrass, line 300: I think these API changes necessitate a version increase and the 1.6 OCPP module in everest-core needs to be updated as well

hikinggrass (changes requested):

include/ocpp/v16/charge_point.hpp
Pietfried: I added it. Propably needs to change in this PR: https://github.com/EVerest/libocpp/pull/71 depending on what we merge first

hikinggrass (approved):

PR #74, opened 2023-02-25
fix error and update example

Afronut:
- in database_handler.cpp the boost optional value expiryDate will not be initialized on idTag if experyDate is not provided or tag  does not have such a date. However, in the chargepoint->validate_against_cache_entries, value() is called on the returned idtag (a boost optional ) from  get_authorization_cache_entry. This will always throw.
- Update the chargepoint example because parameters of he function on_session_started and on_session_stopped has changed

### DIFF
--- a/src/charge_point.cpp
+++ b/src/charge_point.cpp
@@ -234,7 +234,7 @@ int main(int argc, char* argv[]) {
             } else if (command == "start_transaction") {
                 if (!transaction_running) {
                     uuid = boost::lexical_cast<std::string>(boost::uuids::random_generator()());
-                    charge_point->on_session_started(1, uuid, "EVConnected");
+                    charge_point->on_session_started(1, uuid, "EVConnected", boost::none);
                     const auto result = charge_point->authorize_id_token(ocpp::CiString<20>(std::string("DEADBEEF")));
                     if (result.status == ocpp::v16::AuthorizationStatus::Accepted) {
                         charge_point->on_transaction_started(1, uuid, "DEADBEEF", 0, boost::none, ocpp::DateTime(),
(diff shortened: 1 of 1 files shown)


Afronut:
[boost reference](https://www.boost.org/doc/libs/1_34_0/libs/optional/doc/optional.html)

lib/ocpp/common/database_handler.cpp
```diff
@@ -323,6 +323,11 @@ boost::optional<v16::IdTagInfo> DatabaseHandler::get_authorization_cache_entry(c
         const auto expiry_date = DateTime(expiry_date_str);
         id_tag_info.expiryDate.emplace(expiry_date);
     }
+    // if expiry date is not set, calling has_value() on the returned optional will return true, but calling value()
```
Pietfried, line 326: Calling has_value() on an boost:optional will return false if the value of the optional is not set, so setting id_tag_info.expiryDate = boost::none is not necessary

Pietfried (changes requested):
Thanks for the fixes to the build examples.

Please remove the changes to the database_handler.cpp and then we can merge.

Afronut:
I will remove it however the client crash when local whitelist is enabled and the server send tags with no expiration so store locally. I have tried with Steve server
I can create a issue.

Pietfried:
> I will remove it however the client crash when local whitelist is enabled and the server send tags with no expiration so store locally. I have tried with Steve server I can create a issue.

Ok, I tried to reproduce it but could not get the same result. Please create a ticket with further information of your general setup / config file etc. so that we can try to reproduce the error.

Afronut:
I have removed the update to database handler

Pietfried (approved):

PR #76, opened 2023-02-28
Support for OCPP1.6 Plug&Charge Whitepaper

Pietfried:
General:
- Added support for 1.6 ISO15118 PnC whitepaper

PkiHandler:
- rework of pki_handler to support iso15118 certificate management
- added config parameter useSslDefaultVerifyPaths to libocpp config
- removed PkiHandler dependency from ChargePointConfiguration
- now checks number of installed CSMS Root CAs and if more than 1 is allowed

Websocket:
- adjusted websocket_tls to changes to pki_handler
- improved fallback mechanism of CSMS root ca
- WebsocketTLS now checks if single or mutliple csms root cas can be used to verify server cert

ChargePointConfiguration:
- Added UseSslDefaultVerifyPaths as configuration key
- Added OcspRequestInterval as configuration key
- Added OcspRequest timer in order to cache Ocsp Response for CSO Sub CAs
- Added Setters for all PnC configuration keys
- Configuration of Security profile now required

ChargePoint:
- moved csms certificate expiry timer from websocket_tls to charge_point
- added v2g certificate expiry timer
- added DataTransfer/PnC functionality


lib/ocpp/v16/charge_point.cpp
```diff
@@ -2162,6 +2320,435 @@ std::map<int32_t, ChargingSchedule> ChargePoint::get_all_composite_charging_sche
...
+void ChargePoint::data_transfer_pnc_sign_certificate() {
+
+    DataTransferRequest req;
+    req.vendorId = ISO15118_PNC_VENDOR_ID;
+    req.messageId.emplace(CiString<50>(std::string("SignCertificate")));
+
+    ocpp::v201::SignCertificateRequest csr_req;
+
+    // TODO(piet): Fix csr data
+    const auto csr = this->pki_handler->generateCsr(ocpp::CertificateSigningUseEnum::V2GCertificate, "DE", "BW",
```
hikinggrass, line 2415: Should we really hardcode these values for the moment?

tests/certs/INVALID_CERT.pem
hikinggrass, line 1: Do we want to keep these test certificates for the moment or already replace them by a script that generates them automatically?

hikinggrass (changes requested):

lib/ocpp/v16/charge_point.cpp
Pietfried: This will change anyways with the Hubject extension for additional configuration keys. I would keep this as it is now and change that with the PR for that.

PR #81, opened 2023-03-14
Fix behavior on Connection Initialization Error

Pietfried:
removed return; when tls connection could not be initialized


lib/ocpp/common/websocket/websocket_tls.cpp
```diff
@@ -187,7 +187,6 @@ void WebsocketTLS::connect_tls(int32_t security_profile, bool try_once) {
 
     if (ec) {
         EVLOG_error << "Connection initialization error for TLS websocket: " << ec.message();
-        return;
```
hikinggrass, line 190: What's the issue here? And why is it different from connect_plain() in websocket_plain where we still return on error?

Pietfried: We dont attemt to connect/reconnect if we return here before connection handle is setup

hikinggrass (approved):

PR #102, opened 2023-06-10
Adding SoC reporting and Calback for Transaction started 

Afronut:
added callback for start transaction, SoC and Cable temperature meter reporting, updated example code and updated supported measurand


Afronut:
- When transaction is started, the lib should provide the  actual transaction id from the back-end o the application. For example it does not make sense to use a session id with have to direct connection to the actual transaction number on the back-end to track transaction on the application level. This will also, be useful to  energy meters that need start  transaction with a transaction ID
- The current implementation does not report SoC or Temperature.  added

Pietfried (changes requested):
Hi,

Thanks for your contribution which looks nice to me at the first glance. Adding a transaction_started_callback is perfectly fine and reasonable. 

I have one remark for the support of SoC and Temperature:
By adding SoC and Temperature to the Powermeter struct the struct can not really be called a "Powermeter" anymore, because the SoC is read from the EV and the temperature could have any different source. Of course also the OCPP1.6 spec mixes up / combines these values in the MeterValue / SampledValue type so I would keep the general structure that we have in which we supply a struct that combines these values as well. Could you rename the Powermeter struct and related fields to a more reasonable name like "Measurement"?

PR #120, opened 2023-07-13
Remove dependency on edm and find_package to make it easier to build standalone

maaikez:


TheHardGamer:
Great work with this patch, it does fix most of the issues with a standalone build of ocpp but it still fails since libtimer tries to find the "date" library, though it does work if date is added as a submodule which is fine in my case. Thanks!

hikinggrass:
Superseded by https://github.com/EVerest/libocpp/pull/135

PR #128, opened 2023-07-24
Reset with ongoing transaction

maaikez:
Reset with ongoing transaction. 

Because we have to resend a transaction message if there is no 'ack' returned before rebooting, the database is also changed. I made this a draft review because I can not test it yet and I am a bit unsure about:
- Are the database changes ok and is it possible for someone to test this?
- I added the part about OnIdle and ImmediateReset (currently line 865 - 903 in chargepoint.cpp). Is it possible to test this? Will the submit_event work and will it work that the charging station will reset as soon as a transaction ends?

Thanks for looking into it!


maaikez:
Missing the use case where multiple evse's are charging and one stops charging, we still have to wait for a reset.

maaikez:
> Fixed this hopefully.

include/ocpp/common/message_queue.hpp
```diff
@@ -289,12 +314,13 @@ template <typename M> class MessageQueue {
     }
 
     MessageQueue(const std::function<bool(json message)>& send_callback, const int transaction_message_attempts,
-                 const int transaction_message_retry_interval) :
-        MessageQueue(send_callback, transaction_message_attempts, transaction_message_retry_interval, {}){};
+                 const int transaction_message_retry_interval, common::DatabaseHandlerBase& databaseHandler) :
```
hikinggrass, line 317: Would it be possible to pass the databaseHandler as a std::shared_ptr here? It's already a sharedptr in the charge point

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -47,7 +47,7 @@ ChargePointImpl::ChargePointImpl(const std::string& config, const std::filesyste
     this->message_queue = std::make_unique<ocpp::MessageQueue<v16::MessageType>>(
         [this](json message) -> bool { return this->websocket->send(message.dump()); },
         this->configuration->getTransactionMessageAttempts(), this->configuration->getTransactionMessageRetryInterval(),
-        this->external_notify);
+        this->external_notify, *this->database_handler.get());
```
hikinggrass, line 50: If you would use the sharedptr here you would not have to unpack it to pass it

hikinggrass (changes requested):

include/ocpp/common/message_queue.hpp
maaikez: No in the chargepoint it is a unique ptr. I can make it a shared ptr if you want, then I can pass it as shared ptr of course. But actually the charge point is the 'owner' of the pointer

lib/ocpp/v201/charge_point.cpp
```diff
@@ -1094,36 +1165,75 @@ void ChargePoint::handle_reset_req(Call<ResetRequest> call) {
...
+
+            // B12.FR.01: We have to wait until transactions have ended.
+            this->reset_scheduled = true;
+            response.status = ResetStatusEnum::Scheduled;
+        } else if (msg.type == ResetEnum::Immediate and
+                   !msg.evseId.has_value()) {
             // B12.FR.04
+            for (const auto &[evse_id, evse] : this->evses) {
+                if (evse->has_active_transaction()) {
+                    callbacks.stop_transaction_callback(
```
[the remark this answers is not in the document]
maaikez: Changed it, but not with the 5s because that's up to the charging station how long that can take. Can you take a look if this is better?

Also worked on the other comment about the message queue that is sending things before the accepted bootnotification. Is it better with this change?

include/ocpp/common/message_queue.hpp
hikinggrass: Oh I totally forgot about that, then we'll have to check if the access to the database handler with a shared ptr is a good idea. unpacking the unique ptr here is definitely not recommended

hikinggrass (approved):

PR #130, opened 2023-07-27
Add charging state to transaction so it will be sent when a transaction has been started

maaikez:


Pietfried (approved):
when merged we can quickly add support in everest-core for this.

PR #137, opened 2023-08-11
added callback for custom key configuration change

Afronut:
This pull request is to add a callback to set custom key for reconfiguration request. The OCPP 1.6j specification does not limit the key, value pair  that can be set for change configuration request. 
added this optional callback function to set optional keys.


hikinggrass:
Thanks! Wouldn't it also make sense to be able to get these configuration values in the GetConfiguration message?
Additionally how will the read/write permissions for these keys be handled? The implementation in this PR assumes this is WriteOnly, but theoretically they could be ReadOnly or ReadWrite as well

PR #139, opened 2023-08-14
External Websocket Control for OCPP201

Pietfried:
added public function to disconnect and connect the websocket connection for ocpp201


SebaLukas (approved):
👍

PR #145, opened 2023-08-17
Add partial trigger message support

marcemmers:
Still missing the following messages:
- PublishFirmwareStatusNotification
- SignChargingStationCertificate
- SignV2GCertificate
- SignCombinedCertificate

We are free to not implement messages as long as we return `NotImplemented` which is what we currently do.

Edit: I was able to implement LogStatusNotification and FirmwareStatusNotification as well but could not test it due to the logging not being implemented in everest/core yet or not being able to test the firmware update from the ocpp-testing code.

### DIFF
--- a/include/ocpp/v201/evse.hpp
+++ b/include/ocpp/v201/evse.hpp
@@ -99,8 +99,8 @@ public:
     /// \brief Triggers status notification callback for all connectors of the evse
     void trigger_status_notification_callbacks();
 
-    /// @brief Triggers a status notification callback for connector_id of the evse
-    /// @param connector_id id of the connector of the evse
+    /// \brief Triggers a status notification callback for connector_id of the evse
+    /// \param connector_id id of the connector of the evse
     void trigger_status_notification_callback(const int32_t connector_id);
 
     /// \brief Event handler that should be called when a new meter_value for this evse is present
--- a/include/ocpp/v201/evse.hpp
+++ b/include/ocpp/v201/evse.hpp
@@ -99,6 +99,10 @@ public:
     /// \brief Triggers status notification callback for all connectors of the evse
     void trigger_status_notification_callbacks();
 
+    /// @brief Triggers a status notification callback for connector_id of the evse
+    /// @param connector_id id of the connector of the evse
+    void trigger_status_notification_callback(const int32_t connector_id);
+
     /// \brief Event handler that should be called when a new meter_value for this evse is present
     /// \param meter_value
     void on_meter_value(const MeterValue& meter_value);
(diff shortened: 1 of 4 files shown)


lib/ocpp/v201/charge_point.cpp
```diff
@@ -1372,6 +1392,186 @@ void ChargePoint::handle_unlock_connector(Call<UnlockConnectorRequest> call) {
     this->send<UnlockConnectorResponse>(call_result);
 }
 
+void ChargePoint::handle_trigger_message(Call<TriggerMessageRequest> call) {
+    const TriggerMessageRequest& msg = call.msg;
+    TriggerMessageResponse response;
+    Evse* pEvse = nullptr;
```
Pietfried, line 1398: style wise we prefer snake_case here

lib/ocpp/v201/charge_point.cpp
```diff
@@ -1372,6 +1392,186 @@ void ChargePoint::handle_unlock_connector(Call<UnlockConnectorRequest> call) {
...
+        break;
+
+    case MessageTriggerEnum::StatusNotification:
+        if (msg.evse.has_value() and msg.evse.value().connectorId.has_value()) {
+            int32_t connector_id = msg.evse.value().connectorId.value();
+            if (pEvse != nullptr and connector_id > 0 and connector_id <= pEvse->get_number_of_connectors()) {
+                response.status = TriggerMessageStatusEnum::Accepted;
+            }
+        } else {
+            // F06.FR.12: Reject if evse or connectorId is ommited
```
Pietfried, line 1457: F06.FR.12 only applies if evse.id == 0. If its not present a StatusNotification for every connector of every EVSE should be sent.

include/ocpp/v201/evse.hpp
```diff
@@ -99,6 +99,10 @@ class Evse {
     /// \brief Triggers status notification callback for all connectors of the evse
     void trigger_status_notification_callbacks();
 
+    /// @brief Triggers a status notification callback for connector_id of the evse
```
Pietfried, line 102: use consistent markes. so far we have used \ instead of @

Pietfried (changes requested):
Hi Marc, apart from some minor requested changes this looks good to me. The StatusNotification trigger for all evse is missing. Have a look at the note in F06.FR.11

lib/ocpp/v201/charge_point.cpp
marcemmers: I read this as `evse` must be present, otherwise send a rejected so I think what I have is fine:

> If a Charging Station receives a
> TriggerMessageRequest with
> requestedMessage set to:
> StatusNotification AND
> ( evse is omitted OR
> evse.connectorId is omitted )

marcemmers:
Also the note in F06.FR.11 looks to me as, this requirement is not applied to StatusNotification messages.

lib/ocpp/v201/charge_point.cpp
marcemmers: I tried to use snake case but this was just a force of habit I guess. Will change, thanks :+1:

Pietfried:
I was referring to an older version of the spec... sorry for the confusion.

Pietfried (approved):

PR #150, opened 2023-08-23
Make it possible to use own logger instead of everest logger.

maaikez:


include/ocpp/common/charging_station_base.hpp
```diff
@@ -7,6 +7,8 @@
 #include <ocpp/common/pki_handler.hpp>
 #include <ocpp/common/websocket/websocket.hpp>
 
+#include <boost/shared_ptr.hpp>
```
marcemmers, line 10: Why is the boost shared_ptr still used?

maaikez: Because otherwise libocpp does fail to build.

hikinggrass (approved):

PR #153, opened 2023-08-24
Add local authorization list management calls (D01 and D02)

marcemmers:


marcemmers:
I was actually pondering if I should hash the token_id and token_id_type here as well but the ocpp spec does not have a recommendation for it for the local auth list. Any opinions here?

include/ocpp/v201/charge_point.hpp
```diff
@@ -239,6 +241,11 @@ class ChargePoint : ocpp::ChargingStationBase {
     // Functional Block C: Authorization
     void handle_clear_cache_req(Call<ClearCacheRequest> call);
 
+    // Functional Block D: Local authorization list management
+    SendLocalListStatusEnum apply_local_authorization_list(const Call<SendLocalListRequest>& call);
```
Pietfried, line 245: The argument of this function should be of type SendLocalListRequest, you dont need the Call here.

Also, so far we have only defined the OCPP message handlers at this location. The definition of this function could be moved to L162, where other private helper functions of these handlers are located.

Pietfried (changes requested):
Looks good apart from one minor comment. Please rebase and squash your commits

Pietfried (approved):

PR #157, opened 2023-08-31
Refactored pki_handler to evse_security

Pietfried:
* removed pki handler impl and header from libocpp and replaced by evse_security.hpp, which can be provided externally from libocpp consumer. 
* Refactored all parts where pki_handler was used with usage of evse_security
* added dependency to libevse-security and used this for implementation of evse_security interface. This internal evse_security_impl is used if no external implementation is given in the ChargePoint constructor


include/ocpp/common/evse_security.hpp
```diff
@@ -0,0 +1,109 @@
...
+class EvseSecurity {
+
+public:
+    /// \brief Installs the CA \p certificate for the given \p certificate_type . This function respects the
+    /// requirements of OCPP specified for the CSMS initiated message InstallCertificate.req .
+    /// \param certificate PEM formatted CA certificate
+    /// \param certificate_type specifies the CA certificate type
+    /// \return result of the operation
+    virtual InstallCertificateResult install_ca_certificate(const std::string& certificate,
+                                                            const CaCertificateType& certificate_type) = 0;
```
marcemmers, line 25: I was wondering this in our call earlier today too. Why are we passing enum class arguments by const reference? Seems like by value would be simpler and more code efficient as well.

Pietfried: this is just a matter of habbit, they could also be passed by value. In general, there are a couple of things we could target for a larger styling refactor of libocpp, which is already on our list :)

lib/ocpp/v201/charge_point.cpp
```diff
@@ -27,31 +27,27 @@ bool Callbacks::all_callbacks_valid() const {
...
         EVLOG_AND_THROW(std::invalid_argument("All non-optional callbacks must be supplied"));
     }
-
     this->device_model = std::make_unique<DeviceModel>(device_model_storage_address);
-    this->pki_handler = std::make_shared<ocpp::PkiHandler>(
-        certs_path,
-        this->device_model->get_optional_value<bool>(ControllerComponentVariables::AdditionalRootCertificateCheck)
-            .value_or(false));
-    this->database_handler = std::make_shared<DatabaseHandler>(core_database_path, sql_init_path);
+    this->evse_security = evse_security;
```
maaikez, line 49: Please remove this line. evse_security is already initialized in ChargingStationBase. If you set evse_security to nullptr, a new instance is created in ChargingStationBase and overwritten with this line.

hikinggrass (approved):

PR #160, opened 2023-09-04
introduction of measurement

Afronut:
replace power meter with measurement to allow sending SOC, temperature values to CS


Pietfried (changes requested):
Hi Coovy, this looks nice. In your changes the implementation of to_json, from_json and operator<< for StateOfCharge, Temperature, RPM and Measurement is missing in common/types.cpp. Could you add that to this PR?

Pietfried (approved):

PR #162, opened 2023-09-05
Added plug in timeout handler for OCPP1.6 to move from preparing to finished

Pietfried:


corneliusclaussen (approved):

include/ocpp/v16/charge_point.hpp
```diff
@@ -251,6 +251,10 @@ class ChargePoint {
     /// \param connector
     void on_disabled(int32_t connector);
 
+    /// \brief Notifies chargepoint that a ConnectionTimeout occured for the given \p connector . This function should
+    /// be called when an EV is plugged in but the authorization is present within the specified ConnectionTimeout
```
hikinggrass, line 255: Shouldn't this be "but the authorization is *not* present within..."?

hikinggrass (approved):

PR #164, opened 2023-09-05
pg_websocket_v_dtor - Correct WebSocket type destructor behavior

gregsq-proton:
The std::unique_ptr<WebSocketBase> websocket object holds the address of the base WebSocket. Its destructor won't join with the websocket_thread as it does in WebSocketTLS and WebSocketPlain's destructors. Assuming stop_perpetual can finish all work, and the Asio handling thread exits, the join(s) should be called. If stop_perpetual stays busy, then the join will wait indefinitely. If it stays busy and join isn't called, as occurs without this change, then reconnects can fail. If stop_perpetual doesn't stay busy, and join isn't called, we have a race.

### DIFF
--- a/include/ocpp/common/websocket/websocket.hpp
+++ b/include/ocpp/common/websocket/websocket.hpp
@@ -14,6 +14,7 @@ namespace ocpp {
 ///
 class Websocket {
 private:
+    // unique_ptr holds address of base - requires WebSocketBase to have a virtual destructor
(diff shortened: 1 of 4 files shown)


Pietfried (approved):
Thanks for bringing this up. Please rebase and sign off your commit so that DCO check passes: https://github.com/EVerest/libocpp/pull/164/checks?check_run_id=16524390097

PR #166, opened 2023-09-07
Added reduced state machine for connector 0 for OCPP1.6

Pietfried:
- Added reduced state machine for connector 0 for OCPP1.6
- chargepoint can now be initialized with initial chargepoint states ; if no states are provided, last states from persistant storage will be used to initialize the state machine
- reset of state machine is now public

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -175,6 +175,33 @@ void ChargePointImpl::init_websocket() {
 
     this->websocket->register_message_callback([this](const std::string& message) { this->message_callback(message); });
 }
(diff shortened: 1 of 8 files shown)


lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -3328,10 +3365,12 @@ void ChargePointImpl::on_reservation_end(int32_t connector) {
 }
 
 void ChargePointImpl::on_enabled(int32_t connector) {
+    EVLOG_critical << "On enabled: " << connector;
```
hikinggrass, line 3368: These two log messages should be removed

hikinggrass (changes requested):

hikinggrass (approved):

PR #167, opened 2023-09-08
Add check to validate if the callback struct is completely filled before starting

marcemmers:


marcemmers:
@Pietfried Ok I already found out that this might be a bit too simplistic. Maybe I should at least add something that tells you which callback is missing

Pietfried:
> @Pietfried Ok I already found out that this might be a bit too simplistic. Maybe I should at least add something that tells you which callback is missing

I agree, but its already an upgrade to what we have now, so I'll approve

Pietfried (approved):

PR #168, opened 2023-09-08
B04 offline idle behaviour

SNSubramanya:
Offline Behavior Idle Charging Station
1. B04.FR.01
2. B04.FR.02

### DIFF
--- a/include/ocpp/v201/charge_point.hpp
+++ b/include/ocpp/v201/charge_point.hpp
@@ -133,6 +133,24 @@ private:
     int network_configuration_priority;
     bool disable_automatic_websocket_reconnects;
 
+    // store the connector status
+    struct EvseConnectorPair {
+        int32_t evse_id;
+        int32_t connector_id;
+
+        // Define a comparison operator for the struct
+        bool operator<(const EvseConnectorPair& other) const {
+            // Compare based on name, then age
+            if (evse_id != other.evse_id) {
+                return evse_id < other.evse_id;
+            }
+            return connector_id < other.connector_id;
+        }
+    };
+
+    std::map<EvseConnectorPair, ConnectorStatusEnum> conn_state_per_evse;
+    std::chrono::time_point<std::chrono::steady_clock> time_disconnected;
+
     /// \brief Used when an 'OnIdle' reset is requested, to perform the reset after the charging has stopped.
     bool reset_scheduled;
     /// \brief If `reset_scheduled` is true and the reset is for a specific evse id, it will be stored in this member.
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -470,6 +470,35 @@ void ChargePoint::init_websocket() {
     this->websocket->register_connected_callback([this](const int security_profile) {
         this->message_queue->resume();
         this->websocket_connection_status = WebsocketConnectionStatusEnum::Connected;
+
+        if (this->registration_status == RegistrationStatusEnum::Accepted) {
+            // handle offline threshold
+            //  Get the current time point using steady_clock
+            std::chrono::steady_clock::duration offline_duration = std::chrono::steady_clock::now() - time_disconnected;
+
+            // B04.FR.01
+            // If offline period exceeds offline threshold then send the status notification for all connectors
+            if (offline_duration > std::chrono::seconds(this->device_model->get_value<int>(
+                                       ControllerComponentVariables::OfflineThreshold))) {
+                EVLOG_debug << "offline for more than offline threshold ";
+                for (auto const& [evse_id, evse] : this->evses) {
+                    evse->trigger_status_notification_callbacks();
+                }
+            } else {
+                // B04.FR.02
+                // If offline period doesn't exceed offline threshold then send the status notification for all
+                // connectors that changed state
+                EVLOG_debug << "offline for less than offline threshold ";
+                for (auto const& [evse_id, evse] : this->evses) {
+                    int number_of_connectors = evse->get_number_of_connectors();
+                    for (int connector_id = 1; connector_id <= number_of_connectors; connector_id++) {
+                        if (conn_state_per_evse[{evse_id, connector_id}] != evse->get_state(connector_id)) {
+                            evse->trigger_status_notification_callback(connector_id);
+                        }
+                    }
+                }
+            }
+        }
     });
 
     this->websocket->register_closed_callback(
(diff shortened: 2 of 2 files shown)


lib/ocpp/v201/charge_point.cpp
```diff
@@ -425,6 +425,26 @@ void ChargePoint::init_websocket() {
...
+            EVLOG_debug << "LIBOCPP: Offline < threshold";
+            // get the status of all the connectors
+            for (auto const& [evse_id, evse] : this->evses) {
+                int number_of_connectors = evse->get_number_of_connectors();
+                EVLOG_debug << "LIBOCPP: evseId: " << evse_id << " numConn: " << number_of_connectors;
+                for (int connector_id = 1; connector_id <= number_of_connectors; connector_id++) {
+                    if (!(conn_state_per_evse[{evse_id,connector_id}].connector_status ==
+                          evse->get_state(connector_id))) {
+                        // if the state varied when offline
+                        conn_state_per_evse[{evse_id,connector_id}].is_status_changed_when_offline = true;
```
marcemmers, line 441: Why do we need to store this value for use later instead of using it directly?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -436,7 +456,33 @@ void ChargePoint::init_websocket() {
...
+            int offline_threshold = this->device_model->get_value<int>(ControllerComponentVariables::OfflineThreshold);
+            if (offline_threshold > 0) {
+
+                // get the status of all the connectors
+                for (auto const& [evse_id, evse] : this->evses) {
+                    int number_of_connectors = evse->get_number_of_connectors();
+                    conn_states_per_evse_id_key conn_states_struct_key;
+                    conn_states_per_evse_id_value conn_states_struct_value;
+                    EVLOG_debug<< "LIBOCPP: evseId: " << evse_id << " numConn: " << number_of_connectors;
+                    for (int connector_id = 1; connector_id <= number_of_connectors; connector_id++) {
```
marcemmers, line 470: I don't like the fact that you need to iterate the connectors here manually. I know this is currently the only option, but maybe we can improve this by adding a getter function for the entire connector map of an Evse?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -436,7 +456,33 @@ void ChargePoint::init_websocket() {
...
+                        conn_states_struct_key.evse_id = evse_id;
+                        conn_states_struct_key.connectior_id = connector_id;
+                        conn_state_per_evse[conn_states_struct_key].connector_status = evse->get_state(connector_id);
+                        EVLOG_debug << "conn_id: " << conn_states_struct_key.connectior_id
+                                   << " State: " << conn_state_per_evse[conn_states_struct_key].connector_status;
+                    }
+                }
 
+                this->offline_threshold_timer.interval(
```
marcemmers, line 479: I still think it is a cleaner solution to just store the timestamp here and compare it when going back online. Now we have a bool variable and an entire timer object stored and the functionality is just the same.

`std::chrono::time_point<std::chrono::steady_clock>` should do the trick. It is a monotonic timer so no worries about it moving backwards or whatever.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -425,6 +425,26 @@ void ChargePoint::init_websocket() {
...
+
+        if (!is_offline_threshold_expired) {
+            EVLOG_debug << "LIBOCPP: Offline < threshold";
+            // get the status of all the connectors
+            for (auto const& [evse_id, evse] : this->evses) {
+                int number_of_connectors = evse->get_number_of_connectors();
+                EVLOG_debug << "LIBOCPP: evseId: " << evse_id << " numConn: " << number_of_connectors;
+                for (int connector_id = 1; connector_id <= number_of_connectors; connector_id++) {
+                    if (!(conn_state_per_evse[{evse_id,connector_id}].connector_status ==
+                          evse->get_state(connector_id))) {
```
marcemmers, line 439:
```suggestion
                    if (conn_state_per_evse[{evse_id,connector_id}].connector_status !=
                          evse->get_state(connector_id)) {
```

marcemmers (changes requested):

lib/ocpp/v201/charge_point.cpp
marcemmers: What about just checking the registration status in the connection callback? When we are accepted we stay accepted even after a lost connection. If we were not accepted yet we can just drop the changes since we are required to send the status notifications of all connectors anyway after a BootNotificationResponse with accepted (B01.FR.05).

lib/ocpp/v201/charge_point.cpp
```diff
@@ -425,8 +425,43 @@ void ChargePoint::init_websocket() {
...
+        }
+
         this->message_queue->resume();
         this->websocket_connection_status = WebsocketConnectionStatusEnum::Connected;
+
+        if (this->registration_status == RegistrationStatusEnum::Accepted) {
+
+            // B04.FR.01
+            // If offline period exceeds offline threshold then send the status notification for all connectors
+            if (is_offline_threshold_expired) {
```
marcemmers, line 444: What about?
```suggestion
            auto offline_duration = std::chrono::steady_clock::now() - time_disconnected;

            if (offline_duration >= offline_threshold_value) {
```

Then you don't need `is_offline_threshold_expired` anymore and you could also use the threshold value directly without storing anything?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -450,7 +505,7 @@ void ChargePoint::init_websocket() {
         });
 
     this->websocket->register_message_callback([this](const std::string& message) { this->message_callback(message); });
-}
+        }
```
marcemmers, line 508: Please revert this ;)

marcemmers (changes requested):

marcemmers (approved):

include/ocpp/v201/charge_point.hpp
```diff
@@ -133,6 +133,24 @@ class ChargePoint : ocpp::ChargingStationBase {
     int network_configuration_priority;
     bool disable_automatic_websocket_reconnects;
 
+    // store the connector status
+    struct evse_connector_pair {
```
Pietfried, line 137: we prefer CamelCase here

include/ocpp/v201/charge_point.hpp
```diff
@@ -133,6 +133,24 @@ class ChargePoint : ocpp::ChargingStationBase {
...
+        int32_t evse_id;
+        int32_t connectior_id;
```
Pietfried, line 139: typo

lib/ocpp/v201/charge_point.cpp
```diff
@@ -470,27 +470,76 @@ void ChargePoint::init_websocket() {
...
-                        if (reason != websocketpp::close::status::service_restart) {
-                            this->next_network_configuration_priority();
+        if (this->registration_status == RegistrationStatusEnum::Accepted) {
+            // handle offline threshold
+            //  Get the current time point using steady_clock
+            std::chrono::steady_clock::duration offline_duration = std::chrono::steady_clock::now() - time_disconnected;
+
+            // B04.FR.01
+            // If offline period exceeds offline threshold then send the status notification for all connectors
+            if (offline_duration > (std::chrono::seconds)this->device_model->get_value<int>(
```
Pietfried, line 481: nit: Prefer not to cast but use

`std::chrono::seconds(this->device_model->get_value<int>(
                                       ControllerComponentVariables::OfflineThreshold))`

instead

lib/ocpp/v201/charge_point.cpp
```diff
@@ -470,27 +470,76 @@ void ChargePoint::init_websocket() {
...
+    this->websocket->register_closed_callback([this, connection_options,
+                                               configuration_slot](const websocketpp::close::status::value reason) {
+        EVLOG_warning << "Failed to connect to NetworkConfigurationPriority: "
+                      << this->network_configuration_priority + 1
+                      << " which is configurationSlot: " << configuration_slot;
+        this->websocket_connection_status = WebsocketConnectionStatusEnum::Disconnected;
+        this->message_queue->pause();
+
+        // check if offline threshold has been defined
+        if ((std::chrono::seconds)this->device_model->get_value<int>(ControllerComponentVariables::OfflineThreshold) !=
```
Pietfried, line 513: Casting both to std::chrono::seconds not necessary here; Could just be 

`this->device_model->get_value<int>(ControllerComponentVariables::OfflineThreshold) != 0`

Pietfried (changes requested):
Looks good apart from some minor styling remarks and a typo :+1:

include/ocpp/v201/charge_point.hpp
Pietfried, line 137: could also be defined publicly, I could imagine we could reuse this for other purposes. We could also move this later though

Pietfried (approved):

PR #169, opened 2023-09-12
support for boost/filesystem.hpp

bWF0dGhpYXMK:


Dominik-K:
Hi @bWF0dGhpYXMK, thanks for your first contribution. Can you please sign your commits with the DCO, please? See https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md

May you give some context why you need support for a version older than C++17? I'm unsure if we want to support older C++-version. But when you give some good arguments, perhaps we will.

Please add at least the `BOOST_FILESYSTEM_NO_DEPRECATED` (if we're going to proceed here).

bWF0dGhpYXMK:
For the moment, we have only support for BOOST_FILESYSTEM_VERSION 3. We make use of an SDK, which is in the moment not subject of change (HW support issues with later SDKs). I understand the point, as adding this change into mainline path, affects, many files and includes the start of having #defines inside the code and the CMakefile modification.

Without the pull request, the files have to be changed, after each release update to make use of libocpp with our SDK. 
For the moment, I have not found a good way to add changes into the libocpp which are not part of the clone.

Summary: I have no real good argument, only that it is simpler to be follow libocpp updates with our SDK, without hand patching to many files. 

Modification: The #ifndef switches good be placed into a header file, and this header file being included, to increase the readability.

PR #163, opened 2023-09-05
Configurable WaitForStopTransactionsOnResetTimeout for ocpp1.6

Pietfried:

PR #171, opened 2023-09-14
Check if all transctions are closed before reset waiter threat is not…

bWF0dGhpYXMK:
Inside handleStopTransactionResponse, it is checked, if all transcations ( Connector Number >1) are closed, before the Reset Waiter Threat is notified.

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -1579,7 +1579,7 @@ void ChargePointImpl::handleResetRequest(ocpp::Call<ResetRequest> call) {
         this->reset_thread = std::thread([this, reset_type]() {
             EVLOG_debug << "Waiting until all transactions are stopped...";
             std::unique_lock lk(this->stop_transaction_mutex);
-            this->stop_transaction_cv.wait_for(lk, std::chrono::seconds(5), [this] {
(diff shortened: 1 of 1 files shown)


Pietfried:
This looks quite similar to this PR: https://github.com/EVerest/libocpp/pull/163

bWF0dGhpYXMK:
The major change is inside: handleStopTransactionResponse( ... ) in the end before the notify call. Assuming there are two charging sessions running. The first of the two stoptransactionResponse is called, the notify would be triggered and the reset Waiter Threat is awaked. Also there is still one charing session open? This is the change about

Pietfried:
> The major change is inside: handleStopTransactionResponse( ... ) in the end before the notify call. Assuming there are two charging sessions running. The first of the two stoptransactionResponse is called, the notify would be triggered and the reset Waiter Threat is awaked. Also there is still one charing session open? This is the change about.
> 
> With the loop, it is checked if all charging sessions are terminated before the notify is triggered to awake the reset waiter threat.

Its fine that the reset waiter thread is awaked because it would check the condition its waiting for again and in case it is still false it would wait for the next notify or the timeout. So the part that you added to handleStopTransactionResponse is redundant

bWF0dGhpYXMK:
Thank your for your time on that. You are right, the code is redundant. It took me a while to find this link [wait_until](https://cplusplus.com/reference/condition_variable/condition_variable/wait_until/) which describes the behavior more clearly.

PR #172, opened 2023-09-14
Make usage of HostName key in Internal.json, to be able to use differ…

bWF0dGhpYXMK:
Make usage of HostName key in Internal.json, to be able to use different User-Host inside the HTTP request, then used inside centralSystemUri.


Pietfried (changes requested):
This looks good.

Please add the function getHostNameKeyValue() and add this configuration key and new function in the get() function of ChargePointConfiguration.

You can take getWebsocketPingPayload as an example. Otherwise the hostname would not be reported via a GetConfiguration.req 

Please squash and sign-off your commit: https://github.com/EVerest/libocpp/pull/172/checks?check_run_id=16787639262

PR #173, opened 2023-09-15
E04-offline transactions

SNSubramanya:
E04.FR.03

### DIFF
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -181,7 +181,8 @@ void ChargePoint::on_transaction_started(
 
     this->transaction_event_req(TransactionEventEnum::Started, timestamp, transaction, trigger_reason,
                                 enhanced_transaction->get_seq_no(), std::nullopt, evse, enhanced_transaction->id_token,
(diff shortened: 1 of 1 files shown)


marcemmers (approved):

lib/ocpp/v201/charge_point.cpp
```diff
@@ -218,8 +229,13 @@ void ChargePoint::on_transaction_finished(const int32_t evse_id, const DateTime&
         id_token_opt = _id_token;
     }
 
+    std::optional<bool> offline;
+    if (this->websocket_connection_status == WebsocketConnectionStatusEnum::Disconnected) {
```
Pietfried, line 233: nit: I would put this into a separate function because this is the same code 6 times. We could also change the signature of transaction_event_req to not take a std::optional<bool> but just a bool instead and always report true or false in every TransactionEvent.req message

Pietfried (approved):

lib/ocpp/v201/charge_point.cpp
marcemmers: My 2 cents on this: Only send the `offline` value if we were offline and otherwise omit it. All the functionality is there and it saves a bit of data, even if its a minor amount.

You could indeed change it to a bool in transaction_event_req and do the omitting in that function.

PR #177, opened 2023-09-18
Support older C++ versions with `boost::filesystem`

bWF0dGhpYXMK:


bWF0dGhpYXMK:
Please close https://github.com/EVerest/libocpp/pull/169 ( I hope that works )

For the moment, we have only support for BOOST_FILESYSTEM_VERSION 3. We make use of an SDK, which is in the moment not subject of change (HW support issues with later SDKs). I understand the point, as adding this change into mainline path, affects, many files and includes the start of having #defines inside the code and the CMakefile modification.

Without the pull request, the files have to be changed, after each release update to make use of libocpp with our SDK.
For the moment, I have not found a good way to add changes into the libocpp which are not part of the clone.

Summary: I have no real good argument, only that it is simpler to be follow libocpp updates with our SDK, without hand patching to many files.

Modification: The #ifndef switches good be placed into a header file, and this header file being included, to increase the readability.

Dominik-K:
@bWF0dGhpYXMK
> Modification: The #ifndef switches good be placed into a header file, and this header file being included, to increase the readability.

That would be really nice to avoid a lot of duplication. I suggest to name the file `support_older_c++-versions.hpp`.

Dominik-K:
> Please close https://github.com/EVerest/libocpp/pull/169 ( I hope that works )

I think you should be able to do it as well (as PR-creator). At least only you can delete the branch in your fork.

lib/CMakeLists.txt
```diff
@@ -74,13 +74,30 @@ endif()
...
+        date::date-tz
+)
+else()
 target_link_libraries(ocpp
     PUBLIC
         everest::timer
         websocketpp::websocketpp
         nlohmann_json_schema_validator
     PRIVATE
+  		Boost::filesystem
```
Dominik-K, line 100: Would it be possible to add another `target_link_libraries()` with just these lines? To avoid doubling the other lines.

include/ocpp/common/pki_handler.hpp
```diff
@@ -3,7 +3,11 @@
 #ifndef OCPP_COMMON_PKI_HANDLER
 #define OCPP_COMMON_PKI_HANDLER
 
-#include <filesystem>
+#ifndef BOOSTFILESYSTEM
+<filesystem>
```
Dominik-K, line 7: Does this work without an `#include` before it?

CMakeLists.txt
```diff
@@ -16,6 +16,14 @@ option(CMAKE_RUN_CLANG_TIDY "Run clang-tidy" OFF)
 option(LIBOCPP_BUILD_EXAMPLES "Build charge_point and central_system binaries." OFF)
 option(OCPP_INSTALL "Install the library (shared data might be installed anyway)" ${EVC_MAIN_PROJECT})
 
+if (${CMAKE_CXX_STANDARD} LESS_EQUAL 17)
```
Dominik-K, line 19:
```suggestion
if (${CXX_STANDARD} LESS 17) 
```

1. `CMAKE_CXX_STANDARD` is only the default value for `CXX_STANDARD`. See https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html
2. `LESS` because `std::filesystem` exists since C++17.

hikinggrass: We could also skip the C++ version detection here completely and rely on the user the provide the option when this is desired, like I did in https://github.com/EVerest/liblog/pull/13

Dominik-K: Yeah, actually I like this approach more. As this helps that the build works out of the box. And it discourages to use `boost::filesystem` on newer C++-versions. So it's explicit that this serves only as a backwards-compatibility.

include/ocpp/common/pki_handler.hpp
bWF0dGhpYXMK: Thank you i fixed that.

CMakeLists.txt
bWF0dGhpYXMK: On my system CMAKE_CXX_STANDARD delivered 17 although std::filesystem was not present. So I am pruposing to bind it to the compiler version:
Some interesting note on that, i found here: ["C++17 mode is the default since GCC 11; it can be explicitly selected with the -std=c++17 command-line flag, or -std=gnu++17 to enable GNU extensions as well. Some C++17 features are available since GCC 5, but support was experimental and the ABI of C++17 features was not stable until GCC 9. "](https://gcc.gnu.org/projects/cxx-status.html#cxx17)

src/charge_point.cpp
```diff
@@ -5,7 +5,7 @@
 #include <csignal>
 #include <date/date.h>
 #include <date/tz.h>
-#include <filesystem>
+#include <ocpp/common/support_older_c++_versions.hpp>
```
Dominik-K, line 8:
```suggestion
```

Already in `#include <ocpp/v16/charge_point.hpp>` below.

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -56,9 +60,15 @@ ChargePointImpl::ChargePointImpl(const std::string& config, const std::filesyste
...
     this->logging = std::make_shared<ocpp::MessageLogging>(
-        this->configuration->getLogMessages(), message_log_path, DateTime().to_rfc3339(), log_to_console,
-        detailed_log_to_console, log_to_file, log_to_html, session_logging);
+    		        this->configuration->getLogMessages(), message_log_path.string(), DateTime().to_rfc3339(), log_to_console,
+    		        detailed_log_to_console, log_to_file, log_to_html, session_logging);
+	#else
+		this->logging = std::make_shared<ocpp::MessageLogging>(
+		        this->configuration->getLogMessages(), message_log_path.string(), DateTime().to_rfc3339(), log_to_console,
+		        detailed_log_to_console, log_to_file, log_to_html, session_logging);
+	#endif
```
Dominik-K, line 71: Don't see any difference between the two. :thinking:

CMakeLists.txt
Dominik-K: Hmm, interesting. Did you [explicitly set the compiler for `cmake`](https://stackoverflow.com/questions/45933732/how-to-specify-a-compiler-in-cmake)?

Otherwise perhaps the `CXX`-env points to `gcc` and `cmake` then uses the default for `gcc` (which is C++17 since GCC11, as you pointed out).

Which compiler and version are you using? (If you can't name it publicly, perhaps via e-mail to me?)

lib/ocpp/v16/charge_point_impl.cpp
bWF0dGhpYXMK: In the beyond part the snd argument, uses message_log_path.string()

CMakeLists.txt
bWF0dGhpYXMK: I tried CXX_STANDARD but this was empty.

include/ocpp/common/support_older_c++_versions.hpp
```diff
@@ -0,0 +1,26 @@
+/*
+ * support_older_c++-versions.hpp
+ *
+ *  Created on: 20.09.2023
+ *      Author: matthias_suess
+ */
```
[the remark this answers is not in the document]
hikinggrass: Yes, this should be in the SPDX format, like 
```cpp
// SPDX-License-Identifier: Apache-2.0
// Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
```

CMakeLists.txt
hikinggrass: Given that there seem to be issues with the automatic detection of this I would propose using a cmake flag like
LIBOCPP_USE_BOOST_FILESYSTEM that you have to explicitly set (like in the aforementioned PR in liblog)

Dominik-K: Ok, fine. I suggest then to name it without the `LIBOCPP`-prefix. Otherwise you'll need to set multiple flags for the same feature. 

To be future-proof: `USE_NON_STD_FILESYSTEM`. It was suggested within Pionix to consider an alternative to `boost::filesystem`, e.g. https://github.com/gulrak/filesystem. Given that we're currently seeing multiple issues with it. See related
- https://github.com/EVerest/liblog/pull/13
- https://github.com/EVerest/everest-core/pull/348

Dominik-K (approved):
LGTM

@bWF0dGhpYXMK Can you check if it's still working on your side, please?
@hikinggrass Please review.

lib/ocpp/v16/charge_point_impl.cpp
Dominik-K: Actually, there's literally no difference between the two.

bWF0dGhpYXMK:
> LGTM
> 
> @bWF0dGhpYXMK Can you check if it's still working on your side, please? @hikinggrass Please review.

Thank you, i recompiled and worked here using boost::filesystem.

CMakeLists.txt
```diff
@@ -15,6 +15,17 @@ option(BUILD_TESTING "Run unit tests" OFF)
 option(CMAKE_RUN_CLANG_TIDY "Run clang-tidy" OFF)
 option(LIBOCPP_BUILD_EXAMPLES "Build charge_point and central_system binaries." OFF)
 option(OCPP_INSTALL "Install the library (shared data might be installed anyway)" ${EVC_MAIN_PROJECT})
+option(USE_NON_STD_FILESYSTEM "Usage of boost/filesystem.hpp instead of std::filesystem" OFF)
```
hikinggrass, line 18: I would suggest we name this option 
```cmake
option(LIBOCPP_USE_BOOST_FILESYSTEM "Usage of boost/filesystem.hpp instead of std::filesystem" OFF)
```
To make it prefixed to just this library and to clarify the function of it. I would rather rename this option later on if we choose to change the underlying implementation again

hikinggrass (changes requested):

CMakeLists.txt
Dominik-K: Did you see my [last comment on this in the comment above](https://github.com/EVerest/libocpp/pull/177#discussion_r1331797400)?

include/ocpp/common/sqlite_statement.hpp
```diff
@@ -4,9 +4,9 @@
 #ifndef SQLITE_STATEMENT_HPP
 #define SQLITE_STATEMENT_HPP
 
-#include <filesystem>
 #include <sqlite3.h>
 
+#include "support_older_c++_versions.hpp"
```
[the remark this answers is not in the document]
bWF0dGhpYXMK: Thank you for the feedback. I incorporated the change of the header file name and moves the Option request into lib/CMakelist.txt

CMakeLists.txt
hikinggrass: Yes but I would disagree in this instance and strongly recommend the usage of the prefixed variable LIBOCPP_USE_BOOST_FILESYSTEM, we can always set a few of these variables in everest-core if needed but this makes it easier to specifically enable/disable this feature lib-by-lib

include/ocpp/common/support_older_c++_versions.hpp
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
+
+#ifndef OCPP_COMMON_SUPPORT_OLDER_CPP_VERSIONS_HPP_
+#define OCPP_COMMON_SUPPORT_OLDER_CPP_VERSIONS_HPP_
+
+#ifndef BOOSTFILESYSTEM
```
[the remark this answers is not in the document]
bWF0dGhpYXMK: Hi, i renamed it, maybe it increases the chance, that it is included in the september release...

CMakeLists.txt
Dominik-K: Ok, makes sense. :+1:

include/ocpp/common/support_older_c++_versions.hpp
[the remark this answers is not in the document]
Dominik-K: Adapted it also in the macros (what this comment is about): https://github.com/EVerest/libocpp/pull/177/commits/6395c150b78ebe13413e951c2a5ac4347cc3f34a

>  that it is included in the September release...

Yeah, it should be. :+1:

include/ocpp/common/sqlite_statement.hpp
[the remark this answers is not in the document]
Dominik-K: Fix for this comment in https://github.com/EVerest/libocpp/pull/177/commits/5e23594b7cd379f296c73935a81d52b2d12fbdc0

Dominik-K (approved):
LGTM

Pietfried (approved):

hikinggrass (approved):

PR #178, opened 2023-09-18
Binding json default construct to use initializer list based constructor

bWF0dGhpYXMK:
The gcc 7.3 generates for json({}) null string, and not {}. With the proposed workaround, the right constructor is selected providing {}. See [link](https://github.com/nlohmann/json/issues/2046)


bWF0dGhpYXMK:
[Link](https://github.com/nlohmann/json/issues/2046)

Pietfried:
We added this to the code generator and added a separate PR for this

PR #179, opened 2023-09-19
Binding json default construct to use initializer list based constructor

hikinggrass:
GCC 7.3 generates a null string for json({}), and not {}. With this workaround, the right constructor is selected providing {}.

This is based on https://github.com/EVerest/libocpp/pull/178 but fixes the code generator to extend the fix to all messages of OCPP 1.6 and 2.0.1


bWF0dGhpYXMK:
Thank you!

PR #180, opened 2023-09-20
Approach to support openssl 1.1.1f and avoid selection via defintion …

bWF0dGhpYXMK:
…by using function pointer.

Alternative is to use #if based selection inside the code files. A simple definition inside the header, leaded to multiple definition error during linking, that is why a work around via extern is presented. 

Maybe the way via #if is more convenient. The pull request is reasoned, that our openssl is not supporting the functions, and requires minor work around. See commit. Thank you for your time on that.


hikinggrass:
This will be handled in the new [libevse-security](https://github.com/EVerest/libevse-security)

PR #182, opened 2023-09-21
Added support to report and consume security events in OCPP1.6

Pietfried:


hikinggrass (approved):

RobertDeLeeuw:
@Pietfried there is no checking for ciritical security events.
Now all libOCPP internal security events will get send to the CPMS, where A04.FR.01 says: 

The Charge Point SHALL inform the Central System of the security events by sending a SecurityEventNotification.req, to the Central System.

We dont have to send them all. I have added this to the 2.0.1 implementation. Another boolean to the internal function.

Pietfried:
> A04.FR.01

@RobertDeLeeuw I checked the spec and you are correct that only critical security events shall be sent. Still OCTT1.6 tests for SecurityEventRequests being sent like InvalidChargePointCertificate (which is not critical). I would propably go with just sending all of them, as this is not specifically prohibited in the spec as far as I can see. What do you think?

PR #183, opened 2023-09-22
Fix build with clang < 16

hikinggrass:


Dominik-K (approved):
LGTM.

Building `libocpp` with `clang-14`  works now locally on my side.

PR #191, opened 2023-09-27
J01- Add evseId=0 meter values

SNSubramanya:


lib/ocpp/v201/charge_point.cpp
```diff
@@ -284,7 +295,18 @@ void ChargePoint::on_session_finished(const int32_t evse_id, const int32_t conne
 }
 
 void ChargePoint::on_meter_value(const int32_t evse_id, const MeterValue& meter_value) {
-    this->evses.at(evse_id)->on_meter_value(meter_value);
+    std::lock_guard<std::mutex> lk(this->meter_value_mutex);
```
marcemmers, line 298: You only need to lock this in case your writing to `this->meter_value`, not the evse because it has its own lock ;)

lib/ocpp/v201/charge_point.cpp
```diff
@@ -824,9 +846,10 @@ void ChargePoint::update_aligned_data_interval() {
     const auto next_timestamp = this->get_next_clock_aligned_meter_value_timestamp(
         this->device_model->get_value<int>(ControllerComponentVariables::AlignedDataInterval));
     if (next_timestamp.has_value()) {
+        bool transaction_active = false;
         EVLOG_debug << "Next meter value will be sent at: " << next_timestamp.value().to_rfc3339();
         this->aligned_meter_values_timer.at(
-            [this]() {
+            [this, &transaction_active]() {
```
marcemmers, line 852: I don't think this is safe to do because at the time the lambda function will be called the boolean will be out of scope.

marcemmers (changes requested):

config/v201/config.json
```diff
@@ -1,56 +1,53 @@
...
-                "attributes": {
-                    "Actual": ""
-                }
-            },
-            "ConnectorSupplyPhases": {
-                "variable_name": "SupplyPhases",
-                "attributes": {
-                    "Actual": 42
-                }
+{
```
marcemmers, line 1: So it seems like you made some more changes here during the rebase than intended ;)

lib/ocpp/v201/charge_point.cpp
```diff
@@ -302,7 +302,18 @@ void ChargePoint::on_session_finished(const int32_t evse_id, const int32_t conne
 }
 
 void ChargePoint::on_meter_value(const int32_t evse_id, const MeterValue& meter_value) {
-    this->evses.at(evse_id)->on_meter_value(meter_value);
+    if (evse_id == 0) {
+        std::lock_guard<std::mutex> lk(this->meter_value_mutex);
+        // if evseId = 0 ten store in the chargepoint metervalues
```
marcemmers, line 307: Nitpick but:
```suggestion
        // if evseId = 0 then store in the chargepoint metervalues
```

marcemmers (changes requested):

PR #202, opened 2023-09-28
Added a time sync callback that can be used to sync system time

maaikez:


marcemmers:
Please also add a check in `all_callbacks_valid` so that if the optional has a value it is at least valid.

hikinggrass (approved):

PR #205, opened 2023-10-02
Use database to store transaction metervalues to be sent at the end of a transaction

marcemmers:


lib/ocpp/v201/charge_point.cpp
```diff
@@ -82,24 +82,41 @@ ChargePoint::ChargePoint(const std::map<int32_t, int32_t>& evse_connector_struct
         auto transaction_meter_value_callback = [this](const MeterValue& _meter_value, const Transaction& transaction,
                                                        const int32_t seq_no,
                                                        const std::optional<int32_t> reservation_id) {
-            const auto filtered_meter_value = utils::get_meter_value_with_measurands_applied(
-                _meter_value, utils::get_measurands_vec(this->device_model->get_value<std::string>(
-                                  ControllerComponentVariables::SampledDataTxUpdatedMeasurands)));
+            if (_meter_value.sampledValue.empty() or !_meter_value.sampledValue.at(0).context.has_value()) {
+                EVLOG_info << "Not sending due to no values";
```
Pietfried, line 86: nit:
```suggestion
                EVLOG_info << "Not sending MeterValue due to no values";
```

lib/ocpp/v201/charge_point.cpp
```diff
@@ -82,24 +82,41 @@ ChargePoint::ChargePoint(const std::map<int32_t, int32_t>& evse_connector_struct
...
+                return;
+            }
+
+            auto type = _meter_value.sampledValue.at(0).context.value();
+            if (type != ReadingContextEnum::Sample_Clock and type != ReadingContextEnum::Sample_Periodic) {
+                EVLOG_info << "Not sending due to wrong context";
```
Pietfried, line 92: nit:
```suggestion
                EVLOG_info << "Not sending MeterValue due to wrong context";
```

lib/ocpp/v201/charge_point.cpp
```diff
@@ -264,6 +286,8 @@ void ChargePoint::on_transaction_finished(const int32_t evse_id, const DateTime&
                                 std::nullopt, std::nullopt, id_token_opt, meter_values, std::nullopt,
                                 this->is_offline(), std::nullopt);
 
+    this->database_handler->transaction_metervalues_clear(transaction_id);
```
Pietfried, line 289: In case of a powerloss we might want to call this function at startup or clear the complete table in that case after the TransactionEvent.req(Ended) was sent. The complete handling for properly resuming/terminating transactions after power loss is propably something for another PR though.

Pietfried (approved):

lib/ocpp/v201/charge_point.cpp
marcemmers: Currently this is done in the database_handler file itself. On init we clear the whole table. There should indeed be better handling of this but we don't have any context on that transaction anymore anyhow. So indeed something for a later PR.

PR #209, opened 2023-10-10
Add message logging callback

marcemmers:


hikinggrass:
Maybe we can integrate this directly into the websocket so it will also work for 1.6

hikinggrass (approved):

PR #213, opened 2023-10-11
OCPP201: Improved handling of ChangeAvailability.req

Pietfried:
* in case of an active transaction we were responding with `Scheduled` instead of `Accepted` to a ChangeAvailability.req(status=Operative) - this has been fixed and is required by G03.FR.03 and G04.FR.04
* now overrding or removing scheduled availability changes if needed

### DIFF
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -898,20 +898,17 @@ bool ChargePoint::any_transaction_active(const std::optional<EVSE>& evse) {
         }
         return false;
     }
-
-    if (this->evses.at(evse.value().id)->has_active_transaction()) {
-        return true;
-    } else {
-        return false;
-    }
+    return this->evses.at(evse.value().id)->has_active_transaction();
 }
 
 bool operational_status_matches_connector_status(ConnectorStatusEnum connector_status,
                                                  OperationalStatusEnum operational_status) {
-    return (operational_status == OperationalStatusEnum::Inoperative and
-            connector_status == ConnectorStatusEnum::Unavailable) or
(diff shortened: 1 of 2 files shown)


lib/ocpp/v201/charge_point.cpp
```diff
@@ -889,17 +889,59 @@ void ChargePoint::update_aligned_data_interval() {
...
+            }
+        }
+        return false;
+    }
+
+    if (this->evses.at(evse.value().id)->has_active_transaction()) {
+        return true;
+    } else {
+        return false;
+    }
```
marcemmers, line 906:
```suggestion
    return this->evses.at(evse.value().id)->has_active_transaction();
```

lib/ocpp/v201/charge_point.cpp
```diff
@@ -889,17 +889,59 @@ void ChargePoint::update_aligned_data_interval() {
...
+}
+
+bool operational_status_matches_connector_status(ConnectorStatusEnum connector_status,
+                                                 OperationalStatusEnum operational_status) {
+    return (operational_status == OperationalStatusEnum::Inoperative and
+            connector_status == ConnectorStatusEnum::Unavailable) or
+           (operational_status == OperationalStatusEnum::Operative and
+            (connector_status != ConnectorStatusEnum::Unavailable or connector_status != ConnectorStatusEnum::Faulted));
```
marcemmers, line 914: This last statement is always true since it can't be Unavailable and Faulted at the same time.

Also, maybe easier to read using a switch for one of the enums?

marcemmers (changes requested):

marcemmers (approved):

PR #217, opened 2023-10-16
added is_integer and is_decimal functions to device model validation

Pietfried:


lib/ocpp/v201/device_model.cpp
```diff
@@ -10,6 +10,45 @@ namespace ocpp {
...
+static bool is_decimal_number(const std::string& value) {
+    if (value.empty()) {
+        return false;
+    }
+
+    // Check for + or - in the beginning
+    size_t start_pos = 0;
+    if (value[0] == '+' || value[0] == '-')
+        start_pos = 1;
+
```
marcemmers, line 36:
```suggestion
    if (value[0] == '+' || value[0] == '-') {
        start_pos = 1;
    }
```

lib/ocpp/v201/device_model.cpp
```diff
@@ -34,61 +73,77 @@ bool DeviceModel::component_criteria_match(const Component& component,
...
+                    if (std::find(values_list.begin(), values_list.end(), v) == values_list.end()) {
+                        return false;
+                    }
                 }
+                return true;
             }
-            return true;
         }
+    } catch (const std::exception& e) {
+        return false;
```
marcemmers, line 146: Should we print here what failed so it's at least not silent?

Pietfried: Yes we could, although in this function we dont have a lot of context that we can use (no component and variable names)

marcemmers: What about moving the whole try catch to the next layer then? The exception would tell something about what went wrong at least.

Pietfried: Yep Ill have a look and add this

marcemmers (approved):

PR #218, opened 2023-10-18
B06:- Get variables requests 

SNSubramanya:
Returns a rejected as a response to get variable request for write only variables

### DIFF
--- a/lib/ocpp/v201/device_model.cpp
+++ b/lib/ocpp/v201/device_model.cpp
@@ -114,7 +114,7 @@ GetVariableStatusEnum DeviceModel::request_value_internal(const Component& compo
     }
 
     // only internal functions can access WriteOnly variables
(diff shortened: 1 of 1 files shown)


lib/ocpp/v201/charge_point.cpp
```diff
@@ -719,9 +719,9 @@ void ChargePoint::handle_message(const EnhancedMessage<v201::MessageType>& messa
 
 void ChargePoint::message_callback(const std::string& message) {
     auto enhanced_message = this->message_queue->receive(message);
+    enhanced_message.message_size = message.size();
```
marcemmers, line 722: Just wondering if we would like to do this assignment in the queue itself so 1.6 can use it as well.

marcemmers (approved):

PR #222, opened 2023-10-23
fix crash invalid evseid remote start transaction

rinzevdwalAlfen:


lib/ocpp/v201/charge_point.cpp
```diff
@@ -1994,9 +1994,10 @@ void ChargePoint::handle_remote_start_transaction_request(Call<RequestStartTrans
     // Check if evse id is given.
     if (msg.evseId.has_value()) {
         const int32_t evse_id = msg.evseId.value();
-        const auto& evse = this->evses.at(evse_id);
+        const auto itEvse{this->evses.find(evse_id)};
```
Pietfried, line 1997: nit: use snake_case

Pietfried (approved):

PR #223, opened 2023-10-23
websockets: `Uri` as single-source, incl. chargepoint-ID; conforming security-profile

Dominik-K:


include/ocpp/common/types.hpp
```diff
@@ -558,6 +558,16 @@ firmware_status_notification_to_firmware_status_enum_type(const FirmwareStatusNo
 
 } // namespace conversions
 
+namespace security {
+// The security profiles defined in OCPP 2.0.1 resp. in the OCPP 1.6 security-whitepaper.
+enum SecurityProfiles : int {
```
Pietfried, line 563: ALL_CAPS is best practice for C++ enums. The use of `: int` is redundant because by default enums are of type int. So I would go with

`enum SecurityProfiles {
    UNSPECIFIED = 0,
    UNSECURED_TRANSPORT_WITH_BASIC_AUTHENTICATION = 1,
    TLS_WITH_BASIC_AUTHENTICATION = 2,
    TLS_WITH_CLIENT_SIDE_CERTIFICATES = 3,
};`

Pietfried (changes requested):

include/ocpp/common/types.hpp
marcemmers: Why not use an enum class in that case?

PR #224, opened 2023-10-23
Enable the connect callback to update the connection state variable

bWF0dGhpYXMK:
After a reconnect, the connection call back is retriggered to inform the upper business logic that the connection is valid again. A boolean is used to avoid the reinit of the Message queue.


hikinggrass:
Implemented in #238

PR #232, opened 2023-10-27
callback for boot notification

SNSubramanya:


lib/ocpp/v201/charge_point.cpp
```diff
@@ -1444,6 +1445,9 @@ void ChargePoint::handle_boot_notification_response(CallResult<BootNotificationR
             },
             retry_interval);
     }
+
+    // call the registered boot notification callback
+    callbacks.boot_notification_callback.value()(this->registration_status);
```
marcemmers, line 1450: What would now happen if the callback does not have a value?

marcemmers (changes requested):

marcemmers (approved):

PR #234, opened 2023-10-30
ChangeAvailability(Scheduled) moves inactive EVSE/connectors to unavailable directly

Pietfried:
now moving evse and connectors without an active transaction to unavailable when a ChangeAvailability.req is scheduled

### DIFF
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -2107,7 +2107,8 @@ void ChargePoint::handle_change_availability_req(Call<ChangeAvailabilityRequest>
         evse_id = msg.evse.value().id;
     }
 
-    if (!transaction_active or is_already_in_state) {
+    if (!transaction_active or is_already_in_state or
+        (evse_id == 0 and msg.operationalStatus == OperationalStatusEnum::Operative)) {
         response.status = ChangeAvailabilityStatusEnum::Accepted;
         // remove any scheduled availability request in case no transaction is scheduled or the component is already in
         // correct state to override possible requests that have been scheduled before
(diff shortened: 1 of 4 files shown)


lib/ocpp/v201/charge_point.cpp
```diff
@@ -2126,6 +2127,31 @@ void ChargePoint::handle_change_availability_req(Call<ChangeAvailabilityRequest>
     if (!transaction_active) {
         // execute change availability if possible
         this->callbacks.change_availability_callback(msg, true);
+    } else if (response.status == ChangeAvailabilityStatusEnum::Scheduled) {
+        if (evse_id == 0) {
+            // put all EVSEs to unavailable that do not have active transaction
```
marcemmers, line 2132: Shouldn't we put the station to unavailable instead of all the separate EVSEs?

See the note here:
image

Pietfried: This case is there to meet the requirement G04.FR.07 . The charging station itself will be set to unavailable once no transaction is active anymore (scheduled). The persist flag indicates that the operational state for the EVSE should not persist reboots and that the EVSE should return to its previous operational state when the ChargingStation availability state is changed.

Right now we're using the change availbility callback not only for changing the availability (operative/inoperative) but also for enabling/disabling EVSE/connectors when a higher instances availability is changed. This makes it difficult to read and to handle also for the application that uses libocpp.

I think we should have different callbacks for:
1. setting the operational state (Operative/Inoperative for ChargingStation, EVSE and Connector)
2. Enabling/Disabling Connectors (e.g. setting Connectors to unavailable)
because its a different thing and mixed up in libocpp right now because there is only one callback and it moves proper handling for this into the application that uses libocpp

We have a working implementation for the current setup within everest-core, but I tend towards splitting up the callbacks at some point as described above.

marcemmers (approved):

PR #235, opened 2023-10-30
Move disconnected_time to websocket

Pietfried:
moved disconnected time to websocket, because closed_callback isnt always executed when the websocket disconnects / fails. The websocket attempts to reconnect independently without executing the closed callback in certain situations


Pietfried:
Lifetime of  the websocket instance was not suitable for this approach, because a new websocket object can be instantiated on reconnects. This PR is replaced by: https://github.com/EVerest/libocpp/pull/238

PR #236, opened 2023-10-30
Add check for decimal and integer values when set in device model

maaikez:


maaikez:
Duplicate of https://github.com/EVerest/libocpp/pull/217

PR #238, opened 2023-11-01
Added disconnected callback to websocket to track disconnected_time (for OfflineThreshold) correctly

Pietfried:
closed_callback isn't always executed when the websocket disconnects / fails. The websocket attempts to reconnect independently without executing the closed callback in certain situations. Therefore a disconnected_callback is added to the websocket in order to notify the chargepoint in case of a disconnect, so that the disconnected_time can be tracked properly


marcemmers (approved):
Looks good! Also nice to see we could simplify by using the status of the websocket itself :+1:

PR #241, opened 2023-11-02
Averaging Measurements

SNSubramanya:


marcemmers (changes requested):
Can you please take another look to remove all the unrelated changes, debug stuff etc?

Also take a closer look at the mutex uses especially since the AlignedData has its own mutex now.

I am wondering if AlignedData is the best name. It says something currently about what it is used for, not what it is. I normally think that naming something what it is, is a bit clearer and allows for more general use of classes.

lib/ocpp/v201/evse.cpp
```diff
@@ -221,6 +221,16 @@ MeterValue Evse::get_meter_value() {
     return this->meter_value;
 }
 
+MeterValue Evse::get_idle_meter_value() {
+    std::lock_guard<std::recursive_mutex> lk(this->meter_value_mutex);
```
marcemmers, line 225: Still some mutexes here that are no longer needed ;)

lib/ocpp/v201/evse.cpp
```diff
@@ -112,16 +112,13 @@ void Evse::open_transaction(const std::string& transaction_id, const int32_t con
     if (aligned_data_tx_updated_interval > 0s) {
         transaction->aligned_tx_updated_meter_values_timer.interval_starting_from(
             [this] {
-                if (this->device_model.get_optional_value<bool>(ControllerComponentVariables::AlignedDataSendDuringIdle)
```
marcemmers, line 115: We should probably still check on the AlignedDataSendDuringIdle flag

lib/ocpp/v201/charge_point.cpp
```diff
@@ -986,8 +979,9 @@ void ChargePoint::update_aligned_data_interval() {
                 }
             }
 
+            // send evseID = 0 values
             const auto meter_value =
-                get_latest_meter_value_filtered(this->get_meter_value(), ReadingContextEnum::Sample_Clock,
+                get_latest_meter_value_filtered(this->aligned_data_evse0.get_values(), ReadingContextEnum::Sample_Clock,
```
marcemmers, line 984: Are you clearing the evse0 averaging as well?

Maybe you can still do the clearing in the get so you don't always have to group a `get` and `clear` function together. Only `get` is probably not the best name then since you will be changing the internal state which you normally don't do on a get

lib/ocpp/v201/average_meter_values.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+void AverageMeterValues::clear_values() {
+    EVLOG_debug << " Clearing the aligned values";
+    // reset the values
+    this->aligned_meter_values.clear();
+    this->averaged_meter_values.sampledValue.clear();
+}
+
+void AverageMeterValues::set_values(const MeterValue& meter_value) {
+    EVLOG_debug << " Setting the aligned values";
```
marcemmers, line 22: Maybe remove these log lines? Even for debugging this is not really useful IHMO. You'd just get spammed with this line, maybe even a couple of times a second with a fast meter.

At the very least remove the log from the constructor. It is assumed a constructor just works so does not really need logging like this.

lib/ocpp/v201/average_meter_values.cpp
```diff
@@ -0,0 +1,72 @@
...
+                        << " num_ele"
+                        << this->aligned_meter_values[{element.measurand.value(), element.phase.value()}].num_elements;
+            EVLOG_debug << "AVG: " << element.value;
+        }
+    }
+}
+bool AverageMeterValues::is_avg_meas(const SampledValue& sample) {
+
+    // TODO: check up on location values and how they impact the averaging
+    if (sample.measurand.has_value() && sample.phase.has_value()) {
```
marcemmers, line 63: What if we have a single phase system and there is no phase set? Or we want to average the total power value which also does not have a phase?

I think a more close look is needed at this still

marcemmers (changes requested):

include/ocpp/v201/average_meter_values.hpp
```diff
@@ -0,0 +1,55 @@
...
+    };
+    struct MeterValueMeasurands {
+        MeasurandEnum measurand;
+        std::optional<PhaseEnum> phase;       // measurand may or may not have a phase field
+        std::optional<LocationEnum> location; // measurand may or may not have location field
+
+        // Define a comparison operator for the struct
+        bool operator<(const MeterValueMeasurands& other) const {
+
+            return measurand < other.measurand || location.value() < other.location.value() ||
```
marcemmers, line 42: Did you test this with measurands that don't have a location or a phase?

Also, they prefer `or` to `||`

lib/ocpp/v201/average_meter_values.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+void AverageMeterValues::set_values(const MeterValue& meter_value) {
+    std::lock_guard<std::mutex> lk(this->avg_meter_value_mutex);
+    // store all the meter values in the struct
+    this->averaged_meter_values = meter_value;
+
+    // avg all the possible measurerands
+    for (auto element : meter_value.sampledValue) {
+        if (is_avg_meas(element)) {
+            MeterValueCalc temp =
```
marcemmers, line 25: If you make this a reference you can edit the value inside the map instead of having to copy it twice

lib/ocpp/v201/average_meter_values.cpp
```diff
@@ -0,0 +1,66 @@
...
+bool AverageMeterValues::is_avg_meas(const SampledValue& sample) {
+
+    // TODO: check up on location values and how they impact the averaging
+    if (sample.measurand.has_value()) {
+        if ((sample.measurand == MeasurandEnum::Current_Import) || (sample.measurand == MeasurandEnum::Voltage) ||
+            (sample.measurand == MeasurandEnum::Power_Active_Import) || (sample.measurand == MeasurandEnum::Frequency))
+            return true;
+    } else {
+        return false;
+    }
```
marcemmers, line 63: I'm assuming you'd get a warning of this as well, but you have a return without value path here I think?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -1114,12 +1109,14 @@ void ChargePoint::update_aligned_data_interval() {
...
                 const auto meter_value =
-                    get_latest_meter_value_filtered(evse->get_meter_value(), ReadingContextEnum::Sample_Clock,
+                    get_latest_meter_value_filtered(evse->get_idle_meter_value(), ReadingContextEnum::Sample_Clock,
                                                     ControllerComponentVariables::AlignedDataMeasurands);
 
                 if (!meter_value.sampledValue.empty()) {
                     // J01.FR.14 this is the only case where we send a MeterValue.req
                     this->meter_values_req(evse_id, std::vector<ocpp::v201::MeterValue>(1, meter_value));
+                    // clear the values
+                    evse->clear_idle_meter_values();
```
marcemmers, line 1119: Shouldn't we clear regardless of if we actually have filtered data?

lib/ocpp/v201/evse.cpp
```diff
@@ -221,6 +225,15 @@ MeterValue Evse::get_meter_value() {
     return this->meter_value;
 }
 
+MeterValue Evse::get_idle_meter_value() {
+    return this->aligned_data_updated.retrieve_processed_values();
+}
+
+void Evse::clear_idle_meter_values() {
+    std::lock_guard<std::recursive_mutex> lk(this->meter_value_mutex);
```
marcemmers, line 233: No mutex needed here right?

lib/ocpp/v201/average_meter_values.cpp
```diff
@@ -0,0 +1,73 @@
...
+            if (element.phase.has_value()) {
+
+                EVLOG_debug << " phase: " << element.phase.value();
+            }
+            if (element.location.has_value()) {
+
+                EVLOG_debug << " loc: " << element.location.value();
+            }
+
+            EVLOG_debug << " sum:" << temp.sum << "#num" << temp.num_elements << "AVG: " << element.value;
```
marcemmers, line 59: Can we please remove these debug lines? I get that they are useful to test with. But this makes it impossible to enable debug on a system without getting spammed. And now it is working we can remove these right?

marcemmers (approved):

PR #246, opened 2023-11-06
Support of security profile for example charge_point

bWF0dGhpYXMK:


src/charge_point.cpp
```diff
@@ -100,8 +100,42 @@ int main(int argc, char* argv[]) {
...
+        fs::create_directories(ca_v2g_path);
+    }
+
+    const fs::path ca_mo_path = "/tmp/certs/ca/mo/";
+    if (!fs::exists(ca_mo_path)) {
+        fs::create_directories(ca_mo_path);
+    }
+
+    if (!fs::is_regular_file("/tmp/certs/ca/v2g/V2G_CA_BUNDLE.pem")) {
+        std::ofstream("/tmp/certs/ca/v2g/V2G_CA_BUNDLE.pem");
```
barsnick, line 119: See also the conditional use of `boost::filesystem::ofstream` in libevse-security:

https://github.com/EVerest/libevse-security/blob/c6e4f7e22363cd43bf3ee4c3930356dae2dafd43/include/support_older_cpp_versions.hpp#L13-L19

https://github.com/EVerest/libevse-security/blob/fb097c945812a738def096e44a66a895b2c052a0/include/evse_utilities.hpp#L54

(I'm not sure whether that's important to use, and if it doesn't just map to `std::ofstream` anyway.)

src/charge_point.cpp
```diff
@@ -100,8 +100,42 @@ int main(int argc, char* argv[]) {
         fs::create_directories(cso_path);
     }
 
+    const fs::path csms_path = "/tmp/client/csms";
+    if (!fs::exists(csms_path)) {
+        fs::create_directories(csms_path);
```
barsnick, line 105: Shouldn't you explicitly include `ocpp/common/support_older_cpp_versions.hpp` for the `fs` namespace? It seems it is implicitly included via some other `#include`. (I don't know about EVerest's code policy on this, feel free to clarify).

bWF0dGhpYXMK: Thank you for your comments; yes, it is included over the charge_point.hpp, i have no clear opinion on that.

hikinggrass (approved):

src/charge_point.cpp
Dominik-K: @barsnick 
> implicitly included via some other #include

I'm in favour to explicitly include them. Just found the IWYU (https://include-what-you-use.org/) tool which checks this (at least it should, don't have own experience with it, yet).

PR #250, opened 2023-11-07
Add timers for certificate expiration check (A03)

klemmpnx:
Add timers for the renewal of charging station certificates (use case A03). Oriented at OCPP1.6 implementation.

Note: requires update of libevse-security to correctly return remaining valid days (see https://github.com/EVerest/libevse-security/pull/14/files)

Tested in https://github.com/PionixInternal/ocpp-testing/pull/34


klemmpnx:
One concern of mine: I added the initialization of the timers to both the `handle_boot_notification_response` as well the `connected_callback` of the websocket (here conditioned though if there was a "offline" time). The latter is necessary I think because the timers need to be stopped at any disconnect.  
However, I am not sure when exactly the `this->registration_status == RegistrationStatusEnum::Accepted` condition is fulfilled and it might be more useful to just start the timers every time (unconditioned) in the `connected_callback` instead of the boot_notification_response?

Pietfried:
> One concern of mine: I added the initialization of the timers to both the `handle_boot_notification_response` as well the `connected_callback` of the websocket (here conditioned though if there was a "offline" time). The latter is necessary I think because the timers need to be stopped at any disconnect. However, I am not sure when exactly the `this->registration_status == RegistrationStatusEnum::Accepted` condition is fulfilled and it might be more useful to just start the timers every time (unconditioned) in the `connected_callback` instead of the boot_notification_response?

It is not allowed to send this message to the CSMS before the BootNotification.req has been confirmed with a BootNotification.conf(Accepted) from the CSMS (== RegistrationStatus::Accepted), so the implementation is just fine

lib/ocpp/v201/charge_point.cpp
```diff
@@ -714,6 +721,19 @@ void ChargePoint::init_websocket() {
     this->websocket->register_message_callback([this](const std::string& message) { this->message_callback(message); });
 }
 
+void ChargePoint::init_certificate_expiration_check_timers() {
+
+    // Timers started with initial delays; callback functions are supposed to re-schedule on their own!
```
Pietfried, line 726: We only need a client certificate in case we use SecurityProfile 3, else I think we dont need to request it.
We only want to request a V2G certificate in case ISO15118V2GCertificateInstallationEnabled .

Pietfried (changes requested):
Please check again in which situations we should request a V2G or client certificate and only set the timers if required

Pietfried (approved):

Pietfried (approved):

PR #258, opened 2023-11-10
Harden change configuraton on integer input

bWF0dGhpYXMK:
For integer related changeConfiguration: Out-of-range exceptions are rejected and values like "01asdfasdf" with non-integer suffix are rejected.


hikinggrass:
Implemented in #261

PR #260, opened 2023-11-12
Added PUMP Connect as compatible CSMS for OCPP 1.6

krealyt:


RobertDeLeeuw (commented):
I think it should be: PUMP (PUMP Connect)
PUMP is the company, the CPMS is PUMP Connect. Or as the name of the company is in the name of the CPMS we do only PUMP Connect? Both work for me.

PR #264, opened 2023-11-15
Permanently disconnecting websocket in v201 when stop function is called

Pietfried:


lib/ocpp/v201/charge_point.cpp
```diff
@@ -164,7 +164,7 @@ void ChargePoint::stop() {
     this->websocket_timer.stop();
     this->client_certificate_expiration_check_timer.stop();
     this->v2g_certificate_expiration_check_timer.stop();
-    this->disconnect_websocket(websocketpp::close::status::going_away);
```
Dominik-K, line 167: Yes, this status was wrong (since the beginning), according to[ `websocketpp`'s documentation](https://github.com/zaphoyd/websocketpp/blob/1b11fd301531e6df35a6107c1e8665b1e77a2d8e/websocketpp/close.hpp#L78-L80):

```cpp
 /// The endpoint was "going away", such as a server going down or a browser
/// navigating away from a page.
static value const going_away = 1001;
```

Dominik-K (approved):

mennodegraaf:
Looks good to me, I have tested before that a disconnect with `status::normal` does not result in a reconnect

mennodegraaf (approved):

mennodegraaf:
For future reference: there are three issues when calling `ChargePoint::stop()`, only when the websocket is connected, when it is unconnected it is ok:

- It tries to reconnect after gracefully closing the websocket
- On close, it logs a message about a failure to open
- The database does not gracefully close, see the error message above about 'unable to close due to unfinalized statements'

PR #268, opened 2023-11-16
Make sure a transaction cache update also sets the lifetime

marcemmers:


lib/ocpp/v201/charge_point.cpp
```diff
@@ -573,6 +573,7 @@ AuthorizeResponse ChargePoint::validate_token(const IdToken id_token, const std:
     response = this->authorize_req(id_token, certificate, ocsp_request_data);
 
     if (auth_cache_enabled) {
+        update_id_token_cache_lifetime(response.idTokenInfo);
```
Pietfried, line 576: nit: just to be consistent. We are using `this` everywhere
```suggestion
        this->update_id_token_cache_lifetime(response.idTokenInfo);
```

Pietfried (approved):

lib/ocpp/v201/charge_point.cpp
marcemmers: Yep, good call, missed that one. Please keep these nit's coming ;)

PR #276, opened 2023-11-22
Do not send InstallScheduled when changing connectors to unavailable

hikinggrass:
This was accidentally sent all the time but is only valid if there is a transaction ongoing

Fix verified against OCTT tests
    TC_L_01_CS
    TC_L_02_CS
    TC_L_03_CS
    TC_L_05_CS
    TC_L_06_CS
    TC_L_07_CS
    TC_L_08_CS
    TC_L_10_CS
    TC_L_15_CS
    TC_L_18_CS

### DIFF
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -220,11 +220,6 @@ void ChargePoint::on_firmware_update_status_notification(int32_t request_id,
     }
 
     if (this->firmware_status_before_installing == req.status) {
-        this->firmware_status = FirmwareStatusEnum::InstallScheduled;
-        req.status = firmware_status;
-        ocpp::Call<FirmwareStatusNotificationRequest> call(req, this->message_queue->createMessageId());
-        this->send_async<FirmwareStatusNotificationRequest>(call);
-
         this->change_all_connectors_to_unavailable_for_firmware_update();
     }
 }


lib/ocpp/v201/charge_point.cpp
```diff
@@ -220,11 +220,13 @@ void ChargePoint::on_firmware_update_status_notification(int32_t request_id,
...
 
     if (this->firmware_status_before_installing == req.status) {
-        this->firmware_status = FirmwareStatusEnum::InstallScheduled;
-        req.status = firmware_status;
-        ocpp::Call<FirmwareStatusNotificationRequest> call(req, this->message_queue->createMessageId());
-        this->send_async<FirmwareStatusNotificationRequest>(call);
-
+        const auto transaction_active = this->any_transaction_active(std::nullopt);
+        if (transaction_active) {
+            this->firmware_status = FirmwareStatusEnum::InstallScheduled;
```
maaikez, line 225: I will test your solution. But as we are already sending InstallScheduled from our 'core', will that go well?

Anyway, will test, let you know!

hikinggrass: That might be an issue, maybe we can use another method of determining if transactions are still running that doesn't involve this part of the code in libocpp

maaikez: This is indeed failing L_15. Because InstallScheduled is already sent (from the 'core' application), it will send it again.

It is a bit strange that all firmware statusses are sent from 'core' and only one is sent from here isn't it?
But keeping a variable that stores the last sent status will probably do the trick as well.

hikinggrass: You're right, this isn't the proper place for it. I'll remove the commit from this PR and figure something out for L_15

maaikez (approved):

PR #277, opened 2023-11-23
Ensure firmware update status notifications are not sent multiple times

hikinggrass:
Additionally send InstallScheduled if transactions are still ongoing


lib/ocpp/v201/charge_point.cpp
```diff
@@ -192,6 +192,13 @@ void ChargePoint::disconnect_websocket(websocketpp::close::status::value code) {
 
 void ChargePoint::on_firmware_update_status_notification(int32_t request_id,
                                                          const FirmwareStatusEnum& firmware_update_status) {
+    if (this->firmware_status == firmware_update_status) {
+        if (request_id == -1 or
+            this->firmware_status_id.has_value() and this->firmware_status_id.value() == request_id) {
+            // already sent, do not send again
+            return;
+        }
+    }
```
valentin-dimov, line 201: This checks if the request would not change the current status, and has the same request ID (i.e. refers to the same firmware update process) as the last one we received. What is the meaning of `request_id` being -1?

valentin-dimov: Also, are we guaranteed to only ever have one firmware update process (i.e. only one valid `request_id`) valid at any time? This check wouldn't work if we're receiving updates for several concurrent update processes.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -220,6 +227,13 @@ void ChargePoint::on_firmware_update_status_notification(int32_t request_id,
     }
 
     if (this->firmware_status_before_installing == req.status) {
+        const auto transaction_active = this->any_transaction_active(std::nullopt);
+        if (transaction_active) {
+            this->firmware_status = FirmwareStatusEnum::InstallScheduled;
+            req.status = firmware_status;
+            ocpp::Call<FirmwareStatusNotificationRequest> call(req, this->message_queue->createMessageId());
+            this->send_async<FirmwareStatusNotificationRequest>(call);
+        }
```
valentin-dimov, line 236:
```suggestion
        // FIXME(Kai): This is a temporary workaround, because the EvseManager does not keep track of transactions.
        // If the firmware status is expected to become "Installing", but we still have a transaction running,
        // the update will wait for the transaction to finish, and so we send an "InstallScheduled" status.
        // This is necessary for OCTT TC_L_15_CS to pass.
        const auto transaction_active = this->any_transaction_active(std::nullopt);
        if (transaction_active) {
            this->firmware_status = FirmwareStatusEnum::InstallScheduled;
            req.status = firmware_status;
            ocpp::Call<FirmwareStatusNotificationRequest> call(req, this->message_queue->createMessageId());
            this->send_async<FirmwareStatusNotificationRequest>(call);
        }
```

lib/ocpp/v201/charge_point.cpp
hikinggrass: The System module in everest-core currently ensures there's only one firmware update in progress at a given time, this seems to be the idea in OCPP 2.0.1 as well (see Figure 119 for the state machine of the firmware update which strongly suggests only one update at the time)

hikinggrass:
rebased and squashed, should be ready to merge

PR #278, opened 2023-11-23
Fix clock aligned timestamps

SNSubramanya:


lib/ocpp/v201/utils.cpp
```diff
@@ -157,7 +157,22 @@ std::string sha256(const std::string& str) {
 std::string generate_token_hash(const IdToken& token) {
     return sha256(conversions::id_token_enum_to_string(token.type) + token.idToken.get());
 }
+ocpp::DateTime round_to_x_seconds(const DateTime& timestamp, std::chrono::seconds x) {
```
valentin-dimov, line 160: I'd add add a sanity check here to confirm that `x` is positive.

valentin-dimov (commented):
The idea here is that libocpp will round down the timestamps on aligned meter values down before sending them to the CSMS, right? The spec states that the intention here is supplying meter values from certified meters to grid suppliers, and we can't fudge the timestamps on signed meter values like this. Is it okay if the two mismatch?

include/ocpp/v201/utils.hpp
```diff
@@ -52,6 +52,8 @@ std::string sha256(const std::string& str);
 /// @return A SHA256 hash string
 std::string generate_token_hash(const IdToken& token);
 
+ocpp::DateTime round_to_x_seconds(const DateTime& timestamp, std::chrono::seconds x);
```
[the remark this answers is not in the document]
marcemmers: You can also just pass DateTime by value

lib/ocpp/v201/evse.cpp
```diff
@@ -135,6 +142,13 @@ void Evse::open_transaction(const std::string& transaction_id, const int32_t con
                 for (auto& item : meter_value.sampledValue) {
                     item.context = ReadingContextEnum::Sample_Clock;
                 }
+                // Get the aligned timestamp
+                auto old_time = meter_value.timestamp;
+                auto interval_value = std::chrono::seconds(
```
marcemmers, line 147: You could also capture `aligned_data_tx_ended_interval` and then use that as the interval. Same for the other lambda

lib/ocpp/v201/utils.cpp
```diff
@@ -157,7 +157,22 @@ std::string sha256(const std::string& str) {
...
+    auto timestamp_sys = date::utc_clock::to_sys(timestamp.to_time_point());
 
+    // get the current midnight
+    auto midnight = std::chrono::floor<date::days>(timestamp_sys);
+    auto seconds_since_midnight = std::chrono::duration_cast<std::chrono::seconds>(timestamp_sys - midnight).count();
```
marcemmers, line 165: My suggestion is to remove all of the count kind of functions and stay within the chrono types

lib/ocpp/v201/utils.cpp
```diff
@@ -157,7 +157,22 @@ std::string sha256(const std::string& str) {
...
+    auto rounded_seconds = ((seconds_since_midnight + x.count() / 2) / x.count()) * x.count();
```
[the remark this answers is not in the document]
valentin-dimov: Fair enough, so usually we'd be rounding up (since libocpp gets the last published powermeter at the edges of intervals).

PR #279, opened 2023-11-23
OCPP2.0.1 add data transfer in line with OCPP1.6 implementation

hikinggrass:


lib/ocpp/v201/charge_point.cpp
```diff
@@ -2783,6 +2783,7 @@ void ChargePoint::handle_data_transfer_req(Call<DataTransferRequest> call) {
     DataTransferResponse response;
     const auto vendor_id = msg.vendorId.get();
     const auto message_id = msg.messageId.value_or(CiString<50>()).get();
+    // first try the callbacks that are explicitly registered for a vendorId or messageId
```
valentin-dimov, line 2786: I can't find any place where the specific `data_transfer_callbacks` are set. Am I missing something or is the code block below redundant?

hikinggrass: You're correct, this was totally redundant code that I've removed

lib/ocpp/v201/charge_point.cpp
```diff
@@ -2783,31 +2783,22 @@ void ChargePoint::handle_customer_information_req(Call<CustomerInformationReques
...
-                   this->data_transfer_callbacks[vendor_id].count(message_id) == 0) {
-            response.status = ocpp::v201::DataTransferStatusEnum::UnknownMessageId;
-        } else {
-            // there is a callback registered for this vendorId and messageId
-            response = this->data_transfer_callbacks[vendor_id][message_id](msg.data);
-        }
+
+    if (this->callbacks.data_transfer_callback.has_value()) {
+        response = this->callbacks.data_transfer_callback.value()(call.msg);
     }
```
valentin-dimov, line 2789: What should the response be in case the `data_transfer_callback` is empty? Currently it seems you'd just return a default-constructed response, but I'm not sure this is the right way to do that...

valentin-dimov (changes requested):

lib/ocpp/v201/charge_point.cpp
hikinggrass: UnknownVendorId seems appropriate here

PR #281, opened 2023-11-24
Support of mutex inside get/set of configuration, which can be happen…

bWF0dGhpYXMK:
…, when getConfiguration is called from outside, during a changeconfiguration processing.


Pietfried (changes requested):
Hi, thanks for bringing this up and sorry for the the late feedback. I would rather like to avoid the recursive mutex to reduce complexity. This could be done with a small refactor of the private/public functions for setting and getting the custom configuration keys. This might be a helpful resource: https://stackoverflow.com/a/14600868

Pietfried (approved):

PR #282, opened 2023-11-24
If no changeconfiguration callback is register call the generic one, …

bWF0dGhpYXMK:
…if present


lib/ocpp/v16/charge_point.cpp
```diff
@@ -280,6 +280,11 @@ void ChargePoint::register_configuration_key_changed_callback(
     this->charge_point->register_configuration_key_changed_callback(key, callback);
 }
 
+void ChargePoint::register_configuration_key_changed_callback(
+    const std::function<void(const CiString<50>& key, const CiString<500>& valuee)>& callback) {
```
Pietfried, line 284:
```suggestion
    const std::function<void(const CiString<50>& key, const CiString<500>& value)>& callback) {
```

Pietfried, line 284:
```suggestion
    const std::function<void(const CiString<50>& key, const CiString<500>& value)>& callback) {
```

include/ocpp/v16/charge_point.hpp
```diff
@@ -436,6 +436,12 @@ class ChargePoint {
     void register_configuration_key_changed_callback(const CiString<50>& key,
                                                      const std::function<void(const KeyValue& key_value)>& callback);
 
+    /// \brief registers a \p callback function that can be used to react on changed configuration key. This
+    /// callback is called when a configuration key has been changed by the CSMS
+    /// \param callback executed when this configuration key changed
+    void register_configuration_key_changed_callback(
```
Pietfried, line 442:
```suggestion
    void register_generic_configuration_key_changed_callback(
```

include/ocpp/v16/charge_point.hpp
```diff
@@ -436,6 +436,12 @@ class ChargePoint {
     void register_configuration_key_changed_callback(const CiString<50>& key,
                                                      const std::function<void(const KeyValue& key_value)>& callback);
 
+    /// \brief registers a \p callback function that can be used to react on changed configuration key. This
```
Pietfried, line 439: Please adjust the method docs so to point out that this is a generic callback and that it will only be called in case no specific callback for the key is registered.

include/ocpp/v16/charge_point.hpp
Pietfried: Please adjust other respective members accordingly.

Pietfried (changes requested):

Pietfried (approved):

PR #283, opened 2023-11-28
Feat: Delay the resumption of the message queue after reconnect

valentin-dimov:
Some OCTT tests (as well as real CSMS) send certain requests to the charging station after a reconnect (e.g. checking transaction status) and expect the charging station to respond to them before emptying its message queue. 

This PR adds a configurable delay between a reconnect and the resumption of the message queue. During this delay, the charging station can still respond to requests from the CSMS, but any outgoing requests will only be sent after the delay passes.


include/ocpp/common/message_queue.hpp
```diff
@@ -579,18 +596,27 @@ template <typename M> class MessageQueue {
...
+        this->resume_timer.stop();
         this->paused = true;
         this->cv.notify_one();
         EVLOG_debug << "pause() notified message queue";
     }
 
     /// \brief Resumes the message queue
-    void resume() {
-        EVLOG_debug << "resume()";
+    void resume(unsigned int delay_seconds = 0) {
```
marcemmers, line 607: Suggestion: Use chrono type here so it is perfectly clear what the unit is

include/ocpp/common/message_queue.hpp
```diff
@@ -154,6 +161,16 @@ template <typename M> class MessageQueue {
         EVLOG_debug << "Notified message queue worker";
     }
 
+    // Unlike the public resume(), this doesn't schedule anything - it just does the actual resumption
+    // Does not lock this->message_mutex, make sure to do that before calling
```
marcemmers, line 165: Suggestion: Change to recursive mutex and just lock here as well if that is needed.

valentin-dimov: Done

include/ocpp/common/message_queue.hpp
valentin-dimov: Done

hikinggrass (approved):

PR #285, opened 2023-11-29
EV369b:  drop update transactional messages if necessary

klemmpnx:
Part II of EV-369: drop transaction update messages if necessary:

- Slight refactoring if ControlMessage template 
- Heuristic implementation to drop every second update message (does not consider transaction ids, but should be fine for a realistic use case; stop/end transaction events are never dropped) ; behavior is also be explained by  unit test

### DIFF
--- a/tests/lib/ocpp/common/test_message_queue.cpp
+++ b/tests/lib/ocpp/common/test_message_queue.cpp
@@ -2,14 +2,25 @@
 #include <gtest/gtest.h>
 #include <nlohmann/json.hpp>
 #include <ocpp/common/message_queue.hpp>
+#include <ocpp/v16/messages/Authorize.hpp>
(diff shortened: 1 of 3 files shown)


tests/lib/ocpp/common/test_message_queue.cpp
```diff
@@ -2,14 +2,25 @@
 #include <gtest/gtest.h>
```
hikinggrass, line 2: Missing license header
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
#include <gtest/gtest.h>
```

hikinggrass (approved):
:+1:  pretty much just the missing license header

PR #286, opened 2023-11-29
Validate datetime in OCPP 2.0.1 device model

hikinggrass:


tests/utils_tests.cpp
```diff
@@ -0,0 +1,39 @@
...
+    ASSERT_FALSE(is_rfc3339_datetime("1"));
+    ASSERT_FALSE(is_rfc3339_datetime("1.1"));
+    ASSERT_FALSE(is_rfc3339_datetime("true"));
+    ASSERT_FALSE(is_rfc3339_datetime("abc"));
+
+    // more than 3 decimal digits are not allowed in OCPP
+    ASSERT_FALSE(is_rfc3339_datetime("2023-11-29T10:21:04.0001Z"));
+}
+
+} // namespace common
```
gsquires-tesla, line 38: All the tests passed ok?

hikinggrass:
All the new tests are passing on my machine

lib/ocpp/common/utils.cpp
```diff
@@ -73,4 +75,15 @@ bool is_decimal_number(const std::string& value) {
...
 }
 
+bool is_rfc3339_datetime(const std::string& value) {
+    static std::regex datetime_pattern{};
+    static std::once_flag datetime_regex_once;
+    std::call_once(datetime_regex_once, []() {
+        datetime_pattern =
+            std::regex{"\\d{4}-(?:0[1-9]|1[0-2])-(?:0[1-9]|[1-2]\\d|3[0-1])T(?:[0-1]\\d|2[0-3]):[0-5]\\d:["
+                       "0-5]\\d(?:\\.\\d{0,3}|)(?:Z|(?:\\+|\\-)(?:\\d{2}):?(?:\\d{2}))"};
+    });
```
valentin-dimov, line 85: Wouldn't it be simpler to define the regex as a static variable outside the function scope? See e.g. https://github.com/EVerest/libevse-security/commit/3879399059e179e4e1bd8800a3dab8217b90c886#diff-c38ad12c9c2a94bd81b62626c36ac3d2681ce61acf9aea46233135408e677ee1R12-R13

tests/utils_tests.cpp
valentin-dimov: They pass for me, too.

lib/ocpp/common/utils.cpp
valentin-dimov: Actually, you could probably leave it inside the function too, it should only be initialized once if it's static

valentin-dimov: Recap of discussion: This could cause bugs with certain older compilers, so we can leave it like this as far as I'm concerned.

PR #287, opened 2023-11-30
Fix: Validate CN = FQDN for TLS websockets connections

valentin-dimov:


lib/ocpp/common/websocket/websocket_tls.cpp
```diff
@@ -14,6 +14,41 @@
...
+        EVLOG_error << "Could not verify CSMS server certificate";
+        return false;
+    }
+
+    int depth = X509_STORE_CTX_get_error_depth(ctx.native_handle());
+
+    // only check for CSMS server certificate
+    if (depth == 0) {
+        // Get server certificate
+        X509* server_cert = X509_STORE_CTX_get_current_cert(ctx.native_handle());
```
hikinggrass, line 30: We should probably use the types from libevse-security here https://github.com/EVerest/libevse-security/blob/main/include/evse_security/detail/openssl/openssl_types.hpp

Dominik-K: How about moving this functionality as a neat helper function to `libevse-security` instead? 

Then all the OpenSSL-specifics here aren't part of `libocpp`. I mean in the midterm, that's the plan with `libevse-security`, right?

lib/ocpp/common/websocket/websocket_tls.cpp
```diff
@@ -200,6 +235,8 @@ tls_context WebsocketTLS::on_tls_init(std::string hostname, websocketpp::connect
         }
 
         context->set_verify_mode(boost::asio::ssl::verify_peer);
+        context->set_verify_callback(websocketpp::lib::bind(
+            &verify_csms_cn, hostname, websocketpp::lib::placeholders::_1, websocketpp::lib::placeholders::_2));
```
Dominik-K, line 239: `clang-tidy` reports here: "Prefer a lambda to std::bind" (see https://clang.llvm.org/extra/clang-tidy/checks/modernize/avoid-bind.html).

Do you have `clang-tidy` activated locally? E.g. with the `clangd` VSCode-extension.

hikinggrass (approved):

PR #288, opened 2023-11-30
Use sqlite_close_v2()

rinzevdwalAlfen:
Use `sqlite3_close_v2()` instead of `sqlite3_close().` Since we use the close in a destructor and and a destructor shoudn't fail. We should use the `sqlite3_close_v2()` instead since it will automatically deallocate the database connection after all prepared statements are finalized, all BLOB handles are closed, and all backups have finished.

[sqlite_close documentation](https://www.sqlite.org/c3ref/close.html)

### DIFF
--- a/lib/ocpp/v201/database_handler.cpp
+++ b/lib/ocpp/v201/database_handler.cpp
@@ -129,7 +129,7 @@ void DatabaseHandler::open_connection() {
 }
 
 void DatabaseHandler::close_connection() {
-    if (sqlite3_close(this->db) == SQLITE_OK) {
+    if (sqlite3_close_v2(this->db) == SQLITE_OK) {
         EVLOG_debug << "Successfully closed database: " << this->database_file_path;
     } else {
         EVLOG_error << "Error closing database file: " << sqlite3_errmsg(this->db);


valentin-dimov (commented):
To my understanding, this change would mean that we won't get error prints from the destructor, even if the database is busy - however, `sqlite_close_v2` will not wait for in-progress work to finish, it will simply arrange for the connection to be cleaned up whenever the work is done. If we call `close_connection` right before shutting down (which is, I assume, the main use-case), it seems to me that this could still lead to the process exiting with a dirty connection, since we never actually wait for the database connection to be destroyed.

valentin-dimov:
I'm not particularly familiar with sqlite3, but reading the doc page, I'm not sure that when it says the function "makes arrangements to automatically deallocate the database connection after all prepared statements are finalized [...]", this means the function blocks until that work is finished. Is it possible it'll return early and the work continues in the background?

valentin-dimov (approved):
Following our discussion: looks good, noting that we still need to look into why we were getting busy errors before, as those are still unexpected.

Dominik-K (approved):
The [`DatabaseHandler` deconstructor used the `sqlite3_close_v2()` until one month ago](https://github.com/EVerest/libocpp/pull/247/commits/ccf92de76ab7f561ac3bb7bf3b0ea52f15245832#diff-930d5fa7683959aa71af7865795d76f5081ecad62a36cdcb64b99bf6187caf54R21-R23) until PR
- https://github.com/EVerest/libocpp/pull/247

So it's ok to get back to this state.

However, it would be good to reflect this in OCPP 1.6, too. Seems that the `DatabaseHandler` itself could be a generic one.

Dominik-K:
> However, it would be good to reflect this in OCPP 1.6, too. Seems that the `DatabaseHandler` itself could be a generic one.

-> https://github.com/EVerest/libocpp/issues/359

PR #293, opened 2023-12-06
enum conversions from `common` to versions; methods now part of the enums

Dominik-K:
Part of EV-280

Got inspiration for this enum-methods encapsulation mainly from this [Stackoverflow answer](https://stackoverflow.com/a/53284026/1168315).


Dominik-K:
I'm also considering to move these enums to the respective `messages/*.cpp` as these are only used in this functions, AFAICS.

That will also lower the compile-time as the `enums.cpp`  gets smaller and a recompile isn't triggered as often as before.

PR #294, opened 2023-12-07
B08 get custom report

SNSubramanya:
This use case describes how the CSMS requests a Charging Station to send a report of all
Components and Variables limited to those that match ComponentCriteria and/or the list of
ComponentVariables. The result will be returned asynchronously in one or more
`NotifyReportRequest` messages.

### DIFF
--- a/config/v201/config.json
+++ b/config/v201/config.json
@@ -74,8 +74,8 @@
     },
     {
         "name": "Connector",
-        "evse_id": 1,
-        "connector_id": 2,
+        "evse_id": 2,
+        "connector_id": 1,
         "variables": {
             "ConnectorAvailable": {
                 "variable_name": "Available",


lib/ocpp/v201/device_model.cpp
```diff
@@ -34,6 +34,42 @@ bool DeviceModel::component_criteria_match(const Component& component,
...
+            return true;
+        }
+    }
+    return false;
+}
+bool DeviceModel::component_variables_match(const std::vector<ComponentVariable>& component_variables,
+                                            const ocpp::v201::Component& component_,
+                                            const struct ocpp::v201::Variable& variable_) {
+
+    return std::find_if(
```
marcemmers, line 61: After looking at the other functions you could still simplify this with a similar for loop and split up the conditions in multiple if statements.

If a statement is true you can do an early return true. If not then you continue looping. Returning false if you exit the loop after finding no matches.

lib/ocpp/v201/device_model.cpp
```diff
@@ -257,6 +293,43 @@ DeviceModel::get_report_data(const std::optional<ReportBaseEnum>& report_base,
...
+std::vector<ReportData>
+DeviceModel::get_custom_report_data(const std::optional<std::vector<ComponentVariable>>& component_variables,
+                                    const std::optional<std::vector<ComponentCriterionEnum>>& component_criteria) {
+    std::vector<ReportData> report_data_vec;
+
+    for (auto const& [component, variable_map] : this->device_model) {
+        if (!component_criteria.has_value() or component_criteria_match_custom(component, component_criteria.value())) {
+
+            for (auto const& [variable, variable_meta_data] : variable_map) {
+                auto variable_ = variable;
```
marcemmers, line 305: I don't think we need these anymore since we no longer capture them again later.

lib/ocpp/v201/device_model.cpp
```diff
@@ -257,6 +293,43 @@ DeviceModel::get_report_data(const std::optional<ReportBaseEnum>& report_base,
...
+                auto component_ = component;
+
+                if (!component_variables.has_value() or
+                    (component_variables.has_value() and
```
marcemmers, line 309: Don't need the second has_value since you already checked that and entered the loop if has_value is false.

marcemmers (changes requested):
Looking at the code I think `get_report_data` can be way simplified now since it only gets called from get base report which only ever provides a `report_base` argument. The `component_variables` and `component_criteria` arguments are always nullopt there.

This probably also means that you can replace `component_criteria_match` with your function since it is not used anymore and some more simplifications following from that.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -2047,19 +2047,58 @@ void ChargePoint::handle_get_base_report_req(Call<GetBaseReportRequest> call) {
...
+    if (message.message_size > max_bytes_per_message) {
+        // send a CALLERROR
+        const auto call_error = CallError(call.uniqueId, "FormatViolation", "", json({}));
+        this->send(call_error);
+        return;
+    }
+
+    // if a criteria is not supported then send a not supported response.
+    auto sup_criteria =
+        this->device_model->get_optional_value<std::string>(ControllerComponentVariables::SupportedCriteria);
```
marcemmers, line 2079: What happens if this optional value is not implemented? What do you want to happen?

marcemmers (commented):
Just one question left, otherwise looks good to me

marcemmers (approved):

PR #297, opened 2023-12-07
CI: Add building, testing and linting

couryrr-afs:
Update the GitHub Actions to run the build and test as well as the linting.

@shankari please review.


shankari:
@couryrr-afs interesting! Don't you need to change the README as well? 
https://github.com/EVerest/libocpp#unit-testing

When I read

`cmake .. -DBUILD_TESTING=ON`, I thought that was the literal command to run (e.g. `..` is the directory to run `cmake` in). And it looks like the flag is actually `-DBUILD_TESTING_LIBOCPP=ON \` and not `-DBUILD_TESTING=ON`.

Maybe we should remove that section of the README and suggest that people instead refer to the CI for instructions on how to run tests. That will ensure that we don't have two sources of truth that we have to keep consistent.

couryrr-afs:
@shankari I thought the same thing about the `BUILD_TESTING=ON` but was not getting the results I expected. It just came up in a conversation on the mailing list, [here](https://lists.lfenergy.org/g/everest/message/1084), with one of the other devs that: 

>  BUILD_TESTING is the cmake flag, which configures whether to include or exclude testing in everest-core.

Dominik-K (changes requested):
Hi @couryrr-afs,

thanks for your first contribution. It's good that you pushing that there will be build and test at the CI.

Looks like you've copied most parts of the workflow from the `everest-core` repo (see [here](https://github.com/EVerest/everest-core/blob/a82764df2be2e4b9d86d48819446a94a5848a117/.github/workflows/build_and_test.yaml)). That would result in maintaining multiple, >90% similar workflows. (Besides the point that I think it's nice practice to give credit to the source. A simple link would be enough, though, IMO.)

So can you help on improving the workflow in `everest-core` to become a [reuseable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) instead, please? That would be nice. Thanks.

shankari:
> BUILD_TESTING is the cmake flag, which configures whether to include or exclude testing in everest-core.

But you are using `BUILD_TESTING_LIBOCPP=ON`. 

```
cmake \
    -B build \
    -S "$EXT_MOUNT/source" \
    -DBUILD_TESTING_LIBOCPP=ON \
    -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_INSTALL_PREFIX="$WORKSPACE_PATH/dist"
```

Is that actually required, or is it `BUILD_TESTING`?

couryrr-afs:
@shankari 

If I am reading this correctly `BUILD_TESTING` would allow for the tests to run in `everest-core` as well. To generate the tests all I appear to need is the `BUILD_TESTING_LIBOCPP`.

Any guidance from the community would be appreciated if I am interpreting this incorrectly.

.gitignore
```diff
@@ -1,6 +1,9 @@
 *build*
+!.ci/**
+!.github/**
+!.gitignore
```
Dominik-K, line 4: Why do you need this? I mean you don't want to create a whitelist here, or?

.github/workflows/lint.yaml
Dominik-K, line 1: 1. `clang-format` can run independently of the build and test workflow, as it doesn't need to compile the C++ files.
2. It would stop the whole job if there's a (simple) formatting issue and hinders the test results to be created.

So it's better to run the `clang-format` as extra job in parallel with the "build and test" job.

Yeah, that's also not there in `everest-core`. But as you're touching it, we can already fix this in one go, too. So the `clang-format`-job should best be also called in the resuable workflow, too.

Dominik-K (changes requested):

Dominik-K:
@shankari 
> Maybe we should remove that section of the README and suggest that people instead refer to the CI for instructions on how to run tests. That will ensure that we don't have two sources of truth that we have to keep consistent.

Nice idea :+1:  

Actually, if we transform the `README.md` to `README.rst`, i.e. from Markdown to RestructuredText, we could add the shell-scripts as part of the readme again and still have only one single source of truth :smile:  Which is even automatically checked to be up to date :+1:

.gitignore
couryrr-afs: The inclusion (!) was needed because the build exclusion was blocking the `build-kit` dir and the `build_and_test.sh` file

.gitignore
```diff
@@ -1,6 +1,9 @@
...
 *vscode
-.cache/
+cache/
```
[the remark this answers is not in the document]
couryrr-afs: I do not recall the justification for this. I will revert the change.

.github/workflows/lint.yaml
couryrr-afs: I will pull the linting back into its own workflow.

shankari:
> Actually, if we transform the README.md to README.rst, i.e. from Markdown to reStructuredText, we could add the shell-scripts as part of the readme again and still have only one single source of truth :smile: Which is even automatically checked to be up to date :+1:

@Dominik-K this sounds like a great idea but I am not 100% sure how it will happen. Maybe you can link to an example of how the shell scripts from the README are currently automatically checked or elaborate further.

One back of the envelope option is to have the README.rst include a file with the script. But that would also have to be RST. So then is the expectation that the Github action would parse through the included file and execute the scripts that it finds there?

.gitignore
Dominik-K: Oh, ok. Then better fix this, instead. I.e. `build*/` would work.

tests/database_tests.cpp
```diff
@@ -52,8 +52,9 @@ ChargingProfile get_sample_charging_profile() {
 class DatabaseTest : public ::testing::Test {
 protected:
     void SetUp() override {
-        this->db_handler = std::make_unique<DatabaseHandler>(CP_ID, std::filesystem::path("/tmp"),
-                                                             std::filesystem::path("../../config/v16/init.sql"));
+        this->db_handler =
+            std::make_unique<DatabaseHandler>(CP_ID, std::filesystem::path("/tmp"),
+                                              std::filesystem::path("../../dist/share/everest/modules/OCPP/init.sql"));
```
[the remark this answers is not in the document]
couryrr-afs: @Dominik-K I am trying to understand the use of the set reference in the file provided. Should this give the test access to the `SQL_INIT_FILE` environment variable? Do you have an example of this use?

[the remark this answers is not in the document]
couryrr-afs: I am having a hard to identifying the problem that we are trying to solve. For the test to run in CI the database needs to be initialized with the DDL that are in the file located `"../../dist/share/everest/modules/OCPP/init.sql"`. 

As for the other fixed referenced I am unable to run the tests with that path. I get a `file not found` error. That is what led me to making this change.

Is there another `init.sql` that you would like to be used? Should the build process be moving this file to another location?

.ci/build-kit/install_and_test.sh
```diff
@@ -0,0 +1,17 @@
...
+    -DBUILD_TESTING_LIBOCPP=ON \
+    -DCMAKE_BUILD_TYPE=Debug \
+    -DCMAKE_INSTALL_PREFIX="$WORKSPACE_PATH/dist"
+
+make -j$(nproc) -C build install
+
+cd ./build/tests/ 
+
+./database_tests
+./utils_tests
```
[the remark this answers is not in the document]
couryrr-afs: This was updated to use ctest via ninja.

couryrr-afs:
@shankari I cannot add reviewers but there was an update to this PR for the CI work. Please re-review.

Dominik-K:
@couryrr-afs FYI: I just started trying to reuse the workflow from `everest-core`. See 
- https://github.com/EVerest/libocpp/pull/356

CMakeLists.txt
```diff
@@ -103,7 +103,9 @@ if(CMAKE_RUN_CLANG_TIDY)
     -export-fixes=clang-tidy-fixes.yaml)
 endif()
 
-if(BUILD_TESTING_LIBOCPP)
+if(BUILD_TESTING)
+    enable_testing()
+    include(CTest)
```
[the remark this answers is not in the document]
couryrr-afs: I read that was the case in the documentation but was not having any success without having it present. Let me try and test again without it.

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,68 @@
...
+          docker run \
+          --volume "$(pwd):/ext" \
+          --name test-container \
+          build-kit run-script install_and_test
+      - name: Archive test results
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: ctest-report
+          path: /workspace/build/tests/Testing/Temporary/LastTest.log
```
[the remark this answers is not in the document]
couryrr-afs: I may need to move this to a TODO. I need to figure out how to get the test log out of the docker container.

couryrr-afs:
> The [`database_test` doesn't succeed](https://github.com/US-JOET/libocpp/actions/runs/7251426504/job/19753759899#step:5:454). Unclear why because outputs are missing (see comment).
> 
> Please make sure that all tests run accordingly before asking for review. Incl. the `lint`/`clang-format`.

Very sorry for this. It was working before my commit but I had to do some merges. This seems odd now the test that is failing.

>[2023-12-19 15:11:13.554057] [0x0000ffffb1018cc0] [info]    Established connection to Database.
[2023-12-19 15:11:13.554063] [0x0000ffffb1018cc0] [info]    Running SQL initialization script.
/workspaces/libocpp/tests/database_tests.cpp:327: Failure
Expected equality of these values:
  db_profile.validFrom.value()
    Which is: 2023-12-19T15:11:13.571Z
  profile.validFrom.value()
    Which is: 2023-12-19T15:11:13.571Z
[2023-12-19 15:11:13.573121] [0x0000ffffb1018cc0] [info]    Successfully closed database file
[  FAILED  ] DatabaseTest.test_insert_and_get_profiles (19 ms)
[ RUN      ] DatabaseTest.test_update_profile_same_profile_id

Dominik-K:
@couryrr-afs I've figured out the issue: 
- https://github.com/EVerest/libocpp/issues/384
So it's actually already a good result that the unit-tests are run again. I'm wondering why this wasn't noticed before.

However, I suggest to skip this test for now. As it's now tracked to solve it. I wanted to contribute this change, but it seems you disallowed changes by maintainers. Or?

So I want to commit this change in `database_tests.cpp`:
```cpp
TEST_F(DatabaseTest, test_insert_and_get_profiles) {
    // TODO enable again on fixing https://github.com/EVerest/libocpp/issues/384
    GTEST_SKIP() << "validFrom/validTo checks are failing. See https://github.com/EVerest/libocpp/issues/384";
```

I'm surprised that in the test-result summary it's nowhere mentioned that one test is skipped. Or is it just with my setup?

.ci/build-kit/install_and_test.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -e
+
+cmake \
+    -B build \
+    -S "$EXT_MOUNT/source" \
```
Dominik-K, line 7: This line is specific for within the docker-container. I think it can be removed and the source-dir set to the current working-dir in the workflow itself.

.ci/build-kit/install_and_test.sh
```diff
@@ -0,0 +1,17 @@
...
+    -G Ninja \
+    -DBUILD_TESTING=ON \
+    -DCMAKE_BUILD_TYPE=Debug \
+    -DCMAKE_INSTALL_PREFIX="$WORKSPACE_PATH/dist"
+
+ninja -j$(nproc) -C build install
+
+trap "cp build/Testing/Temporary/LastTest.log /ext/ctest-report" EXIT
+
+ninja -j$(nproc) -C build test
```
Dominik-K, line 17: The `/ext/` is specific within the docker-container. When you split-up this script in two files (`build_install.sh` and `test.sh`), you could run the `trap` inside workflow. That's not needed when running locally. (However, I hope the files here will soon be replaced with the reusable workflow anyway.)

Dominik-K (approved):
[LGTM with comments](https://google.github.io/eng-practices/review/reviewer/speed.html#lgtm-with-comments) (in the meaning: I assume you'll address my comments.)

.ci/build-kit/install_and_test.sh
couryrr-afs: Is there a desire to not run the build and test in a docker container? Currently, that was the process. If there needs to be an update to that process could it be a separate issue? I would like to get this PR merged so that tests are being ran in an automated fashion.

Dominik-K: Yes, the scripts run in a Docker-container in GitHub (mainly that the CI could be easily debugged locally, as there no SSH-access, etc. to the CI-runners).

But this script should now also serve as the source for the local run (see https://github.com/EVerest/libocpp/pull/297#issuecomment-1849020534).

However, with the reusable workflow all this files (as all others) will be gone in this repo anyway.

Dominik-K:
@shankari Please review again. I'll merge it when you approved it.

Note: It will be replaced with the reusable workflow in
- #356 
anyway.

shankari (approved):
LGTM as a short-term improvement, soon to be superceded by the reusable workflow.
I'm glad this exposed the issue around timestamps, which is apparently not trivial to fix.
Testing FTW!

@Dominik-K you can go ahead and merge this.

couryrr-afs:
@Dominik-K  and @shankari  I am pushing one more change this morning. Looks like it was still running on branch pushes.

Dominik-K (approved):

PR #298, opened 2023-12-08
OCPP 2.0.1: Make message interval, retries, and timeout configurable at run-time

valentin-dimov:


valentin-dimov:
Something worth getting a second opinion on: When retrying a transaction-related message after a timeout or CallError, we will:
- Reuse the same message ID for the retry if it's retrying due to a timeout;
- Generate a new message ID for the retry if it's retrying due to a CallError.

hikinggrass (approved):

hikinggrass:
> Something worth getting a second opinion on: When retrying a transaction-related message after a timeout or CallError, we will:
> 
>     * Reuse the same message ID for the retry if retrying due to a timeout;
> 
>     * Generate a new message ID for the retry if retrying due to a CallError.

That's something we should probably fix, reusing message id's doesn't seem the correct thing to do

valentin-dimov:
Following some more discussion, it was decided that we will indeed re-generate new message IDs for retries, even if the retry was due to a time-out.

hikinggrass (approved):

PR #299, opened 2023-12-11
Fix potential crashes and fix TxProfile during transaction

hikinggrass:


corneliusclaussen (approved):
Looks good!

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -1865,7 +1865,7 @@ void ChargePointImpl::handleSetChargingProfileRequest(ocpp::Call<SetChargingProf
                       << call.msg.csChargingProfiles.chargingProfilePurpose;
         response.status = ChargingProfileStatus::Rejected;
     } else if (this->smart_charging_handler->validate_profile(
-                   profile, connector_id, false, this->configuration->getChargeProfileMaxStackLevel(),
+                   profile, connector_id, true, this->configuration->getChargeProfileMaxStackLevel(),
```
Pietfried, line 1868: This allows TxProfile(s) without a transaction_id, which I consider invalid for a SetChargingProfile.req .
The purpose of the ignore_no_transaction flag was to allow ChargingProfiles as part of RemoteStartTransaction.req messages, because in that case no transaction is active yet: https://github.com/EVerest/libocpp/blob/main/lib/ocpp/v16/charge_point_impl.cpp#L1606

Can you ellaborate why this had to change?

hikinggrass: I think this is a part of the OCPP 1.6 spec that unfortunately leaves room for interpretation. Maybe we should keep this behavior behind an additional configuration key though.

The relevant section in the 1.6 spec (page 54 in edition 2 FINAL, 2017-09-28):

> To prevent mismatch between transactions and a TxProfile, The Central System SHALL include the transactionId in a SetChargingProfile.req **if the profile applies to a specific transaction**.

If the CSMS just wants to limit a connector during a transaction (just for this transaction) but doesn't care about the specific transaction, at least in my interpretation, it is valid to not include the transaction id

PR #301, opened 2023-12-11
cmake: add `LIBOCPP_` prefix to library-own options

Dominik-K:
Adding the `LIBOCPP_` as best-practice like done recently in
- https://github.com/EVerest/libocpp/pull/177

See also e.g. https://discourse.cmake.org/t/best-practices-for-option-naming/2039/3

This inconsistency was figured out in
-  https://github.com/EVerest/libocpp/pull/297


andistorm:
See https://github.com/EVerest/EVerest/issues/129

andistorm:
@hikinggrass

PR #306, opened 2023-12-14
OCPP 1.6 Fix: add flag for pending firmware update

klemmpnx:
In OCPP 1.6 , a flag that signals a pending firmware update is introduced. This fixes a bug in case the last received firmware status while a pending update is not "Downloaded" but something else (such as "Installing").

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -2380,7 +2380,7 @@ void ChargePointImpl::signed_firmware_update_status_notification(FirmwareStatusE
     req.requestId = requestId;
 
     // The "SignatureVerified" status signals a firmware update is pending which will cause connectors to be set
-    // unavailable (now or after pending transactions are stopped); in case of a status that signals failed firmware
+    // inoperative (now or after pending transactions are stopped); in case of a status that signals a failed firmware
     // update this is revoked
     if (status == FirmwareStatusEnumType::SignatureVerified) {
         this->firmware_update_is_pending = true;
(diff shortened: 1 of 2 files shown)


lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -2371,11 +2373,21 @@ void ChargePointImpl::log_status_notification(UploadLogStatusEnumType status, in
...
     req.requestId = requestId;
 
+    if (status == FirmwareStatusEnumType::SignatureVerified) {
+        this->firmware_update_is_pending = true;
+    } else if (status == FirmwareStatusEnumType::InstallationFailed ||
+               status == FirmwareStatusEnumType::DownloadFailed ||
+               status == FirmwareStatusEnumType::InstallVerificationFailed ||
+               status == FirmwareStatusEnumType::InvalidSignature) {
+        this->firmware_update_is_pending = false;
+    }
```
Dominik-K, line 2389: Use `switch` with enums. With that the compiler gives warnings on missing unhandled values. Try to avoid using the `default` case as it deactivates this functionality with the usual `-Wswitch` (which is in `-Wall`, but not `-Wswitch-enum`, at least in [`gcc`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wswitch-enum)).

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -2371,11 +2373,21 @@ void ChargePointImpl::log_status_notification(UploadLogStatusEnumType status, in
 }
 
 void ChargePointImpl::signed_firmware_update_status_notification(FirmwareStatusEnumType status, int requestId) {
-    EVLOG_debug << "Sending FirmwareUpdateStatusNotification";
+    EVLOG_debug << "Sending FirmwareUpdateStatusNotification with status"
+                << conversions::firmware_status_enum_type_to_string(status);
```
Dominik-K, line 2377: Isn't there a generic debug-log in `this->send()` already? :thinking:

lib/ocpp/v16/charge_point_impl.cpp
klemmpnx: Those two cases are not exhaustive (on purpose) though (so I would need a default case that just breaks to indeed avoid the warning). I added a comment in the code as explanation (of course, I could still also switch to a switch)

lib/ocpp/v16/charge_point_impl.cpp
klemmpnx: Hmm which one do you mean? Could be that the websocket has some generic debug log. But for me the crucial aspect was that the logs actually reveal which (and in which order) firmware status updates (meaning the value) are sent and I don't think any existing log prints those (so I left this in case the firmware update make trouble again in the future)

hikinggrass (approved):

PR #379, opened 2023-12-22
Update README.md

ahzf:
Just two links and a new line ;)


Pietfried (changes requested):
Hi, please also add the link the SteVe so that we have all open source solutions covered :)

Pietfried (approved):

PR #382, opened 2023-12-22
Feature/fk block enable request

klemmpnx:
Enables users to set a custom variable `BlockRemoteSetAvailableRequests` to block incoming ChangeAvailability requests that set the CS / a connector operative (cf. comments in code).

(Choice as "custom variable" is to allow users to set this at runtime via the OCPP interface)


corneliusclaussen:
Will be implemented differently, closing this PR.

Issue #384, opened 2023-12-22
OCPP date-time types should have only seconds precision


Dominik-K:
The `DateTime` class in `ocpp/common/` uses internally the `std::chrono::time_point<date::utc_clock>` which has a sub-second precision by default (at least on most platforms). When this `time_point` is initialized with a value with lower precision, e.g. from a CSMS message or the database, then the sub-second fractions have arbitrary values. But the `==`-equality-check is simply done on the `time_point`s:

https://github.com/EVerest/libocpp/blob/6f73dec368430f8fd67401a6f44cadad9aaf0e08/lib/ocpp/common/types.cpp#L95-L97

That leads to a [failure in the database-tests](https://github.com/EVerest/libocpp/pull/297#issuecomment-1862967797) (thanks @couryrr-afs for catching it). It can be easy verified by using the `.to_time_point()` in the GTest assertions.

## Solution
The `DateTime`'s internal `timepoint` should only have seconds precision. As in both OCPP-versions, the sub-second precision may be omitted:

OCPP 1.6 spec (edition 2 FINAL, 2017-09-2), 3.15. Time notations:
> Implementations MUST use ISO 8601 date time notation. Message receivers must be able to handle fractional
seconds and time zone offsets (another implementation might use them). Message senders MAY save data
usage by **omitting insignificant fractions of seconds**. [emphasis mine]

OCPP 2.0.1 spec (Edition 2 FINAL, 2022-12-15), 3.1. Time Format Requirements:
> All time values exchanged between CSMS and Charging Station SHALL be formatted as defined in RFC-3339 [RFC3339].
Additionally fractional seconds have been given an extra limit. The number of decimal places SHALL NOT exceed the maximum of 3. However, it is RECOMMENDED to **omit fractional seconds entirely**, because it is of limited use and omitting it reduces data usages. [emphasis mine]

But the `DateTime`-class is also used for other internal types (which are not OCPP-messages). At least changing the precision fails a `message_queue` unit-test which hints that perhaps the internal-types need sub-second precision. See https://github.com/EVerest/libocpp/pull/383

Given that, I suggest to add a new type `DateTimeSeconds(Precision)` and deliberately use this in the OCPP-message related types. Then the remaining usages could be checked on precision-level later. (This approach resembles the [Open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle), I guess.)

Dominik-K:
@RobertDeLeeuw RFC

[...]

PR #389, opened 2024-01-03
Fix unit tests: database_tests and evse_security_mock

Pietfried:
minor changes to fix unit tests


Dominik-K (approved):
Technically, the fixes are correct. However, the underlying error isn't fixed with this. See 
- https://github.com/EVerest/libocpp/issues/384

In 
- https://github.com/EVerest/libocpp/pull/297
I've decided to deliberately skip this test to better see° that something needs to be fixed. So with the approach here, I suggest to add a TODO at the fixes, at least.

° Don't know why the skipping isn't displayed in the test-results, yet.

PR #393, opened 2024-01-05
Configurable CentralSystemURI for OCPP1.6

Pietfried:
made CentralSystemURI of OCPP1.6 writable by CSMS. Change to this configuration key is applied only after reboot

### DIFF
--- a/config/v16/profile_schemas/Internal.json
+++ b/config/v16/profile_schemas/Internal.json
@@ -18,7 +18,7 @@
         },
         "CentralSystemURI": {
             "type": "string",
-            "readOnly": true,
+            "readOnly": false,
             "minLength": 1
         },
         "ChargeBoxSerialNumber": {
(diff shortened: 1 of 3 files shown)


Pietfried:
- [ ] Because it can cause loss of OCPP connectivity when this value is changed: Hide feature to write `CentralSystemURI` behind a separate configuration key

lategoodbye:
AFAIR implementing this feature in our stack wasn't trivial. Usually a change of the backend URI should drop all persistent data (authorization cache, local auth list, pending transactions, not sure about smart charging profiles) assigned to the old URI.

Pietfried:
> AFAIR implementing this feature in our stack wasn't trivial. Usually a change of the backend URI should drop all persistent data (authorization cache, local auth list, pending transactions, not sure about smart charging profiles) assigned to the old URI.

Hi @lategoodbye , thanks for pointing this out. I added this to the further TODOs for this PR

Pietfried:
We decided to not merge it like that and probably use another approach to implement this feature

PR #398, opened 2024-01-10
Add support for certificate links

mennodegraaf:
* Add support for certificate links by providing `update_certificate_links()`

### DIFF
--- a/config/v201/component_schemas/standardized/InternalCtrlr.json
+++ b/config/v201/component_schemas/standardized/InternalCtrlr.json
@@ -471,6 +471,21 @@
           "default": "43200",
           "type": "integer"
       },
+      "UpdateCertificateSymlinks": {
+        "variable_name": "UpdateCertificateSymlinks",
+        "characteristics": {
+            "supportsMonitoring": false,
+            "dataType": "boolean"
+        },
+        "attributes": [
+            {
+                "type": "Actual",
+                "mutability": "ReadOnly"
+            }
+        ],
+        "default": "0",
+        "type": "boolean"
+      },
       "MessageQueueSizeThreshold": {
           "variable_name": "MessageQueueSizeThreshold",
           "characteristics": {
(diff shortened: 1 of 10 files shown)


AssemblyJohn:
@mennodegraaf due to compatibility issues, as a feature that might not be always required, it would be best if we could have this configurable, and work for both ocpp1.6 and ocpp2.0.1. The security lib (with the exception of the GetKeyPair which will be refactored in everest-core) does not have the side-effects which libocpp can have.

mennodegraaf:
> @mennodegraaf due to compatibility issues, as a feature that might not be always required, it would be best if we could have this configurable, and work for both ocpp1.6 and ocpp2.0.1. The security lib (with the exception of the GetKeyPair which will be refactored in everest-core) does not have the side-effects which libocpp can have.

This symlink functionality is configurable in the sense that if the symlinks are not configured nothing will be done and the functionality is as it was before.

AssemblyJohn:
> > @mennodegraaf due to compatibility issues, as a feature that might not be always required, it would be best if we could have this configurable, and work for both ocpp1.6 and ocpp2.0.1. The security lib (with the exception of the GetKeyPair which will be refactored in everest-core) does not have the side-effects which libocpp can have.
> 
> This symlink functionality is configurable in the sense that if the symlinks are not configured nothing will be done and the functionality is as it was before.

I think it would be safer to have the function call to be configurable (if the user wants it or not), especially since it is a niche application, and there can always be side-effect. Except that, I can't find any other issues.

mennodegraaf:
@AssemblyJohn I will it up to you if you want to prevent `update_certificate_links()` from being called with some config parameter. Note that this method will also make sure that symlinks are removed, e.g. when no certificate is valid anymore. So if symlinks are first created and then it is configured not to call the method anymore and then a certificate expires they will not be deleted.

hikinggrass (approved):

PR #292, opened 2023-12-06
Fix: OCPP 2.0.1 CS, EVSE, and Connector availability tracking

valentin-dimov:
Companion PR from everest-core: https://github.com/EVerest/everest-core/pull/459

PR #400, opened 2024-01-12
Sync Operative state to connector states. 

gregsq:
Maintain Operative status when made Unavailable.
Pre init Connector status after loading Operative state of the ChargePoint and EVSE's


Pietfried:
Thanks for the contribution. This looks like its conflicting and adressing the same issue like https://github.com/EVerest/libocpp/pull/292 , which will propably be merged at the beginning of next week.

hikinggrass:
Thanks for you contribution, this has been implemented in #292 and https://github.com/EVerest/everest-core/pull/501

PR #406, opened 2024-01-18
Addin WaitForStopTransactionsOnResetTimeout as required parameter as …

bWF0dGhpYXMK:
…it is queried in case of Rest.req handling.
The handling of the Reset.req requires that parameter.


Pietfried:
`WaitForStopTransactionsOnResetTimeout` has a default value in https://github.com/EVerest/libocpp/blob/main/config/v16/profile_schemas/Internal.json#L231 , which will be patched into the config file here: https://github.com/EVerest/libocpp/blob/main/lib/ocpp/v16/charge_point_configuration.cpp#L52 , so this change is not required

PR #407, opened 2024-01-18
Security Log file option

Pietfried:
Added option for writing SecurityEvents to a security log file for ocpp1.6 and ocpp2.0.1.
This can be enabled by adding "security" to LogMessagesFormat configuration parameter.

### DIFF
--- a/lib/ocpp/common/ocpp_logging.cpp
+++ b/lib/ocpp/common/ocpp_logging.cpp
@@ -14,7 +14,7 @@ namespace ocpp {
 
 MessageLogging::MessageLogging(
     bool log_messages, const std::string& message_log_path, const std::string& output_file_name, bool log_to_console,
-    bool detailed_log_to_console, bool log_to_file, bool log_to_html, bool session_logging,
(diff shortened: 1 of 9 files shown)


hikinggrass (approved):

lib/ocpp/common/ocpp_logging.cpp
```diff
@@ -69,6 +70,10 @@ MessageLogging::MessageLogging(
                                    "</style>";
             this->html_log_file << "</head><body><table class=\"log\">\n";
         }
+        if (this->log_security) {
+            EVLOG_info << "Logging SecurityEvents to file";
+            this->security_log_file.open(message_log_path + "/" + output_file_name + ".security.log");
```
corneliusclaussen, line 75: As far as I can see this overwrites the old already existing security log as it does not append?

hikinggrass: In 1.6 (and 2.0.1) the output_file_name is set to the current datetime: https://github.com/EVerest/libocpp/blob/35195e638790b7a6fc123065b418b10c85c403ab/lib/ocpp/v16/charge_point_impl.cpp#L56

So it's very likely different for each run

PR #410, opened 2024-01-22
network connection profiles

SNSubramanya:


include/ocpp/v201/charge_point.hpp
```diff
@@ -166,6 +161,21 @@ struct Callbacks {
...
+    std::function < void(std::optional<NetworkConnectionProfile> network_profile)> websocket_connected_callback;
+
+    /// \brief register a \p callback that is called when the websocket connection is disconnected
+    std::function<void()>websocket_disconnected_callback;
+
+    // callback is called when receiving a SetNetworkProfile.req from the CSMS
+    std::optional<std::function<SetNetworkProfileStatusEnum(
+        const int32_t configuration_slot, const NetworkConnectionProfile& network_connection_profile)>>
+        validate_network_profile_callback;
+    std::optional<std::function<bool(const NetworkConnectionProfile& network_connection_profile,
```
maaikez, line 176: Why don't you return the promise in this callback? 

And I think you should return a future, not the promise.

include/ocpp/v201/charge_point.hpp
```diff
@@ -166,6 +161,21 @@ struct Callbacks {
     /// \brief Callback function that can be used to handle arbitrary data transfers for all vendorId and
     /// messageId
     std::optional<std::function<DataTransferResponse(const DataTransferRequest& request)>> data_transfer_callback;
+
+    /* Callbacks for networking */
+    /// \brief register a \p callback that is called when the websocket is connected successfully
+    std::function < void(std::optional<NetworkConnectionProfile> network_profile)> websocket_connected_callback;
```
maaikez, line 167: This can be optional

include/ocpp/v201/charge_point.hpp
```diff
@@ -166,6 +161,21 @@ struct Callbacks {
...
+
+    /// \brief register a \p callback that is called when the websocket connection is disconnected
+    std::function<void()>websocket_disconnected_callback;
```
maaikez, line 170: This can be optional as well. It is not mandatory to know for the implementing class if the websocket is connected or disconnected.

include/ocpp/v201/charge_point.hpp
```diff
@@ -166,6 +161,22 @@ struct Callbacks {
     /// \brief Callback function that can be used to handle arbitrary data transfers for all vendorId and
     /// messageId
     std::optional<std::function<DataTransferResponse(const DataTransferRequest& request)>> data_transfer_callback;
+
+    /* Callbacks for networking */
+    /// \brief register a \p callback that is called when the websocket is connected successfully
+    std::optional<std::function<void(const std::optional<NetworkConnectionProfile>& network_connection_profile)>>
```
maaikez, line 167: In v16 there is this callback:
    `void register_connection_state_changed_callback(const std::function<void(bool is_connected)>& callback);`
It would be nice to have it synced with 1.6, however, I like the NetworkConnectionProfile here because it is nice to know to which profile we are connected now.

include/ocpp/v201/ocpp_types.hpp
```diff
@@ -861,6 +861,10 @@ void from_json(const json& j, NetworkConnectionProfile& k);
 /// \returns an output stream with the NetworkConnectionProfile written to
 std::ostream& operator<<(std::ostream& os, const NetworkConnectionProfile& k);
 
+struct ConfigNetworkResult {
```
maaikez, line 864: Don't forget to document this, that the interface address can be an ip address string or interface name ("192.168.42.1" or "eth6" for example).

lib/ocpp/v201/charge_point.cpp
```diff
@@ -660,9 +660,28 @@ void ChargePoint::init_websocket() {
...
+                << "NetworkConnectionProfile could not be retrieved or configuration of network with the given "
+                   "profile failed";
+            this->websocket_timer.timeout(
+                [this]() {
+                    this->next_network_configuration_priority();
+                    this->start_websocket();
+                },
+                WEBSOCKET_INIT_DELAY);
+            return;
+        } else if (config_status.wait_for(5s) == std::future_status::timeout) {
```
maaikez, line 679: huh? same code as the previous if?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -660,9 +660,28 @@ void ChargePoint::init_websocket() {
...
+            EVLOG_warning
+                << "NetworkConnectionProfile could not be retrieved or configuration of network with the given "
+                   "profile failed";
+            this->websocket_timer.timeout(
+                [this]() {
+                    this->next_network_configuration_priority();
+                    this->start_websocket();
+                },
+                WEBSOCKET_INIT_DELAY);
+            return;
```
maaikez, line 678: if there is a return, you don't need an 'else' clause?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -673,7 +692,6 @@ void ChargePoint::init_websocket() {
             WEBSOCKET_INIT_DELAY);
         return;
     }
-
     const auto& security_profile_cv = ControllerComponentVariables::SecurityProfile;
```
maaikez, line 695: You are not checking if it was successful before running this code.

And I think it's good to add a TODO here (maybe including a ticket number) that we have to reconnect the websocket with the given interface address / name.

include/ocpp/v201/charge_point.hpp
```diff
@@ -722,6 +739,12 @@ class ChargePoint : ocpp::ChargingStationBase {
                                                 const AttributeEnum& attribute_enum) {
         return this->device_model->request_value<T>(component_id, variable_id, attribute_enum);
     }
+
+    /// @brief Switch to a specifc network connection profile given the configuration slot.
```
maaikez, line 743: I would say try to switch here, or even do nothing. It depends, if this is called multiple times and every time the switch is not possible, then ignore it for example. Or when the network connection profile priority is lower, this can be ignored as well I think.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -27,7 +26,7 @@ const auto DEFAULT_MAX_MESSAGE_SIZE = 65000;
 bool Callbacks::all_callbacks_valid() const {
     return this->is_reset_allowed_callback != nullptr and this->reset_callback != nullptr and
            this->stop_transaction_callback != nullptr and this->pause_charging_callback != nullptr and
-           this->connector_effective_operative_status_changed_callback != nullptr and
+           //    this->connector_effective_operative_status_changed_callback != nullptr and
```
maaikez, line 29: why is this out commented?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -38,7 +37,9 @@ bool Callbacks::all_callbacks_valid() const {
             this->configure_network_connection_profile_callback.value() != nullptr) and
            (!this->time_sync_callback.has_value() or this->time_sync_callback.value() != nullptr) and
            (!this->boot_notification_callback.has_value() or this->boot_notification_callback.value() != nullptr) and
-           (!this->ocpp_messages_callback.has_value() or this->ocpp_messages_callback.value() != nullptr);
+           (!this->ocpp_messages_callback.has_value() or this->ocpp_messages_callback.value() != nullptr) and
+           (!this->websocket_connected_callback.has_value() or this->websocket_connected_callback != nullptr) and
```
maaikez, line 41: those two are not mandatory is it?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -2979,6 +3020,30 @@ void ChargePoint::set_connector_operative_status(int32_t evse_id, int32_t connec
     this->evses.at(evse_id)->set_connector_operative_status(connector_id, new_status, persist);
 }
 
+bool ChargePoint::on_try_switch_network_connection_profile(const std::string configuration_slot) {
```
maaikez, line 3023: is it logical that this is a string instead of an int?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -2979,6 +3020,30 @@ void ChargePoint::set_connector_operative_status(int32_t evse_id, int32_t connec
...
+    EVLOG_info <<  "=============on_try_switch_network_profile============" << configuration_slot;
+
+    // check if the configuration slot is valid
```
maaikez, line 3026: you should also call `configure_network_connection_profile_callback` here and follow the 'normal' flow.

include/ocpp/v201/ocpp_types.hpp
```diff
@@ -861,6 +861,10 @@ void from_json(const json& j, NetworkConnectionProfile& k);
...
+    std::optional<CiString<32>> interface_adress;
```
maaikez, line 865: address is with two d's :).

include/ocpp/v201/ocpp_types.hpp
maaikez, line 864: Maybe, to be sure, add the network profile slot here as well as an extra check.

include/ocpp/common/websocket/websocket.hpp
```diff
@@ -9,6 +9,10 @@
 #include <ocpp/common/websocket/websocket_tls.hpp>
```
AssemblyJohn, line 9: Now that we have abstracted, do we still require these headers? The purpose of the refactor was to remove any dependency on them.

include/ocpp/common/websocket/websocket_base.hpp
```diff
@@ -36,6 +36,7 @@ struct WebsocketConnectionOptions {
     std::optional<std::string> hostName;
     bool verify_csms_common_name;
     bool use_tpm_tls;
+    std::optional<std::string> iface_or_ip;    ///< The interface of the connection or the ip address of the interface
```
AssemblyJohn, line 39: Looks good, however was the equivalent config value added to the db schema?

include/ocpp/common/websocket/websocket_base.hpp
```diff
@@ -47,7 +48,7 @@ class WebsocketBase {
     WebsocketConnectionOptions connection_options;
     std::function<void(const int security_profile)> connected_callback;
     std::function<void()> disconnected_callback;
-    std::function<void(const websocketpp::close::status::value reason)> closed_callback;
+    std::function<void(const WebsocketCloseReason reason)> closed_callback;
     std::function<void(const std::string& message)> message_callback;
     websocketpp::lib::shared_ptr<boost::asio::steady_timer> reconnect_timer;
```
AssemblyJohn, line 53: @hikinggrass maybe we should also abstract usage to the timer and use the evse timer instead?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -2979,6 +3042,40 @@ void ChargePoint::set_connector_operative_status(int32_t evse_id, int32_t connec
...
 }
 
+bool ChargePoint::on_try_switch_network_connection_profile(const std::string configuration_slot) {
+    EVLOG_info << "=============on_try_switch_network_profile============" << configuration_slot;
+
+    // check if the configuration slot is valid
+    try {
+        // call disconnect
+        this->disconnect_websocket(); // normal close
+        while (!this->is_offline()) {
```
AssemblyJohn, line 3052: There must be a better way to wait for this, than a while loop, they're never a good idea.

include/ocpp/common/websocket/websocket.hpp
maaikez: Well I only changed the interface, I think you are right.

include/ocpp/common/websocket/websocket_base.hpp
maaikez: What do you mean?

This will not be a value from the device model but it is returned in the future of the configure_network_connection_profile_callback function

include/ocpp/common/websocket/websocket_base.hpp
maaikez: Yes and there is more to be abstracted I think. I did not go through all the websocketpp things there, only the things I needed to be changed. I did not look into detail how websocketpp and the websocket in general works.

include/ocpp/common/websocket/websocket_base.hpp
AssemblyJohn: Understood, I missed that part.

include/ocpp/common/websocket/websocket.hpp
```diff
@@ -50,7 +54,7 @@ class Websocket {
 
     /// \brief register a \p callback that is called when the websocket connection has been closed and will not attempt
     /// to reconnect
-    void register_closed_callback(const std::function<void(const websocketpp::close::status::value reason)>& callback);
+    void register_closed_callback(const std::function<void(const WebsocketCloseReason reason)>& callback);
```
Pietfried, line 57: We can keep the `closed_callback` and `disconnected_callback` for now because they are implicitly used to manage the reconnect behavior, but to reduce complexity of the interface I would prefer to not have the seperated after this refactor is done

maaikez:
There are a lot of changes now. Tested this with websocket plain (not tls) and only ocpp 2.0.1. Is it possible to take a look again if this is the direction we want to go? 

Maybe we want to move 'ping' to the connectivity manager as well.

AssemblyJohn:
It has been concluded that in the current form for network profiles serves the purpose. However, the config value will be used only as the 'default' for a network profile, if at runtime the connection manager executes an config parameter over-write, that will be the used network interface.

Linked PR: https://github.com/EVerest/libocpp/pull/552

lib/ocpp/v201/connectivity_manager.cpp
```diff
@@ -0,0 +1,748 @@
...
+
+// cmake .. -DLIBOCPP_ENABLE_LIBWEBSOCKETS=ON
+//     https://github.com/EVerest/libocpp
+
+namespace ocpp {
+namespace v201 {
+
+/// \brief Default timeout for the return value (future) of the `configure_network_connection_profile_callback`
+///        function.
+constexpr int32_t default_network_config_timeout_seconds = 60;
```
marcemmers, line 21: These can (should) be chrono types

lib/ocpp/v201/connectivity_manager.cpp
```diff
@@ -0,0 +1,748 @@
...
+        // No config slot given. Get first available config slot.
+        const std::vector<std::string> config_slot_vector = ocpp::get_vector_from_csv(
+            this->device_model.get_value<std::string>(ControllerComponentVariables::NetworkConfigurationPriority));
+        if (config_slot_vector.size() > 0) {
+            configuration_slot = config_slot_vector.at(0);
+            config_slot_int = std::stoi(configuration_slot);
+            EVLOG_debug << "Init websocket without config slot, use first available: " << config_slot_int;
+        } else {
+            // No network connection profile. Retry connecting after some time, maybe it is set manually.
+            EVLOG_info << "No network connection profile set. Try again later.";
```
marcemmers, line 308: Maybe we can try this more often then every 300 seconds but not print this message every time?

PR #421, opened 2024-01-26
V1.6 in Pending state: Handling of the TriggerMessage.req to set connection state to accepted

bWF0dGhpYXMK:
If the answer of backend is bootnotificaiton.conf pending, then a TriggerMessage.req shall set the charge point to booted.


bWF0dGhpYXMK:
In order to accept TriggerMessage.req to send out the relevant requested *.req e.g. Heartbeat requires a bit of refactor as there can happen a interference with calls comming from outside. Stacked variable would be preferred to manage acceptance to allow sending of the package, although, it is in pending state.

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -993,7 +993,14 @@ void ChargePointImpl::message_callback(const std::string& message) {
                 if (enhanced_message.messageType == MessageType::BootNotificationResponse) {
                     this->handleBootNotificationResponse(json_message);
                 } else {
-                    this->handle_message(enhanced_message);
+                    if (enhanced_message.messageType == MessageType::GetConfiguration ||
+                        enhanced_message.messageType == MessageType::ChangeConfiguration ||
+                        enhanced_message.messageType == MessageType::TriggerMessage) {
```
hikinggrass, line 998: I think this is too restrictive, the 1.6 spec in section 4.2 states

> While in pending state, the following Central System initiated messages are not allowed:
RemoteStartTransaction.req and RemoteStopTransaction.req

There are no other restriction (other than not issuing chargepoint initiated requests)

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -993,7 +993,14 @@ void ChargePointImpl::message_callback(const std::string& message) {
...
+                        this->handle_message(enhanced_message);
+                    } else {
+                        auto call_error = CallError(enhanced_message.uniqueId, "InPendingState", "", json({}, true));
```
hikinggrass, line 1001: I think this InPendingState error code does not exist, should probably be something like "GenericError" with a explanatory description

hikinggrass (changes requested):

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -2010,6 +2014,11 @@ void ChargePointImpl::handleTriggerMessageRequest(ocpp::Call<TriggerMessageReque
         valid = false;
     }
 
+    if (this->connection_state = ChargePointConnectionState::Pending) {
+        EVLOG_debug << "Set connection_state from pending to accepted";
+        this->connection_state = ChargePointConnectionState::Booted;
+    }
```
lategoodbye: How do you come to this conclusion? From my understanding only the BootNotification.Conf should change this.

lib/ocpp/v16/charge_point_impl.cpp
hikinggrass: @bWF0dGhpYXMK what's your opinion on this?

bWF0dGhpYXMK: Hi, a review of the latest code, there is a PR https://github.com/EVerest/libocpp/pull/520 managing that case with an answer on Rejected. From the ocpp1.6, for me it is not clear what means not allowed, initially I thought on "no" answer, which is two edged. Because an on the RPC framework point of view you require at least an answer such as Call Error. 
So on that level, i think this PR can be closed. Either reject or CALL Error.

It remains the clarification on ReserverNow.req, Reset.req, ... what behavior here is the most likely in the sense of the describtion. 

From this " The Central System MAY send request messages to retrieve information from the
Charge Point or change its configuration. The Charge Point SHOULD respond to these messages.". It sound like behave as in Remote(Start/Stop).req...

bWF0dGhpYXMK:
Following discussions on [Issue 478](https://github.com/EVerest/libocpp/issues/478) related to the topic, the PR is set as done/closed.

PR #425, opened 2024-01-30
Extended Use of SecurityEventNotification.req in OCPP1.6

Pietfried:
* Sending SecurityEventNotification in OCPP1.6 on startup if StartUpOfTheDevice or ResetOrReboot
* added optional BootReason to charge_point start function
* added optional BootReason to charge_point restart function
* using string definitions for all SecurityEventNotifications in 1.6


hikinggrass (approved):

Pietfried:
Before merge:
- Introduce configuration key for 1.6 to control if SecurityEventNotification.req should be send or not [ ]

PR #429, opened 2024-01-31
Localised per-thread conn data for a clean disconnect

AssemblyJohn:
- updated thread notify for fast message sending
- fixed the issue in which a message send could timeout due to a wait in the processing thread
- clearer message log
- localized per-thread connection data for a clean state


james-ctc (approved):
Changes look good.
Testing locally successful.

PR #431, opened 2024-01-31
Improved handling in case CSMS responds with CALLERROR or not at all to BootNotification.req

Pietfried:
* Added handling for BootNotification.req in case CSMS responds with CALLERROR or no response at all
* BootNotification now has special handling in the message queue so that it will be repeately sent in case of CALLERROR or no message timeout

### DIFF
--- a/include/ocpp/common/message_queue.hpp
+++ b/include/ocpp/common/message_queue.hpp
@@ -38,7 +38,9 @@ struct MessageQueueConfig {
     bool queue_all_messages; // cf. OCPP 2.0.1. "QueueAllMessages" in OCPPCommCtrlr
 
     int message_timeout_seconds = 30;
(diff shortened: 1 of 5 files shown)


include/ocpp/common/message_queue.hpp
```diff
@@ -38,6 +38,9 @@ struct MessageQueueConfig {
     bool queue_all_messages; // cf. OCPP 2.0.1. "QueueAllMessages" in OCPPCommCtrlr
 
     int message_timeout_seconds = 30;
+    int boot_notification_retry_interval_seconds =
+        60; // interval for BootNotification.req in case response by CSMS is CALLERROR or CSMS does not respond at all
```
hikinggrass, line 42: This formatting looks really strange

hikinggrass (approved):

PR #440, opened 2024-02-01
Add 1.6 SmartChargingHandler Validate and Add Profile Tests

folkengine:
Adds unit tests for `lib/ocpp/v16/smart_charging.cpp` `SmartChargingHandler::validate_profile` method as well as tests for the `add_charge_point_max_profile`, `add_tx_default_profile` and `add_tx_profile` methods. 

## Validate Profile

`SmartChargingHandler::validate_profile` covers the following positive (returns true) and negative (returns false) boundary conditions:

Positive Boundary Conditions:
- PB01 Valid Profile
- PB02 Valid Profile No startSchedule & handler allows no startSchedule & profile.chargingProfileKind == Absolute
- PB03 Valid Profile No startSchedule & handler allows no startSchedule & profile.chargingProfileKind == Relative
- PB04 Absolute ChargePointMaxProfile Profile with connector id 0
- PB05 Absolute TxDefaultProfile 
- PB06 Absolute TxProfile ignore_no_transaction == true
- PB07 Absolute TxProfile && connector transaction != nullptr && transaction_id matches SKIPPED: was not able to test

Negative Boundary Conditions:
- NB01 Valid Profile, ConnectorID gt this->connectors.size()
- NB02 Valid Profile, ConnectorID lt 0
- NB03 profile.stackLevel lt 0
- NB04 profile.stackLevel gt profile_max_stack_level
- NB05 profile.chargingProfileKind == Absolute && !profile.chargingSchedule.startSchedule
- NB06 Number of installed Profiles is > max_charging_profiles_installed
- NB07 Invalid ChargingSchedule
- NB08 profile.chargingProfileKind == Recurring && !profile.recurrencyKind
- NB09 profile.chargingProfileKind == Recurring && !startSchedule
- NB10 profile.chargingProfileKind == Recurring && !startSchedule && !allow_charging_profile_without_start_schedule
- NB11 Absolute ChargePointMaxProfile Profile with connector id not 0
- NB12 Absolute TxProfile connector_id == 0  

## Add Profile Tests

- Add ChargePointMaxProfile
- Add ChargePointMaxProfile accepts a profile that is not a ChargePointMaxProfile
- Add TxDefaultProfile with ConnectorID of 0 Retrieved as ConnectorID of 1
- Add TxDefaultProfile with ConnectorID of 0 demonstrating that no profiles are returned with a ConnectorID of 0
- Add TxDefaultProfile with ConnectorID > 0
- Add TxDefaultProfile demonstrating how to trigger exception if the connector_id is greater than the number of connectors in the SmartChargingHandler's connectors map.
- Add TxProfile

## Steps to run:

```bash
cmake  -B build -G Ninja -DBUILD_TESTING=O -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="dist" -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
ninja -j$(nproc) -C build install
ninja -j$(nproc) -C build test 

```

### DIFF
--- a/tests/lib/ocpp/v16/test_smart_charging_handler.cpp
+++ b/tests/lib/ocpp/v16/test_smart_charging_handler.cpp
@@ -16,7 +16,7 @@ namespace v16 {
 /**
  * Chargepoint Test Fixture
  *
- * Test Matrix:
+ * Validate Profile Test Matrix:
  *
  * Positive Boundary Conditions:
  * - PB01 Valid Profile
@@ -201,6 +201,35 @@ protected:
         std::shared_ptr<DatabaseHandlerMock> database_handler =
             std::make_shared<DatabaseHandlerMock>(chargepoint_id, database_path, init_script_path);
         auto handler = new SmartChargingHandler(connectors, database_handler, true);
+        return handler;
+    }
+
+    SmartChargingHandler* createSmartChargingHandler(const int number_of_connectors) {
+        for (int i = 0; i <= number_of_connectors; i++) {
+            addConnector(i);
+        }
+
+        const std::string chargepoint_id = "1";
+        const fs::path database_path = "na";
+        const fs::path init_script_path = "na";
+
+        std::shared_ptr<DatabaseHandlerMock> database_handler =
+            std::make_shared<DatabaseHandlerMock>(chargepoint_id, database_path, init_script_path);
+
+        auto handler = new SmartChargingHandler(connectors, database_handler, true);
+
+        return handler;
+    }
+
+    SmartChargingHandler* createSmartChargingHandlerWithChargePointMaxProfile() {
+        auto profile = createChargingProfile(createChargeSchedule(ChargingRateUnit::A));
+        const std::vector<ChargingRateUnit>& charging_schedule_allowed_charging_rate_units{ChargingRateUnit::A};
+        auto handler = createSmartChargingHandler(10);
+
+        profile.chargingProfilePurpose = ChargingProfilePurposeType::ChargePointMaxProfile;
+        profile.chargingProfileKind = ChargingProfileKindType::Absolute;
+        handler->add_charge_point_max_profile(profile);
+
         return handler;
     }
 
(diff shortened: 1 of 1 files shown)


folkengine:
@shankari please review

tests/lib/ocpp/v16/test_smart_charging_handler.cpp
```diff
@@ -0,0 +1,764 @@
...
+ * - NB11 Absolute ChargePointMaxProfile Profile with connector id not 0
+ * - NB12 Absolute TxProfile connector_id == 0
+ */
+class ChargepointTestFixture : public testing::Test {
+protected:
+    void SetUp() override {
+    }
+
+    // Adding transactiosn to the connectors because Profiles other than ChargePointMaxProfile need it in order to be valid.
+    void addConnector(int id) {
```
shankari, line 50: It looks like these first few functions are utility functions that are used by the tests - I would suggest moving them to a `test_smart_charging_handler_utils.cpp` file to ensure that the test file only has tests.

I have not used `gtest` in particular before, so if this is the recommended practice for it, please add a link to the source of the recommendation.

tests/lib/ocpp/v16/test_smart_charging_handler.cpp
```diff
@@ -0,0 +1,764 @@
...
+    const int max_charging_profiles_installed = 20;
+    const int charging_schedule_max_periods = 10;
+};
+
+/**
+ * PB01 Valid Profile
+ *
+ * Happy path simple test
+ */
+TEST_F(ChargepointTestFixture, ValidateProfile) {
```
shankari, line 225: Since these are unit tests, I understand that they are linked to code functions. But it might also be helpful to indicate which functional tests the functions map to (and the unit tests influence) (e.g. A05)?

shankari (changes requested):
A few minor change, once they are done, I am happy to pass it over to the community

tests/lib/ocpp/v16/test_smart_charging_handler.cpp
folkengine: Done. Moved ChargepointTestFixture to its own file.

tests/lib/ocpp/v16/test_smart_charging_handler.cpp
folkengine: Since VSCode can find all that information with a right mouse click, I feel that adding these comments is unnecessary and can quickly become out of date if not careful.

shankari: - not everybody uses VSCode
- I am not sure how VSCode would find the match to the OCPP specification which is a pdf document that is not in the code.

Maybe you can share some screenshots to clarify?

couryrr-afs: @shankari The naming convention of the test is showing what functionality is being covered at a highlevel. I am not sure we have the test setup to the level of granularity of a functional requirement in the spec. The SUT itself is covering the functional requirement. Each conditional in there is probably more accurate for labeling like this. In general though comments like that can become stale pretty quickly.

In relation to the VS code context of finding the reference to a method call. That should apply to most modern IDEs.

shankari: > In general though comments like that can become stale pretty quickly.

I agree that in general it is better to not duplicate information (see https://github.com/EVerest/everest-utils/pull/111#discussion_r1434250662) 

But given that the goal of this library is to implement the spec, and the functional spec is in a pdf, I think it would be helpful to have explicit comments that link code to (potentially multiple) cases in the specification.

I don't see that they will get stale quickly since the spec is fixed - if we refactor code, part of the refactor is to keep these up to date.

I am open to suggestions on where the mapping can be maintained, but I think that it shouid be maintained.

shankari:
@folkengine this is also failing static code analysis now

folkengine:
> @folkengine this is also failing static code analysis now

@shankari Yeah, I've been trying to figure out how to turn off the lint, since it's actually wrong. Those fields are a part of the test fixture. Removing them makes the files not compile. Going to need to figure out how to refactor them without copying them over and over. Alas, their linter doesn't let you disable a lint for a single line.

folkengine:
@shankari unfortunately, the refactoring you requested in moving the GTest test fixture createde Codacy linting issues, saying that class fields weren't being used, even though they were. Could not find a way in their docs to ignore specific lints on certain lines unlike other linters. It seems that the tool has bugs dealing with Google Test. 

I have moved the fixture back to being inside the test to get past the issue. We can escalate this issue, but that will slow us down.

folkengine:
@shankari is this 👍?

shankari:
@folkengine you need to make sure that all the commits are signed

shankari (approved):
I am approving pending addressing the DCO

folkengine:
> I am approving pending addressing the DCO

Fixed. Thanks @shankari !

folkengine:
> @folkengine can you please resolve the conflicts that occured up after merging: #447

@Pietfried Tis' done. Thanks!

Pietfried (approved):

PR #443, opened 2024-02-02
Fixed reconnect attempt, fixed sys error on timer dtor

AssemblyJohn:


lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc, line 1: After a connection is closed (from with EVerest) is_marked_close is set to true.
If should probably be set to false when there is a call to connect()

james-ctc (changes requested):

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: new ConnectionData(); is done on connect() which ensures is_marked_close is false.

james-ctc (approved):

PR #448, opened 2024-02-05
Number of fixes to the Metervalue implementation for 2.0.1

marcemmers:
* Configurable rounding of clock aligned timestamps
* Filtering of tx ended meter values after the transaction has ended
* Correctly aligning the sampled data to the start of the transaction
* Clearing of accumulated metervalues before starting a transaction to clear out the idle metervalues
* Fix edge case where transaction start time is before an interval and code is run after an interval

### DIFF
--- a/config/v201/component_schemas/standardized/InternalCtrlr.json
+++ b/config/v201/component_schemas/standardized/InternalCtrlr.json
@@ -16,7 +16,7 @@
                   "mutability": "ReadWrite"
               }
           ],
-          "default": "1",
+          "default": true,
           "type": "boolean"
       },
       "ChargePointId": {
@@ -240,7 +240,7 @@
                   "mutability": "ReadOnly"
               }
           ],
-          "default": "1",
+          "default": true,
           "type": "boolean"
       },
       "LogMessages": {
@@ -255,7 +255,7 @@
                   "mutability": "ReadOnly"
               }
           ],
-          "default": "1",
+          "default": true,
           "type": "boolean"
       },
       "LogMessagesFormat": {
@@ -338,7 +338,7 @@
               }
           ],
           "description": "Use default verify paths for validating CSMS server certificate",
-          "default": "1",
+          "default": true,
           "type": "boolean"
       },
       "OcspRequestInterval": {
@@ -483,7 +483,7 @@
                 "mutability": "ReadOnly"
             }
         ],
-        "default": "0",
+        "default": false,
         "type": "boolean"
       },
       "MessageQueueSizeThreshold": {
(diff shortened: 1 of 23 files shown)


include/ocpp/v201/utils.hpp
```diff
@@ -52,6 +52,12 @@ std::string sha256(const std::string& str);
 /// @return A SHA256 hash string
 std::string generate_token_hash(const IdToken& token);
 
+/// @brief Align the clock aligned timestamps to the interval values
+/// @param timestamp the timestamp to align
+/// @param align_interval the clock aligned interval to align to since midnight 00:00
+/// @return DateTime type timestamp
```
Pietfried, line 58:
```suggestion
/// \brief Align the clock aligned timestamps to the interval values
/// \param timestamp the timestamp to align
/// \param align_interval the clock aligned interval to align to since midnight 00:00
/// \return DateTime type timestamp
```

config/v201/component_schemas/standardized/InternalCtrlr.json
```diff
@@ -537,6 +537,21 @@
           "description": "List of criteria supported for a get custom report. Enabled,Active,Problem,Available",
           "default": "Enabled,Active,Problem,Available",
           "type": "string"
+      },
+      "RoundClockAlignedTimestamps": {
+          "variable_name": "RoundClockAlignedTimestamps",
```
Pietfried, line 542: Please add a description for the variable

config/v201/component_schemas/standardized/InternalCtrlr.json
```diff
@@ -537,6 +537,22 @@
...
+              "dataType": "boolean"
+          },
+          "attributes": [
+              {
+                  "type": "Actual",
+                  "mutability": "ReadOnly"
+              }
+          ],
+          "description": "If enabled the metervalues configured with the AlignedDataCtrlr will be rounded to the exact time intervals",
+          "default": "0",
```
Pietfried, line 554:
```suggestion
          "default": true,
```
I just noticed that this is an issue at several other places

Pietfried (approved):

config/v201/component_schemas/standardized/InternalCtrlr.json
marcemmers: Yep that's what made me unsure if this was right. Shall I change those as well?

Also, I thought to leave it off by default or would you rather have it enabled?

Pietfried: Yes, would be great if you can change all of them, thanks :+1: . Leave it off by default is fine for me

Issue #361, opened 2023-12-20
Use Case K01: SetChargingProfile


RobertDeLeeuw:
SetChargingProfile already implemented for OCPP 1.6. Still needs to be implemented for OCPP 2.0.1.
Hopefully a lot of code can be reused/shared.

drmrd:
My team is actively working on this use case and others in the Smart Charging functional block.

PR #452, opened 2024-02-06
SmartChargingHandler: Add validation for TxProfiles

christopher-davis-afs:
Creates a skeleton for SmartChargingHandler and adds a method
to validate TxProfiles and TxProfiles only. This method
ensures that a given profile (assumed to be TxProfile) fits all of
the functional requirements specific to TxProfiles per K01.

The function, `validate_tx_profile ()`, takes a profile and an
EVSE, and returns a variant of `ProfileValidationResultEnum`.
This enum allows us to have detailed responses for the different types
of invalid profiles, and eventually will let us log *why* a profile
is invalid if the general validation fails.

Part of the prerequisites for https://github.com/EVerest/libocpp/issues/361

Related to https://github.com/EVerest/libocpp/issues/467 and https://github.com/EVerest/libocpp/issues/468

### DIFF
--- a/include/ocpp/v201/component_state_manager.hpp
+++ b/include/ocpp/v201/component_state_manager.hpp
@@ -23,12 +23,116 @@ struct FullConnectorStatus {
 
     /// \brief Translates the individual state to an Available/Unavailable/Occupied/Reserved/Faulted state
     /// This does NOT take into account the state of the EVSE or CS,
-    /// and is intended to be used internally by the ComponentStateManager.
+    /// and is intended to be used internally by the ComponentStateManagerInterface.
     ConnectorStatusEnum to_connector_status();
 };
 
+class ComponentStateManagerInterface {
+public:
+    virtual ~ComponentStateManagerInterface();
+
+    /// \brief Set a callback to be called when the effective Operative/Inoperative state of the CS changes.
+    virtual void set_cs_effective_availability_changed_callback(
+        const std::function<void(const OperationalStatusEnum new_status)>& callback) = 0;
+
+    /// \brief Set a callback to be called when the effective Operative/Inoperative state of an EVSE changes.
+    virtual void set_evse_effective_availability_changed_callback(
+        const std::function<void(const int32_t evse_id, const OperationalStatusEnum new_status)>& callback) = 0;
+
+    /// \brief Set a callback to be called when the effective Operative/Inoperative state of a connector changes.
+    virtual void set_connector_effective_availability_changed_callback(
+        const std::function<void(const int32_t evse_id, const int32_t connector_id,
+                                 const OperationalStatusEnum new_status)>& callback) = 0;
+
+    /// \brief Get the individual status (Operative/Inoperative) of the CS, as set by the CSMS
+    virtual OperationalStatusEnum get_cs_individual_operational_status() = 0;
+
+    /// \brief Get the individual status (Operative/Inoperative) of an EVSE, as set by the CSMS
+    /// Note: This is not the same as the effective status.
+    /// The EVSE might be effectively Inoperative if the CS is Inoperative.
+    virtual OperationalStatusEnum get_evse_individual_operational_status(int32_t evse_id) = 0;
+
+    /// \brief Get the individual status (Operative/Inoperative) of a connector, as set by the CSMS
+    /// Note: This is not the same as the effective status.
+    /// The connector might be effectively Inoperative if its EVSE or the CS is Inoperative.
+    virtual OperationalStatusEnum get_connector_individual_operational_status(int32_t evse_id,
+                                                                              int32_t connector_id) = 0;
+
+    /// \brief Get the individual status (Operative/Inoperative) of the CS, as persisted in the database
+    /// This status is restored after reboot, and differs from the individual status if non-persistent
+    /// status changes were made.
+    virtual OperationalStatusEnum get_cs_persisted_operational_status() = 0;
+
+    /// \brief Get the individual status (Operative/Inoperative) of an EVSE, as persisted in the database
+    /// This status is restored after reboot, and differs from the individual status if non-persistent
+    /// status changes were made.
+    virtual OperationalStatusEnum get_evse_persisted_operational_status(int32_t evse_id) = 0;
+
+    /// \brief Get the individual status (Operative/Inoperative) of a connector, as persisted in the database
+    /// This status is restored after reboot, and differs from the individual status if non-persistent
+    /// status changes were made.
+    virtual OperationalStatusEnum get_connector_persisted_operational_status(int32_t evse_id, int32_t connector_id) = 0;
+
+    /// \brief Set the individual status (Operative/Inoperative) of the CS
+    virtual void set_cs_individual_operational_status(OperationalStatusEnum new_status, bool persist) = 0;
+
+    /// \brief Set the individual status (Operative/Inoperative) of an EVSE
+    /// Note: This is not the same as the effective status.
+    /// The EVSE might be effectively Inoperative if the CS is Inoperative.
+    virtual void set_evse_individual_operational_status(int32_t evse_id, OperationalStatusEnum new_status,
+                                                        bool persist) = 0;
+
+    /// \brief Set the individual status (Operative/Inoperative) of a connector
+    /// Note: This is not the same as the effective status.
+    /// The connector might be effectively Inoperative if its EVSE or the CS is Inoperative.
+    virtual void set_connector_individual_operational_status(int32_t evse_id, int32_t connector_id,
+                                                             OperationalStatusEnum new_status, bool persist) = 0;
+
+    /// \brief Get the effective Operative/Inoperative status of an EVSE
+    /// This is computed from the EVSE's and the CS's individual statuses.
+    virtual OperationalStatusEnum get_evse_effective_operational_status(int32_t evse_id) = 0;
+
+    /// \brief Get the effective Operative/Inoperative status of a connector.
+    /// This is computed from the connector's, the EVSE's, and the CS's individual statuses.
+    virtual OperationalStatusEnum get_connector_effective_operational_status(int32_t evse_id, int32_t connector_id) = 0;
+
+    /// \brief Get the effective Available/Unavailable/Occupied/Faulted/Reserved status of a connector.
+    /// If the EVSE or the CS is Inoperative, the connector will be effectively Unavailable.
+    virtual ConnectorStatusEnum get_connector_effective_status(int32_t evse_id, int32_t connector_id) = 0;
+
+    /// \brief Update the state of the connector when plugged in or out
+    virtual void set_connector_occupied(int32_t evse_id, int32_t connector_id, bool is_occupied) = 0;
+
+    /// \brief Update the state of the connector when reservations are made or expire
+    virtual void set_connector_reserved(int32_t evse_id, int32_t connector_id, bool is_reserved) = 0;
+
+    /// \brief Update the state of the connector when errors are raised and cleared
+    virtual void set_connector_faulted(int32_t evse_id, int32_t connector_id, bool is_faulted) = 0;
+
+    /// \brief Update the state of the connector when unavailable or enabled
+    virtual void set_connector_unavailable(int32_t evse_id, int32_t connector_id, bool is_unavailable) = 0;
+
+    /// \brief Call the {cs, evse, connector}_effective_availability_changed_callback callback once for every component.
+    /// This is usually only done once on boot to notify the rest of the system what the state manager expects the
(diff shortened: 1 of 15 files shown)


christopher-davis-afs:
Codacy is saying a private member field (`charging_profiles`) is never used, but that's objectively false - it's accessed and modified from within our source file. There's nothing we can really do in this case, since in C++ member fields *must* be set up in the header files.

include/ocpp/v201/smart_charging.hpp
```diff
@@ -0,0 +1,51 @@
...
+    TxProfileEvseHasNoActiveTransaction,
+    TxProfileConflictingStackLevel
+};
+
+/// \brief This class handles and maintains incoming ChargingProfiles and contains the logic
+/// to calculate the composite schedules
+class SmartChargingHandler {
+private:
+    std::shared_ptr<ocpp::v201::DatabaseHandler> database_handler;
+    std::vector<ChargingProfile> charging_profiles;
```
drmrd, line 32: If this field being unused is intentional, you should be able to disable the Codacy error by suppressing this check inline.
```suggestion
    // cppcheck-suppress unusedStructMember
    std::vector<ChargingProfile> charging_profiles;
```

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -0,0 +1,375 @@
...
+            chargingSchedules,
+            {}, // transactionId
+            recurrencyKind,
+            {}, // validFrom
+            {}  // validTo
+        };
+    }
+
+    /**
+     * TxDefaultProfile, stack #1: time-of-day limitation to 2 kW, recurring every day from 17:00h to 20:00h.
```
drmrd, line 166: Are the comments in this file accurate? They appear to refer to `TxDefaultProfile`s instead of `TxProfile`.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -0,0 +1,375 @@
...
+     *
+     * This profile is Example #1 taken from the OCPP 2.0.1 Spec Part 2, page 241.
+     */
```
drmrd, line 169: Are these `TxDefaultProfile` builders being used? If not, can you remove them?

include/ocpp/v201/smart_charging.hpp
gberardi-pillar: But what if the field is being used and Codacy is confidently wrong?

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+TEST_F(ChargepointTestFixtureV201, IfTxProfileIsMissingTransactionId_ThenProfileIsInvalid) {
+    createEvseWithId(evseId);
+    auto profile = createTxProfileWithMissingTransactionId(createChargeSchedule(ChargingRateUnitEnum::A));
+    auto sut = handler.validate_tx_profile(profile, *evses[evseId]);
+
+    ASSERT_THAT(sut, testing::Eq(ProfileValidationResultEnum::TxProfileMissingTransactionId));
+}
+
+TEST_F(ChargepointTestFixtureV201, IfTxProfileHasEvseIdNotGreaterThanZero_ThenProfileIsInvalid) {
```
drmrd, line 197: This test verifies behavior from K01.FR.16. Can you indicate this in the test name? One option would be to prefix its name:
```
K01FR16_IfTxProfileHasEvseIdNotGreaterThanZero_ThenProfileIsInvalid
```

drmrd: The same request goes for the other tests in this PR verifying (partially or completely) behavior included in OCPP 2.0.1 functional requirements.

drmrd (changes requested):
Let's discuss tomorrow, but I think now is the time to norm on a naming convention for tests that verify FRs from the 2.0.1 spec.

drmrd:
Thanks for adding the FRs! I went ahead and added an issue that we can point to in the future regarding the functional requirement naming convention: #467.

drmrd:
@christopher-davis-afs Could you link to #467 and #468 from the PR description? They provide additional context for future developers that motivates how we are implementing functionality for 2.0.1.

christopher-davis-afs:
Done

shankari (approved):
Glad to see the new naming convention in place here!

.github/workflows/build_and_test.yaml
```diff
@@ -3,6 +3,7 @@
 
 name: Build and test libocpp
 on: 
+  push: {}
```
Pietfried, line 6: We dont want to run the action on every push so please remove this

include/ocpp/v201/component_state_manager.hpp
```diff
@@ -23,12 +23,102 @@ struct FullConnectorStatus {
...
-    /// and is intended to be used internally by the ComponentStateManager.
+    /// and is intended to be used internally by the ComponentStateManagerInterface.
     ConnectorStatusEnum to_connector_status();
 };
 
+class ComponentStateManagerInterface {
+public:
+    virtual ~ComponentStateManagerInterface();
+
+    virtual void set_cs_effective_availability_changed_callback(
```
Pietfried, line 34: Please add the method docs for this function

include/ocpp/v201/component_state_manager.hpp
```diff
@@ -23,12 +23,102 @@ struct FullConnectorStatus {
...
+    virtual void set_connector_effective_availability_changed_callback(
+        const std::function<void(const int32_t evse_id, const int32_t connector_id,
+                                 const OperationalStatusEnum new_status)>& callback) = 0;
+
+    /// \brief Get the individual status (Operative/Inoperative) of the CS, as set by the CSMS
+    virtual OperationalStatusEnum get_cs_individual_operational_status() = 0;
+    /// \brief Get the individual status (Operative/Inoperative) of an EVSE, as set by the CSMS
+    /// Note: This is not the same as the effective status.
+    /// The EVSE might be effectively Inoperative if the CS is Inoperative.
+    virtual OperationalStatusEnum get_evse_individual_operational_status(int32_t evse_id) = 0;
```
Pietfried, line 51: Please add newlines between function definition and method docs

Pietfried (changes requested):
This looks good to me except for some requested changes.

Two more general comments in addition to the inline comments:
In general we try to use snake_case for our variable and function definitions. I see some inconsistencies between snake_case and camelCase.

As already mentioned in the cloud working group meeting we would like to keep the scope of PRs as small as possible so we would rather see the ComponentStateManagerInterface refactor as a separate PR to the changes for SmartCharging. For this PR we can keep, but please consider this for future PRs.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -0,0 +1,276 @@
...
+#include <ocpp/v201/evse.hpp>
+#include <ocpp/v201/smart_charging.hpp>
+#include <optional>
+
+#include <sstream>
+
+namespace ocpp {
+namespace v201 {
+
+class DeviceModelStorageMock : public DeviceModelStorage {
```
marcemmers, line 24: Might be smart to put these mocks in their own include files so they can be reused?

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -0,0 +1,276 @@
...
+    SmartChargingHandler handler = createSmartChargingHandler();
+    boost::uuids::random_generator uuid_generator = boost::uuids::random_generator();
+};
+
+TEST_F(ChargepointTestFixtureV201, K01FR03_IfTxProfileIsMissingTransactionId_ThenProfileIsInvalid) {
+    createEvseWithId(evseId);
+    auto profile = createTxProfileWithMissingTransactionId(createChargeSchedule(ChargingRateUnitEnum::A));
+    auto sut = handler.validate_tx_profile(profile, *evses[evseId]);
+
+    ASSERT_THAT(sut, testing::Eq(ProfileValidationResultEnum::TxProfileMissingTransactionId));
```
marcemmers, line 194: Wouln't it be better to use `EXPECT_EQ` here so testing can continue if one of these fail? I think that's also preferred from google test

include/ocpp/v201/component_state_manager.hpp
```diff
@@ -23,12 +23,102 @@ struct FullConnectorStatus {
...
 };
 
+class ComponentStateManagerInterface {
+public:
+    virtual ~ComponentStateManagerInterface();
+
+    virtual void set_cs_effective_availability_changed_callback(
+        const std::function<void(const OperationalStatusEnum new_status)>& callback) = 0;
+
+    /// \brief Set a callback to be called when the effective Operative/Inoperative state of an EVSE changes.
```
marcemmers, line 37: I see we have duplicates now of the documentation. Is it an idea to only add documentation to either the interface or the implementation to keep it DRY?

marcemmers (changes requested):

include/ocpp/v201/smart_charging.hpp
```diff
@@ -0,0 +1,52 @@
...
+    // cppcheck-suppress unusedStructMember
+    std::vector<ChargingProfile> charging_profiles;
+
+public:
+    explicit SmartChargingHandler(std::shared_ptr<DatabaseHandler> database_handler);
+
+    ///
+    /// \brief validates the given \p profile according to the specification
+    ///
+    ProfileValidationResultEnum validate_tx_profile(ChargingProfile& profile, Evse& evse) const;
```
marcemmers, line 41: I think these can be const references which probably indicates better what is going on. Same for `add_profile`

include/ocpp/v201/component_state_manager.hpp
christopher-davis-afs: This change is just from refactoring the ComponentStateManager into an interface so that we could mock it. So, while we'd love to have documentation, we don't really have a good grasp on what this is supposed to do so that we can document it.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
christopher-davis-afs: Changed to `EXPECT_THAT`

include/ocpp/v201/component_state_manager.hpp
christopher-davis-afs: Turns out we missed the original docs. Copied those over.

include/ocpp/v201/component_state_manager.hpp
christopher-davis-afs: Removed the comments from the implementation.

include/ocpp/v201/smart_charging.hpp
christopher-davis-afs: We made the profile const, but the evse can't be const as the methods we use aren't const.

marcemmers: That does make sense, but then again, you aren't actually changing anything so it might make sense to add the const functions you need? Not sure if that should be in scope of this PR though..

include/ocpp/v201/smart_charging.hpp
```diff
@@ -0,0 +1,48 @@
...
+public:
+    explicit SmartChargingHandler();
+
+    ///
+    /// \brief validates the given \p profile according to the specification
+    ///
+    ProfileValidationResultEnum validate_tx_profile(const ChargingProfile& profile, Evse& evse) const;
+
+    /// \brief Adds a given \p profile to our stored list of profiles
+    void add_profile(ChargingProfile& profile);
```
marcemmers, line 43: I actually though you could make this one `const &` too

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
marcemmers: Any reason for preferring `EXPECT_THAT` above `EXPECT_EQ`?

gberardi-pillar: EXPECT_EQ() takes two parameters, but it isn't obvious which one is the expected and which one is the actual.

EXPECT_THAT() reads more naturally, and it is obvious which one is the expected value and which one is the actual value.

Pietfried (approved):

marcemmers (approved):

hikinggrass (approved):

drmrd (approved):

drmrd:
Thanks for the reviews, @Pietfried, @hikinggrass, and @marcemmers! When can we expect this to be merged?

PR #463, opened 2024-02-13
Reliability improvements for OpenSSL tpm2 provider on an embedded system

james-ctc:
related to [libevse-security PR42](https://github.com/EVerest/libevse-security/pull/42)

# Background

The existing implementation was occasionally failing especially when there was a migration from using non-TPM keys to supporting TPM protected keys (TSS2). Issues were observed during the loading and unloading of providers. There was also some concern that another call could impact which providers were loaded and their property strings.

There were significant improvements once the the tpm2-abrmd daemon was included and running. Generating keys, CSRs etc. didn't appear to need tpm2-abrmd however once TLS was being used tpm2-abrmd became essential.

It also became clear that the provider configuration depends on the key type. Hence the first line of the PEM key file is read so that the provider can be configured before actually loading the key.
Updates

Note: the tpm2-abrmd daemon needs to be running for tpm2 where TLS is being used.

# Functionality change

The meaning of UseTPM has been clarified to support interworking and upgrades from non-TPM keys to TPM protected keys.
For TLS the provider chosen is based on the PEM file and not the UseTPM OCPP configuration variable.
UseTPM is used to determine how a new key is to be generated and not how an existing key is used.

# Changes

The code has been updated to use the new OpenSSL provider implementation in libevse-security. It provides a suitable implementation for:
- OpenSSL v3
  - TPM protected keys
  - non-TPM protected keys
- OpenSSL v1 
  - non-TPM protected keys

### DIFF
--- a/lib/ocpp/common/websocket/websocket_tls_tpm.cpp
+++ b/lib/ocpp/common/websocket/websocket_tls_tpm.cpp
@@ -1,31 +1,31 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
+#include <evse_security/crypto/openssl/openssl_tpm.hpp>
 #include <ocpp/common/websocket/websocket_tls_tpm.hpp>
-
-#include <evse_security/detail/openssl/openssl_types.hpp>
 
 #include <everest/logging.hpp>
 
 #include <libwebsockets.h>
-#include <openssl/provider.h>
 
 #include <chrono>
 #include <memory>
 #include <stdexcept>
 #include <string>
 
+#include <openssl/opensslv.h>
+#include <openssl/ssl.h>
+#define USING_OPENSSL_3 (OPENSSL_VERSION_NUMBER >= 0x30000000L)
+
+#if USING_OPENSSL_3
+#include <openssl/provider.h>
+#else
+#define SSL_CTX_new_ex(LIB, PROP, METHOD) SSL_CTX_new(METHOD)
+#endif
+
 template <> class std::default_delete<lws_context> {
 public:
     void operator()(lws_context* ptr) const {
         ::lws_context_destroy(ptr);
-    }
-};
-
-template <> class std::default_delete<OSSL_LIB_CTX> {
-public:
-    void operator()(OSSL_LIB_CTX* ptr) const {
-        ::OPENSSL_thread_stop_ex(ptr);
-        ::OSSL_LIB_CTX_free(ptr);
     }
 };
 
@@ -37,6 +37,9 @@ public:
 };
 
 namespace ocpp {
+
+using evse_security::is_tpm_key_filename;
+using evse_security::OpenSSLProvider;
 
 enum class EConnectionState {
     INITIALIZE, ///< Initialization state
@@ -114,7 +117,6 @@ public:
     // This public block will only be used from client loop thread, no locking needed
     // Openssl context, must be destroyed in this order
     std::unique_ptr<SSL_CTX> sec_context;
-    std::unique_ptr<OSSL_LIB_CTX> sec_lib_context;
     // libwebsockets state
     std::unique_ptr<lws_context> lws_ctx;
 
@@ -211,57 +213,8 @@ constexpr auto local_protocol_name = "lws-everest-client";
 static const struct lws_protocols protocols[] = {{local_protocol_name, callback_minimal, 0, 0, 0, NULL, 0},
                                                  LWS_PROTOCOL_LIST_TERM};
 
-static void create_sec_context(bool use_tpm, OSSL_LIB_CTX*& out_libctx, SSL_CTX*& out_ctx) {
-    OSSL_LIB_CTX* libctx = OSSL_LIB_CTX_new();
-
-    if (libctx == nullptr) {
-        EVLOG_AND_THROW(std::runtime_error("Unable to create ssl lib ctx."));
-    }
-
-    out_libctx = libctx;
-
-    if (use_tpm) {
-        OSSL_PROVIDER* prov_tpm2 = nullptr;
-        OSSL_PROVIDER* prov_default = nullptr;
-
-        if ((prov_tpm2 = OSSL_PROVIDER_load(libctx, "tpm2")) == nullptr) {
-            EVLOG_AND_THROW(std::runtime_error("Could not load provider tpm2."));
-        }
-
-        if (!OSSL_PROVIDER_self_test(prov_tpm2)) {
-            EVLOG_AND_THROW(std::runtime_error("Could not self-test provider tpm2."));
-        }
-
-        if ((prov_default = OSSL_PROVIDER_load(libctx, "default")) == nullptr) {
-            EVLOG_AND_THROW(std::runtime_error("Could not load provider default."));
-        }
-
-        if (!OSSL_PROVIDER_self_test(prov_default)) {
-            EVLOG_AND_THROW(std::runtime_error("Could not self-test provider default."));
-        }
-    }
-
-    const SSL_METHOD* method = SSLv23_client_method();
-    SSL_CTX* ctx = SSL_CTX_new_ex(libctx, nullptr, method);
-
-    if (ctx == nullptr) {
-        EVLOG_AND_THROW(std::runtime_error("Unable to create ssl ctx."));
-    }
-
-    out_ctx = ctx;
-}
-
-void WebsocketTlsTPM::tls_init() {
-    SSL_CTX* ctx = nullptr;
-
-    if (auto* data = conn_data.get()) {
-        ctx = data->sec_context.get();
-    }
-
-    if (nullptr == ctx) {
-        EVLOG_AND_THROW(std::runtime_error("Invalid SSL context!"));
-    }
-
+void WebsocketTlsTPM::tls_init(SSL_CTX* ctx, const char* path_chain, const char* path_key, bool tpm_key,
+                               std::optional<std::string>& password) {
     auto rc = SSL_CTX_set_cipher_list(ctx, this->connection_options.supported_ciphers_12.c_str());
     if (rc != 1) {
         EVLOG_debug << "SSL_CTX_set_cipher_list return value: " << rc;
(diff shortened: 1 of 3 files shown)


AssemblyJohn:
One note related to the OpenSSL provider. At the moment we seem to use a global OpenSSL context using the OpenSSLProvider. I am not sure what side effects this could have. Would it be better to re-create the openssl context upon each connection?

james-ctc:
> … Would it be better to re-create the openssl context upon each connection?
Loading providers on demand was causing unreliability (in one instance there was a lockup while attempting to load a provider). The approach is based on loading providers as would be done in /etc/ssl/openssl.cfg; without needing to change the global configuration. Loading a provider is loading a DLL and running some initialisation and setup code. The SSL context for the connection doesn't change any of that, it just uses the loaded provider.

AssemblyJohn:
> > … Would it be better to re-create the openssl context upon each connection?
> > Loading providers on demand was causing unreliability (in one instance there was a lockup while attempting to load a provider). The approach is based on loading providers as would be done in /etc/ssl/openssl.cfg; without needing to change the global configuration. Loading a provider is loading a DLL and running some initialisation and setup code. The SSL context for the connection doesn't change any of that, it just uses the loaded provider.

I understand, probably that's why the old problems could arise, since loading/unloading the DLL can cause issues it seems. If there are checks so that concurrency problems do not arise (I've noticed the mutex) and also that there is no mis-calling of the wrong function (using the TPM instead of the normal openssl) then it's fine on my side.

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
```diff
@@ -276,29 +229,27 @@ void WebsocketTlsTPM::tls_init() {
...
 
-        path_chain = certificate_key_pair.value().certificate_path.c_str();
-        path_key = certificate_key_pair.value().key_path.c_str();
-
         if (1 != SSL_CTX_use_certificate_chain_file(ctx, path_chain)) {
+            ERR_print_errors_fp(stderr);
             EVLOG_AND_THROW(std::runtime_error("Could not use client certificate file within SSL context"));
         }
 
+        OpenSSLProvider provider;
```
AssemblyJohn, line 242: I couldn't find any usage for this. Isn't it used outside the function already?

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
```diff
@@ -398,24 +349,65 @@ void WebsocketTlsTPM::client_loop() {
...
-    // Connection acquire the contexts
-    conn_data->sec_lib_context = std::unique_ptr<OSSL_LIB_CTX>(lib_ctx);
-    conn_data->sec_context = std::unique_ptr<SSL_CTX>(ssl_ctx);
+        if (!certificate_key_pair.has_value()) {
+            EVLOG_AND_THROW(std::runtime_error(
+                "Connecting with security profile 3 but no client side certificate is present or valid"));
+        }
+
+        path_chain = std::string(certificate_key_pair.value().certificate_path.c_str());
+        path_key = std::string(certificate_key_pair.value().key_path.c_str());
```
hikinggrass, line 369: Couldn't you just use
```suggestion
        path_chain = certificate_key_pair.value().certificate_path;
        path_key = certificate_key_pair.value().key_path;
```
in this case?

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
```diff
@@ -398,24 +349,65 @@ void WebsocketTlsTPM::client_loop() {
...
+    }
+
+    SSL_CTX* ssl_ctx = nullptr;
+    bool tpm_key = false;
+
+    {
+        OpenSSLProvider provider;
+
+        if (!path_key.empty()) {
+            tpm_key = is_tpm_key_filename(path_key.c_str());
```
hikinggrass, line 380: I don't think this needs the .c_str()
```suggestion
            tpm_key = is_tpm_key_filename(path_key);
```

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
```diff
@@ -398,24 +349,65 @@ void WebsocketTlsTPM::client_loop() {
...
+
+        if (ssl_ctx == nullptr) {
+            ERR_print_errors_fp(stderr);
+            EVLOG_AND_THROW(std::runtime_error("Unable to create ssl ctx."));
+        }
+    }
 
     // Init TLS data
-    tls_init();
+    tls_init(ssl_ctx, path_chain.c_str(), path_key.c_str(), tpm_key, password);
```
hikinggrass, line 403: How about changing the function signature so you can pass a const string& here
```suggestion
    tls_init(ssl_ctx, path_chain, path_key, tpm_key, password);
```

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
```diff
@@ -211,57 +213,8 @@ constexpr auto local_protocol_name = "lws-everest-client";
...
-
-    if (auto* data = conn_data.get()) {
-        ctx = data->sec_context.get();
-    }
-
-    if (nullptr == ctx) {
-        EVLOG_AND_THROW(std::runtime_error("Invalid SSL context!"));
-    }
-
+void WebsocketTlsTPM::tls_init(SSL_CTX* ctx, const char* path_chain, const char* path_key, bool tpm_key,
```
hikinggrass, line 216:
```suggestion
void WebsocketTlsTPM::tls_init(SSL_CTX* ctx, const std::string& path_chain, const std::string& path_key, bool tpm_key,
```

and then do the conversion to const char* further down

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
```diff
@@ -276,29 +229,27 @@ void WebsocketTlsTPM::tls_init() {
...
-                "Connecting with security profile 3 but no client side certificate is present or valid"));
+        if ((path_chain == nullptr) || (path_key == nullptr)) {
+            EVLOG_error << "Cert chain: " << path_chain << " key: " << path_key;
+            EVLOG_AND_THROW(std::runtime_error("No certificate and key for SSL"));
         }
 
-        path_chain = certificate_key_pair.value().certificate_path.c_str();
-        path_key = certificate_key_pair.value().key_path.c_str();
-
         if (1 != SSL_CTX_use_certificate_chain_file(ctx, path_chain)) {
```
hikinggrass, line 237:
```suggestion
        if (1 != SSL_CTX_use_certificate_chain_file(ctx, path_chain.c_str())) {
```

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
```diff
@@ -276,29 +229,27 @@ void WebsocketTlsTPM::tls_init() {
...
+            ERR_print_errors_fp(stderr);
             EVLOG_AND_THROW(std::runtime_error("Could not use client certificate file within SSL context"));
         }
 
+        OpenSSLProvider provider;
+
+        SSL_CTX_set_default_passwd_cb_userdata(ctx, reinterpret_cast<void*>(password.value_or("").data()));
+
         if (1 != SSL_CTX_use_PrivateKey_file(ctx, path_key, SSL_FILETYPE_PEM)) {
```
hikinggrass, line 246:
```suggestion
        if (1 != SSL_CTX_use_PrivateKey_file(ctx, path_key.c_str(), SSL_FILETYPE_PEM)) {
```

hikinggrass (changes requested):

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: it adds no value any more - removing

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
```diff
@@ -398,24 +349,65 @@ void WebsocketTlsTPM::client_loop() {
...
+        path_chain = std::string(certificate_key_pair.value().certificate_path.c_str());
+        path_key = std::string(certificate_key_pair.value().key_path.c_str());
+        password = certificate_key_pair.value().password;
+    }
+
+    SSL_CTX* ssl_ctx = nullptr;
+    bool tpm_key = false;
+
+    {
+        OpenSSLProvider provider;
```
AssemblyJohn, line 377: Question: the propquery, who sets it here?

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
```diff
@@ -398,24 +349,65 @@ void WebsocketTlsTPM::client_loop() {
...
 
-    // Connection acquire the contexts
-    conn_data->sec_lib_context = std::unique_ptr<OSSL_LIB_CTX>(lib_ctx);
-    conn_data->sec_context = std::unique_ptr<SSL_CTX>(ssl_ctx);
+        if (!certificate_key_pair.has_value()) {
+            EVLOG_AND_THROW(std::runtime_error(
+                "Connecting with security profile 3 but no client side certificate is present or valid"));
+        }
+
+        path_chain = std::string(certificate_key_pair.value().certificate_path.c_str());
```
AssemblyJohn, line 368: There is a possible error here. The get_key_pair can return 2 values for a valid certificate, one is certificate_path and another is 'certificate_single_path'. if the certificate_path there should be a fallback to the 'certificate_single_path'.

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: it's done a few lines later

AssemblyJohn: No, I meant who sets: 

static const char* s_default_provider = PROPQUERY_DEFAULT;
static const char* s_tpm2_provider = PROPQUERY_TPM2;

I know in libevse it's done in the CMAKE. Are those defaults used? Are they configurable for libocpp too? Or they're common? How is that handled when we require TPM with libevse AND libocpp?

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: I'll make the change.
However get_key_pair() could be updated so that this isn't required.
What if both certificate and certificate_single have values?
Is there a use case where certificate_single is needed on its own?

AssemblyJohn: Yes, there is a case with symlinks in the library. Because most of the times both will be present (chain & single) it is better to have this as both, and one can be empty. If both are non-null then the usage depends on the certificate setup.

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: The property strings are only in libevse-security. Users of libevse-security chose the settings to apply via OpenSSLProvider::mode_t. This deliberately prevents setting other property strings and causing runtime issues.
Hence libocpp just needs to decide to use mode_t::default_provider or mode_t::tpm2_provider
The coder doesn't need to know how that happens (within libevse-security)

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: updated

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: updated

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: updated

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: updated

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: updated

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: updated

AssemblyJohn (approved):

hikinggrass (approved):

PR #466, opened 2024-02-13
Special handling for value 0 for certain variables

Pietfried:
Allow value of 0 for AlignedDataInterval, SampledDataTxUpdatedInterval, SampledDataTxEndedInterval

### DIFF
--- a/lib/ocpp/v201/device_model.cpp
+++ b/lib/ocpp/v201/device_model.cpp
@@ -10,6 +10,14 @@ namespace ocpp {
 
 namespace v201 {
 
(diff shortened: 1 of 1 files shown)


hikinggrass (approved):

marcemmers:
I don't really get why it wasn't already possible to configure 0 to these values?

hikinggrass:
> I don't really get why it wasn't already possible to configure 0 to these values? If it was failing on the minlimit shouldn't that be set to 0 then? Or do you want to only allow 0 and something higher than a certain value?

Yes that's it, because 0 has a "special meaning" for these variables. When you want to use a minLimit > 0 but still want to allow the feature to be disabled with setting the variable to 0

PR #483, opened 2024-02-20
CiString conversion fixes for GCC<9

hikinggrass:
## Describe your changes
Fix compilation issues with GCC 8

### DIFF
--- a/include/ocpp/common/cistring.hpp
+++ b/include/ocpp/common/cistring.hpp
@@ -23,6 +23,12 @@ public:
     CiString(const char* data) : String<L>(data) {
     }
 
+    CiString(json data) : String<L>(data.get<std::string>()) {
+    }
+
+    template <size_t LL> CiString(const CiString<LL>& data) : String<L>(data.get()) {
+    }
+
     /// \brief Creates a string
     CiString() : String<L>() {
     }
@@ -39,7 +45,7 @@ public:
     }
 
     /// \brief Conversion operator to turn a String into std::string
-    operator std::string() {
+    operator std::string() const {
         return this->get();
     }
 };
--- a/lib/ocpp/v16/charge_point_configuration.cpp
+++ b/lib/ocpp/v16/charge_point_configuration.cpp
@@ -219,47 +219,47 @@ std::string ChargePointConfiguration::getChargeBoxSerialNumber() {
 }
 
 CiString<20> ChargePointConfiguration::getChargePointModel() {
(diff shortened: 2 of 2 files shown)


include/ocpp/common/cistring.hpp
```diff
@@ -23,6 +23,12 @@ template <size_t L> class CiString : public String<L> {
     CiString(const char* data) : String<L>(data) {
     }
 
+    CiString(json data) : String<L>(data.get<std::string>()) {
```
marcemmers, line 26: I'm not sure about having this conversion implicit since it is quite a big step from json to a CiString.

include/ocpp/common/cistring.hpp
```diff
@@ -23,6 +23,12 @@ template <size_t L> class CiString : public String<L> {
...
+    }
+
+    template <size_t LL> CiString(const CiString<LL>& data) : String<L>(data.get()) {
```
marcemmers, line 29: So this allows us to assign strings of a different length to this type, is that intended?

lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -219,47 +219,47 @@ std::string ChargePointConfiguration::getChargeBoxSerialNumber() {
 }
 
 CiString<20> ChargePointConfiguration::getChargePointModel() {
-    return CiString<20>(this->config["Internal"]["ChargePointModel"]);
+    return CiString<20>(this->config["Internal"]["ChargePointModel"].get<std::string>());
```
marcemmers, line 222: Can't you use the new constructor for this now? I'm not sure if both changes are needed.

include/ocpp/common/cistring.hpp
hikinggrass: You are right, I removed this constructor again because it's a bit too drastic

lib/ocpp/v16/charge_point_configuration.cpp
hikinggrass: I've removed the json constructor, so these changes make a lot more sense now

include/ocpp/common/cistring.hpp
hikinggrass: That's really not what we want, I removed that again

marcemmers (approved):
Now it makes a bunch more sense indeed ;)

Pietfried (approved):

Issue #480, opened 2024-02-19
Resume interrupted transactions due to system failure


SNSubramanya:
### OCPP Version

OCPP2.0.1

### Describe the problem

Currently the stack doesn't provide a means to resume interrupted transactions due to a system failure or power outage. This feature will address this issue. 

### Describe your solution

- Store relevant details of the transaction in the database
- check for any ongoing transactions that were interrupted in the database on bootup and resume them.
- refactored code to restart metering timers.

PR #501, opened 2024-02-28
Resume interrupted transactions

SNSubramanya:
## Describe your changes
- Added Documentation
- Store details of transactions in database
- Restore details any interrupted transaction  on bootup
- Resume any interrupted transactions after boot notification accepted.
- Refactor some code to restart metering timers
https://github.com/EVerest/libocpp/issues/480


config/v201/init_core.sql
```diff
@@ -17,6 +17,18 @@ CREATE TABLE IF NOT EXISTS AVAILABILITY(
     CHECK (EVSE_ID >= 0 AND CONNECTOR_ID >= 0 AND (EVSE_ID > 0 OR CONNECTOR_ID = 0))
 );
 
+-- Transactions --
+CREATE TABLE IF NOT EXISTS TRANSACTIONS (
+    TRANSACTION_ID TEXT PRIMARY KEY NOT NULL,
+    SEQ_NO INT NOT NULL,
+    MESSAGE_TYPE TEXT NOT NULL,
```
marcemmers, line 24: `MESSAGE_TYPE` is not used and I expected the start meter value to be in here for recovery.

config/v201/init_core.sql
```diff
@@ -17,6 +17,18 @@ CREATE TABLE IF NOT EXISTS AVAILABILITY(
...
+    EVSE_ID INT NOT NULL,
+    CONNECTOR_ID INT NOT NULL,
+    ID_TOKEN TEXT NOT NULL,
+    TIME_START TEXT NOT NULL,
```
marcemmers, line 28: The time can be an INT64 type. There is a automatic conversion from Datetime to int64 in `SQliteStatement` class.

lib/ocpp/v201/database_handler.cpp
```diff
@@ -41,9 +41,18 @@ void DatabaseHandler::sql_init() {
         throw std::runtime_error("Database access error");
     }
 
-    this->inintialize_enum_tables();
-}
+    // process any interrupted transactions before initializing
+    process_interrupted_transactions();
```
marcemmers, line 45: I know you have to do this to fix the issue I created with this init table but I don't like that fact that we have to do this init here. I will think of a better solution to handle the `inintialize_enum_tables` so we don't have to do this step.

include/ocpp/v201/database_handler.hpp
```diff
@@ -128,6 +135,35 @@ class DatabaseHandler : public ocpp::common::DatabaseHandlerBase {
...
+    bool clear_transaction(const std::string& transaction_id);
+
+    /// @brief Get any interrupted transactions that hasn't ended.
+    /// @return TransactionInterruptedResponse
+    std::vector<TransactionInterruptedResponse> get_ongoing_transactions();
+
+    /// @brief Update the sequence number of the given transaction id in the database to retrieve later.
+    /// @param transaction_id 
+    /// @param seq_no 
+    void update_transaction_seq_no(const std::string& transaction_id, int32_t seq_no);
```
marcemmers, line 166: Shouldn't there be a function for updating the charging state too?

marcemmers (changes requested):
I think we can still improve this a bit. There is a lot to be gained by moving the database actions to the `EnhancedTransaction` I think.

lib/ocpp/v201/database_handler.cpp
```diff
@@ -227,6 +236,41 @@ void DatabaseHandler::insert_availability(int32_t evse_id, int32_t connector_id,
     }
 }
 
+void DatabaseHandler::process_interrupted_transactions() {
```
marcemmers, line 239: I think it would be better to replace this function with a `get_transaction` function with an `evse_id` argument that then returns an optional result. Maybe you could even return an enhanced transaction type so we don't need the `InterruptedTransaction` type anymore.

This could simplify the continuation of transaction in the Evse a bit. It also removes the need to remember these transactions in the database handlers memory like we have it now.

lib/ocpp/v201/database_handler.cpp
marcemmers: What you could do here is do a simple `SELECT` statement of the `TRANSACTIONS` table without any criteria. If you get a `SQLITE_ROW` back you know there are transactions. If you get `SQLITE_DONE` back immediately you'd know there are no transactions.

include/ocpp/v201/transaction.hpp
```diff
@@ -19,14 +20,23 @@ struct EnhancedTransaction : public Transaction {
     int32_t seq_no = 0;
     std::optional<float> active_energy_import_start_value;
     bool check_max_active_import_energy;
+    std::shared_ptr<DatabaseHandler> enhanced_transaction_database_handler;
```
marcemmers, line 23:
```suggestion
    std::shared_ptr<DatabaseHandler> database_handler;
```

Just this should be fine. It's already part of the enhanced transaction

doc/resumeInterruptedTransactions/resumeInterruptedTransactions.puml
```diff
@@ -0,0 +1,42 @@
...
+'https://plantuml.com/sequence-diagram
+!pragma teoz true
+participant csms order 10
+participant libocpp order 20
+participant core order 30
+actor user order 40 
+autonumber "<b><font color=red>"
+skinparam sequenceArrowThickness 2
+
+== Start a transaction ==
```
marcemmers, line 11: Might have to update this one

lib/ocpp/v201/charge_point.cpp
```diff
@@ -2622,14 +2617,14 @@ void ChargePoint::handle_transaction_event_response(const EnhancedMessage<v201::
...
         // add every evse_id with active transaction and given token
-        for (const auto& [evse_id, evse] : this->evses) {
-            const auto& transaction = evse->get_transaction();
-            if (transaction != nullptr and transaction->id_token.has_value()) {
-                if (transaction->id_token.value().idToken.get() == id_token.idToken.get()) {
-                    evse_ids.push_back(evse_id);
-                }
-            }
-        }
+        // for (const auto& [evse_id, evse] : this->evses) {
```
marcemmers, line 2620: Still not really sure what to do with this. It can probably be removed @Pietfried ?

marcemmers:
This PR is blocked by the smart charging unit tests. I can fix the compilation but the test still fails because the `Evse` now tries to use the database and there is only an uninitialized on in the test.

Instead of spending time to fix it we should no longer use the `Evse` class directly and start using mocks/fakes for this. This PR would help a lot: **https://github.com/EVerest/libocpp/pull/668**

marcemmers:
@hikinggrass This is ready for review now

tests/lib/ocpp/v201/mocks/evse_manager_fake.hpp
```diff
@@ -56,7 +56,8 @@ class EvseManagerFake : public EvseManagerInterface {
 
     void open_transaction(int evse_id, const std::string& transaction_id) {
         auto& transaction = this->transactions.at(evse_id - 1);
-        transaction = std::make_unique<EnhancedTransaction>();
+        // Really dirty way of fixing unit test. Make proper interface for transactions to solve this
```
hikinggrass, line 59: :scream:

hikinggrass (commented):
Can we make the resumption of transaction optional based on a configuration option?

tests/lib/ocpp/v201/mocks/evse_manager_fake.hpp
marcemmers: Found another dirty but less dirty way. I have some ideas on how to improve the transaction class that will solve this but that is too much for this PR.

marcemmers:
> Can we make the resumption of transactions optional based on a configuration option?

Yes good point, I added this

lib/ocpp/v201/database_handler.cpp
```diff
@@ -616,5 +621,102 @@ OperationalStatusEnum DatabaseHandler::get_connector_availability(int32_t evse_i
...
+    // Fill transaction
+    transaction->transactionId = get_stmt->column_text(0);
+    transaction->seq_no = get_stmt->column_int(1);
+    transaction->connector_id = get_stmt->column_int(3);
+    transaction->start_time = get_stmt->column_datetime(4);
+    transaction->chargingState = conversions::string_to_charging_state_enum(get_stmt->column_text(5));
+    transaction->id_token_sent = get_stmt->column_int(6) != 0;
+
+    if (get_stmt->step() == SQLITE_ROW) {
+        // Something strange is happening with multiple transactions
```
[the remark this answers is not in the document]
marcemmers: This happens when there are 2 transactions in the database. It should never happen but there is no guarantee of that and this is also not the right way to solve it I would say, since you can come back from this anymore. (no way to get a transaction means no way to delete one either).

I'll take a look on how to improve this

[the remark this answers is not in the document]
marcemmers: Should be fixed now, as well as the proper exception handling

hikinggrass (approved):

marcemmers (approved):

PR #513, opened 2024-03-05
fix: libwebsockets retry on no-connection

james-ctc:
## Describe your changes

When there is no internet connection then DNS lookups and other IP activities can take a while to return. A 60 second connect timeout was expiring and reconnect attempts were not scheduled.

Code updated to ensure that reconnection attempts are still tried.

fix: shared pointer conn_data not always checked for nullptr

Code updated to make con_data thread-safe and ensure that nullptr checks are applied.

- [X] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

### DIFF
--- a/include/ocpp/common/websocket/websocket_libwebsockets.hpp
+++ b/include/ocpp/common/websocket/websocket_libwebsockets.hpp
@@ -85,7 +85,6 @@ private:
     Everest::SteadyTimer reconnect_timer_tpm;
     std::unique_ptr<std::thread> websocket_thread;
     std::shared_ptr<ConnectionData> conn_data;
-    std::mutex conn_data_mutex;
     std::condition_variable conn_cv;
 
     std::mutex queue_mutex;
--- a/include/ocpp/common/websocket/websocket_libwebsockets.hpp
+++ b/include/ocpp/common/websocket/websocket_libwebsockets.hpp
@@ -7,6 +7,8 @@
 #include <ocpp/common/websocket/websocket_base.hpp>
 
 #include <condition_variable>
+#include <memory>
+#include <mutex>
 #include <optional>
 #include <queue>
 #include <string>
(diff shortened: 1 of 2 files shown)


lib/ocpp/common/websocket/websocket_tls_tpm.cpp
```diff
@@ -320,10 +324,11 @@ void WebsocketTlsTPM::recv_loop() {
 }
 
 void WebsocketTlsTPM::client_loop() {
+    conn_data_mutex.lock();
```
AssemblyJohn, line 327: Question: instead of using manually lock/unlock can we define it in a local scope? The local data can be defined outside that mentioned scope.

AssemblyJohn (approved):

lib/ocpp/common/websocket/websocket_tls_tpm.cpp
james-ctc: I've re-read the C++ spec on shared_ptr and thread safety and the mutex can be removed providing that all access is via a separate shared pointer.
i.e. the following would have problems since `conn_data` could change between the nullptr check and the call to do_interrupt()
```
        if (conn_data != nullptr) {
            // another thread could call conn_data.reset() and 
            // the next line would dereference nullptr
            conn_data->do_interrupt();
        }
        conn_data.reset();
```
This is fine since the check is on a separate shared_ptr that can't be reset between operations
```
        std::shared_ptr<ConnectionData> local_data = conn_data;
        if (local_data != nullptr) {
            // another thread could call conn_data.reset() which doesn't 
            // impact local_data other than the reference count
            local_data->do_interrupt();
        }
        conn_data.reset();
```

Updating the code with the new approach

AssemblyJohn (approved):

AssemblyJohn:
All cool, can be merged in the current libwebsocket beta.

Pietfried (approved):

Issue #494, opened 2024-02-26
libocpp verify_csms_cn seems to be unable to handle wildcard domain certificate


lategoodbye:
### OCPP Version

OCPP1.6, OCPP2.0.1

### Describe the bug

Our customer tried to use our EVerest image (EVerest-core 2024.01) with Security profile 2, but it fails during the CSMS CN validation.

```
Feb 25 10:47:02 tarragon manager[4859]: 2024-02-25 10:47:02.344092 [INFO] OCPP0:OCPP       :: Connecting TLS websocket to uri: wss://cpc.gridfluxx.com/gridfluxx/????????? with security-profile 2
Feb 25 10:47:02 tarragon manager[4864]: 2024-02-25 10:47:02.409081 [INFO] cb_tarragon_dri  :: handle_pwm_off: Setting new duty cycle of 100.00%
Feb 25 10:47:02 tarragon manager[4859]: 2024-02-25 10:47:02.566583 [INFO] OCPP0:OCPP       :: Found 0 charging profile(s) in the database
Feb 25 10:47:02 tarragon manager[4859]: 2024-02-25 10:47:02.572983 [INFO] OCPP0:OCPP       :: OCPP initialized
Feb 25 10:47:02 tarragon manager[4859]: 2024-02-25 10:47:02.577917 [INFO] OCPP0:OCPP       :: Processing queued session event for evse_id: 1, event: Enabled
Feb 25 10:47:02 tarragon manager[4859]: 2024-02-25 10:47:02.584815 [INFO] OCPP0:OCPP       :: Processing queued session event for evse_id: 1, event: Enabled
Feb 25 10:47:02 tarragon manager[4864]: 2024-02-25 10:47:02.669221 [INFO] cb_tarragon_dri  :: handle_pwm_off: Setting new duty cycle of 100.00%
Feb 25 10:47:02 tarragon manager[4865]: 2024-02-25 10:47:02.707731 [WARN] connector_1:Evs virtual bool module::evse::evse_managerImpl::handle_external_ready_to_start_charging() :: Ignoring external ready to start charging command, this could be a configuration issue. Please check if 'external_ready_to_start_charging' is set to true if you want to use this feature.
Feb 25 10:47:02 tarragon manager[4798]: 2024-02-25 10:47:02.822629 [INFO] manager          :: There are no errors of type evse_manager/MREC4OverCurrentFailure and of module connector_1->evse, so no errors are cleared
[...]

PR #516, opened 2024-03-06
Bugfix/494 libocpp verify csms cn wildcard

AssemblyJohn:

resolves: https://github.com/EVerest/libocpp/issues/494


config/v16/profile_schemas/Internal.json
```diff
@@ -196,6 +196,12 @@
             "readOnly": true,
             "default": true
         },
+        "VerifyCsmsAllowWildcards": {
+            "$comment": "Allow wildcards when verifying the CSMS commonName",
+            "type": "boolean",
+            "readOnly": true,
```
Pietfried, line 202:
```suggestion
            "readOnly": false,
```

config/v201/component_schemas/standardized/InternalCtrlr.json
```diff
@@ -341,6 +341,51 @@
...
+      "VerifyCsmsCommonName": {
+        "variable_name": "VerifyCsmsCommonName",
+        "characteristics": {
+            "supportsMonitoring": false,
+            "dataType": "boolean"
+        },
+        "attributes": [
+            {
+                "type": "Actual",
+                "mutability": "ReadOnly"
```
Pietfried, line 353:
```suggestion
                "mutability": "ReadWrite"
```

config/v201/component_schemas/standardized/InternalCtrlr.json
```diff
@@ -341,6 +341,51 @@
...
+      "VerifyCsmsAllowWildcards": {
+        "variable_name": "VerifyCsmsAllowWildcards",
+        "characteristics": {
+            "supportsMonitoring": false,
+            "dataType": "boolean"
+        },
+        "attributes": [
+            {
+                "type": "Actual",
+                "mutability": "ReadOnly"
```
Pietfried, line 383:
```suggestion
                "mutability": "ReadWrite"
```

config/v16/profile_schemas/Internal.json
Pietfried: I think we should make `VerifyCsmsAllowWildcards` and `VerifyCsmsCommonName` RW

include/ocpp/v16/charge_point_configuration.hpp
```diff
@@ -84,6 +84,7 @@ class ChargePointConfiguration {
     KeyValue getUseSslDefaultVerifyPathsKeyValue();
     bool getVerifyCsmsCommonName();
     KeyValue getVerifyCsmsCommonNameKeyValue();
+    bool getVerifyCsmsAllowWildcards();
```
Pietfried, line 87:
```suggestion
    bool getVerifyCsmsAllowWildcards();
    KeyValue getVerifyCsmsAllowWildcardsKeyValue();
```

Pietfried: Please also implement the `setXXX` functions when we turn these to RW and add the respective handling here:
* https://github.com/EVerest/libocpp/blob/main/lib/ocpp/v16/charge_point_configuration.cpp#L2176
* https://github.com/EVerest/libocpp/blob/main/lib/ocpp/v16/charge_point_configuration.cpp#L2478

Pietfried (changes requested):

include/ocpp/v16/charge_point_configuration.hpp
AssemblyJohn: Are you sure the 'set' is required? This doesn't seem to be the case based on the spec (a.k.a the csms will not request it) nor could I find any example of others being set like the 'VerifyCsmsCommonName'.

hikinggrass: There's an argument for both variants. If we see these variables as something purely used for debugging we should make them _Readonly_ and only provide getters.
If we want this to be set by a csms (not sure about that) we should make them _ReadWrite_ and provide getters and setters

AssemblyJohn: > There's an argument for both variants. If we see these variables as something purely used for debugging we should make them _Readonly_ and only provide getters. If we want this to be set by a csms (not sure about that) we should make them _ReadWrite_ and provide getters and setters

It doesn't seem that they should be set by the CSMS, they're internal anyway (I've looked over the spec). While not sure about the 'VerifyCsmsCommonName' the wildcard is certainly not set by the CSMS. If i'm wrong, please provide me the spec part.

Pietfried: To me it looks like wildcard certificates are often used in production systems, so I would expect `VerifyCsmsAllowWildcard` to be used not only for debugging. It provides more flexibility to the CSMS when e.g. migrating to another CSMS (that uses wildcard certificates). For that reason I would make them RW, altough I see that this would drop the level of security

Pietfried: > > There's an argument for both variants. If we see these variables as something purely used for debugging we should make them _Readonly_ and only provide getters. If we want this to be set by a csms (not sure about that) we should make them _ReadWrite_ and provide getters and setters
> 
> It doesn't seem that they should be set by the CSMS, they're internal anyway (I've looked over the spec). While not sure about the 'VerifyCsmsCommonName' the wildcard is certainly not set by the CSMS. If i'm wrong, please provide me the spec part.

Just because they are not in the spec it doesnt mean that we can not make internal configuration keys ReadWrite

AssemblyJohn: The configuration setting has been added.

lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -2836,6 +2861,14 @@ ConfigurationStatus ChargePointConfiguration::set(CiString<50> key, CiString<500
...
     }
 
+    if (key == "VerifyCsmsAllowWildcards") {
+        if (isBool(value.get())) {
+            this->setVerifyCsmsAllowWildcards(ocpp::conversions::string_to_bool(value.get()));
+        } else {
+            return ConfigurationStatus::Rejected;
+        }
+    }
+
```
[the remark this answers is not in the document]
AssemblyJohn: Removed config to set 'VerifyCsmsCommonName' since it creates a security issue.

lib/ocpp/common/websocket/websocket_tls.cpp
```diff
@@ -302,15 +304,34 @@ tls_context WebsocketTLS::on_tls_init(std::string hostname, websocketpp::connect
         }
 
         context->set_verify_mode(boost::asio::ssl::verify_peer);
+
         if (this->connection_options.verify_csms_common_name) {
```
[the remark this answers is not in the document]
AssemblyJohn: Issues with the libwebsocket impl, the hostname verification is skipped, looking into the issue.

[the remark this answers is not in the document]
AssemblyJohn: The issue was too long to look into, an workaround has been added, now the check is done manually since it seems that libwebsockets either skips certificate CN verification or it over-writes the verification parameter. 'X509_check_host' is manually used until a proper fix will be found.

However, with the manual verification in place, a very clear error message can be print.

config/v16/profile_schemas/Internal.json
```diff
@@ -193,9 +193,15 @@
         "VerifyCsmsCommonName": {
             "$comment": "Verify that the CSMS certificates commonName matches the CSMS FQDN",
             "type": "boolean",
-            "readOnly": true,
+            "readOnly": false,
```
hikinggrass, line 196: This should stay as readOnly = true

config/v201/component_schemas/standardized/InternalCtrlr.json
```diff
@@ -341,6 +341,51 @@
...
+      "VerifyCsmsCommonName": {
+        "variable_name": "VerifyCsmsCommonName",
+        "characteristics": {
+            "supportsMonitoring": false,
+            "dataType": "boolean"
+        },
+        "attributes": [
+            {
+                "type": "Actual",
+                "mutability": "ReadWrite"
```
hikinggrass, line 353: This should be ReadOnly

hikinggrass (changes requested):

AssemblyJohn:
Note: when trying to use X509_VERIFY_PARAM_set1_host and SSL_CTX_set1_param(ctx, param); for the current context in the 'tls_init' function it seems that libwebsockets does not take into account the set parameter and any server certificate passes the check.

Because of that, at the moment the server certificate's host is manually checked using the 'X509_check_host' routine. When we can see why libwebsockets does not take into account the automatic parameter check, a new fix should be introduced, removing the manual X509_check_host.

hikinggrass (approved):

PR #522, opened 2024-03-08
Add unit tests for using components as keys in map

christopher-davis-afs:
## Describe your changes

These tests help us understand how components are
found in e.g. DeviceModelMap.

### DIFF
--- a/tests/lib/ocpp/v201/test_device_model.cpp
+++ b/tests/lib/ocpp/v201/test_device_model.cpp
@@ -45,5 +45,40 @@ TEST_F(DeviceModelTest, test_allow_zero) {
     ASSERT_EQ(r, 0);
 }
 
+TEST_F(DeviceModelTest, test_component_as_key_in_map) {
+    std::map<Component, int32_t> components_to_ints;
+
+    const Component base_comp = {.name = "Foo"};
+    components_to_ints[base_comp] = 1;
+
+    const Component different_instance_comp = {
+        .name = "Foo",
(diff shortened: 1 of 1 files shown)


christopher-davis-afs:
CC @shankari

Pietfried (approved):

tests/lib/ocpp/v201/test_device_model.cpp
```diff
@@ -45,5 +45,40 @@ TEST_F(DeviceModelTest, test_allow_zero) {
...
+    };
+    const Component comp_with_custom_data = {
+        .name = "Foo",
+        .customData = CustomData{.vendorId = "Baz"},
+    };
+    const Component different_name_comp = {
+        .name = "Bar",
+    };
+
+    EXPECT_EQ(components_to_ints.find(base_comp)->second, 1);
```
marcemmers, line 75: What would happen here if `base_comp` could not be found in the map?

marcemmers (approved):
Looks good, just wondering about a small thing we might want to look at.

tests/lib/ocpp/v201/test_device_model.cpp
christopher-davis-afs: If the component isn't found and we access `->second` without checking, it would be undefined behavior from what I understand. Here in the test I don't check because `base_comp` is in the map.

drmrd:
Contributes to #361

PR #524, opened 2024-03-11
Validate a TxDefaultProfile.

gberardi-pillar:
These changes apply to K01.FR.52 and K01.FR.53.

Given a TxDefaultProfile on an EVSE or on the charging station as a whole, we can validate whether a new TxDefaultProfile is allowed.

### DIFF
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -56,10 +56,9 @@ public:
     ProfileValidationResultEnum validate_evse_exists(int32_t evse_id) const;
 
     ///
-    /// \brief validates the given \p profile according to the specification
+    /// \brief validates the given \p profile and associated \p evse_id according to the specification
     ///
-    ProfileValidationResultEnum validate_tx_default_profile(const ChargingProfile& profile,
-                                                            std::optional<EvseInterface*> evse_opt) const;
+    ProfileValidationResultEnum validate_tx_default_profile(const ChargingProfile& profile, int32_t evse_id) const;
 
     ///
     /// \brief validates the given \p profile according to the specification
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -31,11 +31,6 @@ enum class ProfileValidationResultEnum {
     DuplicateTxDefaultProfileFound
 };
 
-struct EvseProfile {
-    int32_t evse_id;
-    ChargingProfile profile;
-};
-
 /// \brief This class handles and maintains incoming ChargingProfiles and contains the logic
 /// to calculate the composite schedules
 class SmartChargingHandler {
@@ -44,7 +39,7 @@ private:
 
     std::shared_ptr<ocpp::v201::DatabaseHandler> database_handler;
     // cppcheck-suppress unusedStructMember
-    std::vector<EvseProfile> charging_profiles;
+    std::map<int32_t, ChargingProfile> charging_profiles;
     std::vector<ChargingProfile> station_wide_charging_profiles;
 
 public:
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -39,7 +39,7 @@ private:
 
     std::shared_ptr<ocpp::v201::DatabaseHandler> database_handler;
     // cppcheck-suppress unusedStructMember
-    std::map<int32_t, ChargingProfile> charging_profiles;
+    std::map<int32_t, std::vector<ChargingProfile>> charging_profiles;
     std::vector<ChargingProfile> station_wide_charging_profiles;
 
 public:
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -28,6 +28,12 @@ enum class ProfileValidationResultEnum {
     ChargingProfileExtraneousStartSchedule,
     ChargingSchedulePeriodsOutOfOrder,
     ChargingSchedulePeriodInvalidPhaseToUse,
+    DuplicateTxDefaultProfileFound
+};
+
+struct EvseProfile {
+    int32_t evse_id;
+    ChargingProfile profile;
 };
 
 /// \brief This class handles and maintains incoming ChargingProfiles and contains the logic
@@ -38,7 +44,8 @@ private:
 
     std::shared_ptr<ocpp::v201::DatabaseHandler> database_handler;
     // cppcheck-suppress unusedStructMember
-    std::vector<ChargingProfile> charging_profiles;
+    std::vector<EvseProfile> charging_profiles;
+    std::vector<ChargingProfile> station_wide_charging_profiles;
 
 public:
     explicit SmartChargingHandler(std::map<int32_t, std::unique_ptr<EvseInterface>>& evses);
@@ -51,13 +58,25 @@ public:
     ///
     /// \brief validates the given \p profile according to the specification
     ///
+    ProfileValidationResultEnum validate_tx_default_profile(const ChargingProfile& profile,
+                                                            std::optional<EvseInterface*> evse_opt) const;
+
+    ///
+    /// \brief validates the given \p profile according to the specification
+    ///
     ProfileValidationResultEnum validate_tx_profile(const ChargingProfile& profile, EvseInterface& evse) const;
 
     /// \brief validates that the given \p profile has valid charging schedules
     ProfileValidationResultEnum validate_profile_schedules(const ChargingProfile& profile) const;
 
+    ///
     /// \brief Adds a given \p profile to our stored list of profiles
-    void add_profile(const ChargingProfile& profile);
+    ///
+    void add_profile(int32_t evse_id, ChargingProfile& profile);
+
+private:
+    std::vector<ChargingProfile> get_evse_specific_tx_default_profiles() const;
+    std::vector<ChargingProfile> get_station_wide_tx_default_profiles() const;
 };
 
 } // namespace ocpp::v201
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -69,7 +69,7 @@ public:
     ProfileValidationResultEnum validate_profile_schedules(const ChargingProfile& profile) const;
 
     ///
-    /// \brief Adds a given \p profile to our stored list of profiles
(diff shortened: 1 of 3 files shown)


gberardi-pillar:
@shankari Please review these changes.

gberardi-pillar:
@shankari In case you missed my previous comment, can you please review this PR?

shankari (changes requested):
- I think that some additional documentation would be helpful
- there is a static code analysis error in here

include/ocpp/v201/evse.hpp
```diff
@@ -71,6 +71,14 @@ class Evse {
     /// \return
     uint32_t get_number_of_connectors();
 
+    /// \brief Returns true if evse_id is 0.
+    /// \return
+    bool is_station_wide() const;
```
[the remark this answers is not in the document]
gberardi-pillar: An EVSE ID > 0 is a specific EVSE, but 0 is meant to be a station-wide value to apply to all EVSEs for that station.

The spec does not explicitly say so in so many words, but there are a few places in which an evseid = 0 means it applies to all EVSEs: "For TxDefaultProfile an evseId=0 applies the profile to each individual evse. For ChargingStationMaxProfile and ChargingStationExternalConstraints an evseId=0 contains an overal limit for the whole Charging Station."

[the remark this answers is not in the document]
gberardi-pillar: I have updated the comment with the above quote from the spec.

Pietfried (changes requested):
It is not intended that an Evse instance is initialized with ID 0. Ids of Evses must start from 1 upwards as required by 7.1 EVSE Numbering in OCPP2.0.1 Part 1 - Architecture & Topology. This requires a redesign of the `validate_tx_default_profile` function signature, where I think the `Evse` argument could simply be replaced with int32_t `evse_id`.

The Evse class does there not require a member function `is_station_wide`. 

Additionally, a more general note: Im not sure if the member `std::vector<EvseProfile> charging_profiles` of the `std::vector<EvseProfile> charging_profiles;` is complex enough to represent more complex combinations of profiles. The OCPP1.6 implementation makes use of two different std::map per EVSE (in 1.6 Evse and Connector is equivilant):
https://github.com/EVerest/libocpp/blob/main/include/ocpp/v16/connector.hpp#L20

Pietfried (changes requested):
lgtm, except for some minor function doc updates and the argument for `validate_tx_default_profile`

include/ocpp/v201/smart_charging.hpp
```diff
@@ -48,6 +55,12 @@ class SmartChargingHandler {
     ///
     ProfileValidationResultEnum validate_evse_exists(int32_t evse_id) const;
 
+    ///
+    /// \brief validates the given \p profile according to the specification
+    ///
+    ProfileValidationResultEnum validate_tx_default_profile(const ChargingProfile& profile,
+                                                            std::optional<Evse*> evse_opt) const;
```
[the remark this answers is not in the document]
gberardi-pillar: We've been treating the existence of an Evse as a specific instance and the lack of one as indicating ID = 0. I did this work to follow the pattern we have established elsewhere. 

But I can use an ID and check if it is 0 instead.

include/ocpp/v201/smart_charging.hpp
```diff
@@ -56,8 +69,14 @@ class SmartChargingHandler {
     /// \brief validates that the given \p profile has valid charging schedules
     ProfileValidationResultEnum validate_profile_schedules(const ChargingProfile& profile) const;
 
+    ///
     /// \brief Adds a given \p profile to our stored list of profiles
-    void add_profile(const ChargingProfile& profile);
+    ///
+    void add_profile(int32_t evse_id, ChargingProfile& profile);
```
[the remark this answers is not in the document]
gberardi-pillar: Done.

Pietfried (approved):
:+1: 
This needs a rebase since 
https://github.com/EVerest/libocpp/pull/534 has been merged

include/ocpp/v201/smart_charging.hpp
```diff
@@ -38,7 +44,8 @@ class SmartChargingHandler {
 
     std::shared_ptr<ocpp::v201::DatabaseHandler> database_handler;
     // cppcheck-suppress unusedStructMember
-    std::vector<ChargingProfile> charging_profiles;
+    std::vector<EvseProfile> charging_profiles;
```
marcemmers, line 47: Would an `std::map` with the evse_id and a vector of charging profiles make sense here? That way we don't need the specific type for `EvseProfile` and we can search a bit quicker too.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -102,8 +121,34 @@ ProfileValidationResultEnum SmartChargingHandler::validate_profile_schedules(con
...
-    charging_profiles.push_back(profile);
+void SmartChargingHandler::add_profile(int32_t evse_id, ChargingProfile& profile) {
+    if (STATION_WIDE_ID == evse_id) {
+        station_wide_charging_profiles.push_back(profile);
+    } else {
+        charging_profiles.push_back(EvseProfile{evse_id = evse_id, profile = profile});
+    }
+}
+
+std::vector<ChargingProfile> SmartChargingHandler::get_evse_specific_tx_default_profiles() const {
```
marcemmers, line 132: If there are evse specific default profiles, wouldn't it make sense to filter on them as well? Now you return all profiles for all evse's. I would expect a test to expose this issue as well.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -102,8 +121,34 @@ ProfileValidationResultEnum SmartChargingHandler::validate_profile_schedules(con
...
+    std::vector<ChargingProfile> evse_specific_tx_default_profiles;
+    for (auto iter = charging_profiles.begin(); iter != charging_profiles.end(); ++iter) {
```
marcemmers, line 134: If it is possible I would suggest to use the range based for loop here and in the other for loops. More readable and less error prone.

marcemmers (changes requested):
Sorry I am a bit late to the party here but I do have some suggestions/questions.

lib/ocpp/v201/smart_charging.cpp
gberardi-pillar: I'm unclear on what issue you are referring to. Are you saying that it shouldn't return all profiles for all EVSEs?

Otherwise, while you might be right that it might make sense to eventually filter based on EVSE, it wasn't necessary for this code, and I did not want to write code for something I might anticipate a need for, increasing the size of the PR and distracting from the focus of it.

Unless I misunderstand what you mean?

marcemmers: It could very well be that I misunderstood the intended functionality here.

What I meant is that if you add a TxDefaultProfile for EVSE 1 and then try to add a TxDefaultProfile for EVSE 2 I would expect that to be possible. Interpreting the code here I expect you to get a `ProfileValidationResultEnum::DuplicateTxDefaultProfileFound` error instead.

gberardi-pillar: Ah, I see what you are saying now. You're right. I can add a test case for that case.

gberardi-pillar:
Put PR back into draft due to still waiting for @shankari to re-review.

lib/ocpp/v201/smart_charging.cpp
gberardi-pillar: Turned out that it was already going to pass.

marcemmers: I guess it now passes since you are adding another profile with a different `ChargingProfile.id`. This makes sense now I've read this part of the spec because we can never have multiple profiles with the same ID according to usecase K01.FR.05. I did not know that before.

lib/ocpp/v201/smart_charging.cpp
gberardi-pillar: Done.

marcemmers: Looks quite a lot cleaner :+1:

include/ocpp/v201/smart_charging.hpp
gberardi-pillar: Ah, yes, you're correct.

include/ocpp/v201/smart_charging.hpp
```diff
@@ -20,7 +20,13 @@ enum class ProfileValidationResultEnum {
     TxProfileEvseIdNotGreaterThanZero,
     TxProfileTransactionNotOnEvse,
     TxProfileEvseHasNoActiveTransaction,
-    TxProfileConflictingStackLevel
+    TxProfileConflictingStackLevel,
+    DuplicateTxDefaultProfileFound
+};
+
+struct EvseProfile {
+    int32_t evse_id;
```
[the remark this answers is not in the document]
gberardi-pillar: I am assuming you saw the response to this same question on the EVSE validation PR.

include/ocpp/v201/evse.hpp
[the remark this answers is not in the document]
shankari: Please also add this assumption to the OCPP 201 document in https://github.com/EVerest/libocpp/blob/main/doc/ocpp_201_status.md

include/ocpp/v201/smart_charging.hpp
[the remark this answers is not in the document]
shankari: Yes. I am unable to resolve my comments, though. Feel free to resolve this if you have write acces.

shankari (approved):
@gberardi-pillar the changes look fine to me. However, I would suggest that, in the future, you have more detailed commit messages than "Review fix". A short description of what part of the review is being addressed by the commit would be helpful, along with a summary of the changes. You could even add a link to the commit to make it easier for future maintainers to easily find the review without having to search through all the comments in the PR

marcemmers (approved):

gberardi-pillar:
@Pietfried @marcemmers Is this PR ready to be merged?

marcemmers:
> @Pietfried @marcemmers Is this PR ready to be merged?

Yes, go for it :+1: 
Might be good to do a rebase before merging. Don't expect any issues there.

include/ocpp/v201/evse.hpp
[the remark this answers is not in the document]
gberardi-pillar: This code doesn't exist in the PR anymore.

drmrd:
Contributes to #361

Issue #324, opened 2023-12-18
Use Case C07: Authorization using Contract Certificates


RobertDeLeeuw:

mennodegraaf:
Work started, progress can be seen in: https://github.com/EVerest/libocpp/compare/main...mg-c07-authorize-contract-certs

mennodegraaf:
During implementation, it appeared that also some changes to libevse-security are needed in order to reuse method `verify_certificate`: https://github.com/EVerest/libevse-security/compare/main...mg-refactor-verify-certificate

PR #527, opened 2024-03-13
Mg c07 authorize contract certs

mennodegraaf:
## Describe your changes
- Refactor `verify_certificate` so that it can also be used for ISO P&C chain validation
- Allow generation of OCSP request data
- Implement ISO P&C in line with OCPP C07 requirements

https://github.com/EVerest/libocpp/issues/324

### DIFF
--- a/include/ocpp/common/types.hpp
+++ b/include/ocpp/common/types.hpp
@@ -335,6 +335,31 @@ CaCertificateType string_to_ca_certificate_type(const std::string& s);
 /// \returns an output stream with the CaCertificateType written to
 std::ostream& operator<<(std::ostream& os, const CaCertificateType& ca_certificate_type);
 
+enum class CertificateValidationResult {
+    Accepted,
+    Expired,
+    InvalidSignature,
+    IssuerNotFound,
+    InvalidLeafSignature,
+    InvalidChain,
+    Unknown,
+};
+
+namespace conversions {
+/// \brief Converts the given InstallCertificateResult \p e to human readable string
+/// \returns a string representation of the InstallCertificateResult
+std::string certificate_validation_result_to_string(CertificateValidationResult e);
+
+/// \brief Converts the given std::string \p s to InstallCertificateResult
+/// \returns a InstallCertificateResult from a string representation
+CertificateValidationResult string_to_certificate_validation_result(const std::string& s);
+} // namespace conversions
+
+/// \brief Writes the string representation of the given InstallCertificateResult \p
+/// install_certificate_result to the given output stream \p os \returns an output stream with the
+/// InstallCertificateResult written to
+std::ostream& operator<<(std::ostream& os, const CertificateValidationResult& certificate_validation_result);
+
 enum class InstallCertificateResult {
     InvalidSignature,
     InvalidCertificateChain,
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -1714,7 +1714,7 @@ void ChargePoint::sign_certificate_req(const ocpp::CertificateSigningUseEnum& ce
     std::optional<std::string> organization;
 
     if (certificate_signing_use == ocpp::CertificateSigningUseEnum::ChargingStationCertificate) {
-        req.certificateType = CertificateSigningUseEnum::ChargingStationCertificate;
+        req.certificateType = ocpp::v201::CertificateSigningUseEnum::ChargingStationCertificate;
         common =
             this->device_model->get_optional_value<std::string>(ControllerComponentVariables::ChargeBoxSerialNumber);
         organization =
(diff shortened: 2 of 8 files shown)


include/ocpp/common/types.hpp
```diff
@@ -335,6 +335,31 @@ CaCertificateType string_to_ca_certificate_type(const std::string& s);
 /// \returns an output stream with the CaCertificateType written to
 std::ostream& operator<<(std::ostream& os, const CaCertificateType& ca_certificate_type);
 
+enum class CertificateValidationResult {
```
marcemmers, line 338: I see this done before so maybe not the right time, but I think we should prevent making a copy of an enum definition at all cost. This is going to be really error prone.

If we want it in scope I would suggest using some kind of `using` statement here.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -563,6 +589,113 @@ AuthorizeResponse ChargePoint::validate_token(const IdToken id_token, const std:
...
+            bool ContractValidationOffline =
+                this->device_model->get_optional_value<bool>(ControllerComponentVariables::ContractValidationOffline)
+                    .value_or(true);
+            bool LocalAuthorizeOffline =
+                this->device_model->get_optional_value<bool>(ControllerComponentVariables::LocalAuthorizeOffline)
+                    .value_or(true);
+
+            // C07.FR.01: When CS is online, it shall send an AuthorizeRequest
+            // C07.FR.02: The AuthorizeRequest shall at least contain the OCSP data
+            // TODO: local validation results are ignored if response is based only on OCSP data, is that acceptable?
```
marcemmers, line 623: I am not sure about this. One could even argue that we should always generate the OCSP data in libocpp instead of allowing this data coming from the outside but I don't know if that is possible. Are there sources that only give us the OCSP data?

The way it is now a caller to this function can give a cert chain with different data then the OCSP contains so then we might need to check. In that case not even having OCSP as an argument would be just as quick.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -563,6 +589,113 @@ AuthorizeResponse ChargePoint::validate_token(const IdToken id_token, const std:
...
+                        response.idTokenInfo.status = AuthorizationStatusEnum::Expired;
+                        response.certificateStatus = AuthorizeCertificateStatusEnum::CertificateExpired;
+                        break;
+                    default:
+                        response.idTokenInfo.status = AuthorizationStatusEnum::Invalid;
+                        break;
+                    }
+                } else {
+                    // C07.FR.07: CS shall not allow charging
+                    response.idTokenInfo.status = AuthorizationStatusEnum::NotAtThisTime;
```
marcemmers, line 679: I would suggest a different return type here. `NotAtThisTime` holds a special meaning an could be used for UI purposes later. In this case I would just go with `Unknown`.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -563,6 +589,113 @@ AuthorizeResponse ChargePoint::validate_token(const IdToken id_token, const std:
         return response;
     }
 
+    // C07: Authorization using contract certificates
+    if (id_token.type == IdTokenEnum::eMAID) {
+        // Temporary variable that is set to true to avoid immediate response to allow the local auth list
+        // or auth cache to be tried
+        bool tryLocalAuthListOrCache = false;
+        bool forwardToCsms = false;
```
marcemmers, line 597: Nit: rename to `forwardedToCsms`. I now expected you to still take an action in the if statement checking this.

marcemmers (commented):
I think there is still something to discuss for this one.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -544,6 +544,32 @@ bool ChargePoint::on_charging_state_changed(const uint32_t evse_id, ChargingStat
     return false;
 }
 
+std::vector<OCSPRequestData> ChargePoint::generate_ocsp_data(const CiString<5500>& certificate) {
+    std::vector<OCSPRequestData> ocsp_request_data_list;
+    const auto ocsp_data_list =
+        this->evse_security->get_ocsp_request_data(certificate.get(), ocpp::CaCertificateType::MO);
```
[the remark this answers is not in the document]
mennodegraaf: During P&C the OCSP data needs to be generated for a contract chain, therefore MO is the default. If there are other use cases, we could indeed add the CertificateType as an argument.

include/ocpp/common/types.hpp
mennodegraaf: I guess how I did it is similar to how it is typically done in libocpp. @Pietfried any suggestions?

lib/ocpp/v201/charge_point.cpp
mennodegraaf: I implemented it this way to avoid breaking existing functionality. AFAIK in EVerest, it is possible that local chain validation and generation of OCSP data is done already by the ISO15118 module.

lib/ocpp/v201/charge_point.cpp
mennodegraaf: I have set it to `NotAtThisTime` for this purpose as the CSMS is offline. If the user comes back some other time if the CSMS if online again, authorization might be ok. The user gets feedback that the failed authorization is because of some problem in the CS, not with his contract.

lib/ocpp/v201/charge_point.cpp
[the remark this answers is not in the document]
mennodegraaf: Just noticed that https://github.com/EVerest/libevse-security/pull/61 is another use case.

lib/ocpp/v201/charge_point.cpp
marcemmers: Fair enough but that would be for other modules to decide what to show to the used based on the online status. NotAtThisTime is a specific return type specified in OCPP and in my opinion should hold special meaning if the CSMS reports this.

In other cases when we are offline and not able to authorize we also respond with unknown.

mennodegraaf: ok, done

Pietfried:
Closed because replaced by: https://github.com/EVerest/libocpp/pull/536

PR #534, opened 2024-03-19
Make EVSE mockable and add mock

christopher-davis-afs:
## Describe your changes

This set of changes makes EVSE derived from an interface similar to what we did with ComponentStateManager,
and adds a mock to our set of mocks for v201.

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] I have made corresponding changes to the documentation
- [x] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

PR #537, opened 2024-03-20
Schedule Validation Follow-Ups

christopher-davis-afs:
## Describe your changes

Depends on #534

Handles FRs 44, 45, and 49 for EVSEs.

A question: Should we also be checking the ChargingStation's `SupplyPhases` for these FRs for profiles set to EVSE \#0? (In the code, this would be the case when a value for `evse` is not provided to `validate_profile_schedules()`)

### DIFF
--- a/include/ocpp/v201/evse.hpp
+++ b/include/ocpp/v201/evse.hpp
@@ -18,6 +18,12 @@
 namespace ocpp {
 namespace v201 {
 
+enum class CurrentPhaseType {
+    AC,
+    DC,
+    Unknown,
+};
+
 class EvseInterface {
 public:
     virtual ~EvseInterface();
@@ -118,6 +124,9 @@ public:
     /// \brief Restores the operative status of a connector within this EVSE to the persisted status and recomputes its
     /// effective status \param connector_id The ID of the connector
     virtual void restore_connector_operative_status(int32_t connector_id) = 0;
+
+    /// \brief Returns the phase type for the EVSE based on its SupplyPhases. It can be AC, DC, or Unknown.
+    virtual CurrentPhaseType get_current_phase_type() = 0;
 };
 
 /// \brief Represents an EVSE. An EVSE can contain multiple Connector objects, but can only supply energy to one of
@@ -199,6 +208,8 @@ public:
     void set_evse_operative_status(OperationalStatusEnum new_status, bool persist);
     void set_connector_operative_status(int32_t connector_id, OperationalStatusEnum new_status, bool persist);
     void restore_connector_operative_status(int32_t connector_id);
+
+    CurrentPhaseType get_current_phase_type();
 };
 
 } // namespace v201
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -72,8 +72,10 @@ public:
     ///
     ProfileValidationResultEnum validate_tx_profile(const ChargingProfile& profile, EvseInterface& evse) const;
 
-    /// \brief validates that the given \p profile has valid charging schedules
-    ProfileValidationResultEnum validate_profile_schedules(const ChargingProfile& profile,
+    /// \brief validates that the given \p profile has valid charging schedules.
+    /// If a profiles charging schedule period does not have a valid numberPhases,
+    /// we set it to the default value (3).
+    ProfileValidationResultEnum validate_profile_schedules(ChargingProfile& profile,
                                                            std::optional<EvseInterface*> evse_opt = std::nullopt) const;
 
     ///
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -28,6 +28,7 @@ enum class ProfileValidationResultEnum {
     ChargingProfileExtraneousStartSchedule,
     ChargingSchedulePeriodsOutOfOrder,
     ChargingSchedulePeriodInvalidPhaseToUse,
+    ChargingSchedulePeriodExtraneousPhaseValues,
     DuplicateTxDefaultProfileFound
 };
 
@@ -69,7 +70,8 @@ public:
     ProfileValidationResultEnum validate_tx_profile(const ChargingProfile& profile, EvseInterface& evse) const;
 
     /// \brief validates that the given \p profile has valid charging schedules
-    ProfileValidationResultEnum validate_profile_schedules(const ChargingProfile& profile) const;
+    ProfileValidationResultEnum validate_profile_schedules(const ChargingProfile& profile,
+                                                           std::optional<EvseInterface*> evse_opt = std::nullopt) const;
 
     ///
     /// \brief Adds a given \p profile and associated \p evse_id to our stored list of profiles
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -31,6 +31,14 @@ enum class ProfileValidationResultEnum {
     DuplicateTxDefaultProfileFound
 };
 
+namespace conversions {
+/// \brief Converts the given ProfileValidationResultEnum \p e to human readable string
+/// \returns a string representation of the ProfileValidationResultEnum
+std::string profile_validation_result_to_string(ProfileValidationResultEnum e);
+} // namespace conversions
+
+std::ostream& operator<<(std::ostream& os, const ProfileValidationResultEnum validation_result);
+
 /// \brief This class handles and maintains incoming ChargingProfiles and contains the logic
 /// to calculate the composite schedules
 class SmartChargingHandler {
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -14,6 +14,8 @@
 
 namespace ocpp::v201 {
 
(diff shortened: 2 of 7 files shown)


christopher-davis-afs:
CC: @shankari

shankari:
@christopher-davis-afs there is a merge conflict here. I am happy to review regardless, but you will need to resolve the merge conflict before sending this to community review.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -50,36 +51,58 @@ ProfileValidationResultEnum SmartChargingHandler::validate_tx_profile(const Char
...
                     return ProfileValidationResultEnum::ChargingSchedulePeriodsOutOfOrder;
-                } else {
-                    charging_schedule_period = next_charging_schedule_period;
+                }
+            }
+
+            if (evse_opt.has_value()) {
+                auto evse = evse_opt.value();
+                // K01.FR.44 for EVSEs; We reject profiles that provide invalid numberPhases/phaseToUse instead
+                // of silently acccepting them.
```
shankari, line 88: we are now getting into the territory of "may" (e.g. for K01.FR.44):

> The Charging Station MAY respond with status = Accepted, instead of Rejected and ignore the provided values for numberPhases and phaseToUse.

Have we discussed this decision (reject instead of accept + ignore) with the community? Can you provide a link to the issue where it was discussed?

shankari (changes requested):
I haven't

include/ocpp/v201/charge_point.hpp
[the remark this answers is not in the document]
christopher-davis-afs: That is intentional - we depend on that PR, and these changes will be handled properly once that is merged.

lib/ocpp/v201/smart_charging.cpp
christopher-davis-afs: This was discussed on the Cloud Communication WG meeting last week, and the community agreed that for "MAY" cases that allow invalid state like this, we would reject the invalid state instead of silently correcting it.

The Everest community has also agreed to include a status document listing the state of the various functional requirements and relevant remarks. [This PR](https://github.com/EVerest/libocpp/pull/540/files) adds the document, and we plan to add remarks for functional requirements like this.

shankari: That sounds like a great plan. I don't see these test cases updated in that document though, I think it would be best to include the updates in this PR (as indicated in the modified template in https://github.com/EVerest/libocpp/pull/540

christopher-davis-afs: Updated.

shankari (approved):

christopher-davis-afs:
> @christopher-davis-afs This PR: #553 adds support for requesting the `SupplyPhases` variable for different EVSEs. For EVSE#0 it sounds logical to me to use the SupplyPhases variable that belongs to the component `ChargingStation`

I have another PR planned after this one to cover the rest of the FRs that are unimplemented. If it's okay, I plan to handle the EVSE#0 case there. Otherwise, I think this PR should be good to go.

include/ocpp/v201/smart_charging.hpp
```diff
@@ -28,9 +30,16 @@ enum class ProfileValidationResultEnum {
     ChargingProfileExtraneousStartSchedule,
     ChargingSchedulePeriodsOutOfOrder,
     ChargingSchedulePeriodInvalidPhaseToUse,
+    ChargingSchedulePeriodUnsupportedNumberPhases,
+    ChargingSchedulePeriodExtraneousPhaseValues,
     DuplicateTxDefaultProfileFound
 };
 
+inline std::ostream& operator<<(std::ostream& os, const ProfileValidationResultEnum validation_result) {
```
marcemmers, line 38: I would suggest to keep the implementation in the cpp file for this. Also it would be nice if we actually print the result in a human readable form. Take a look at the `v201/enums.hpp` for how it is done there (that is autogenerated btw).

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -77,36 +77,58 @@ ProfileValidationResultEnum SmartChargingHandler::validate_tx_profile(const Char
...
  */
-ProfileValidationResultEnum SmartChargingHandler::validate_profile_schedules(const ChargingProfile& profile) const {
-    auto schedules = profile.chargingSchedule;
+ProfileValidationResultEnum
+SmartChargingHandler::validate_profile_schedules(ChargingProfile& profile,
+                                                 std::optional<EvseInterface*> evse_opt) const {
+    auto schedules = &profile.chargingSchedule;
 
-    for (auto schedule : schedules) {
+    for (ChargingSchedule& schedule : *schedules) {
```
marcemmers, line 87: Why not do this in one step?
`for (ChargingSchedule& schedule : profile.chargingSchedule) {`

If you do want to do this (for readability for example which i can understand). Please don't make a pointer and then de-reference that if its not needed.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -77,36 +77,58 @@ ProfileValidationResultEnum SmartChargingHandler::validate_tx_profile(const Char
...
         // A schedule must have at least one chargingSchedulePeriod
         if (schedule.chargingSchedulePeriod.empty()) {
             return ProfileValidationResultEnum::ChargingProfileNoChargingSchedulePeriods;
         }
 
-        auto charging_schedule_period = schedule.chargingSchedulePeriod[0];
-
         for (auto i = 0; i < schedule.chargingSchedulePeriod.size(); i++) {
+            auto charging_schedule_period = &schedule.chargingSchedulePeriod[i];
```
marcemmers, line 94:
```suggestion
            const auto& charging_schedule_period = schedule.chargingSchedulePeriod[i];
```

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -77,36 +77,58 @@ ProfileValidationResultEnum SmartChargingHandler::validate_tx_profile(const Char
...
 
             // K01.FR.31
-            if (i == 0 && charging_schedule_period.startPeriod != 0) {
+            if (i == 0 && charging_schedule_period->startPeriod != 0) {
                 return ProfileValidationResultEnum::ChargingProfileFirstStartScheduleIsNotZero;
                 // K01.FR.35
-            } else if (i != 0) {
-                auto next_charging_schedule_period = schedule.chargingSchedulePeriod[i];
-                if (next_charging_schedule_period.startPeriod <= charging_schedule_period.startPeriod) {
+            } else if (i != 0 && i + 1 < schedule.chargingSchedulePeriod.size()) {
```
marcemmers, line 104: This does not need to be a `else if` since the first if already returns.

marcemmers, line 104: Shouldn't we check this for `i == 0` too since you check the next period against the current

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -77,36 +77,58 @@ ProfileValidationResultEnum SmartChargingHandler::validate_tx_profile(const Char
  * - K01.FR.20
  * - K01.FR.34
  * - K01.FR.43
- * - K01.FR.45
  * - K01.FR.48
  */
-ProfileValidationResultEnum SmartChargingHandler::validate_profile_schedules(const ChargingProfile& profile) const {
-    auto schedules = profile.chargingSchedule;
+ProfileValidationResultEnum
+SmartChargingHandler::validate_profile_schedules(ChargingProfile& profile,
```
marcemmers, line 83: At first I was thinking this could have `const` arguments but it also fixes some stuff like the default to 3 phases.

I would argue we should either change the name of the method to reflect that or split the validation and applying assumptions in different methods.

include/ocpp/v201/evse.hpp
```diff
@@ -18,6 +18,12 @@
 namespace ocpp {
 namespace v201 {
 
+enum CurrentPhaseType {
```
marcemmers, line 21:
```suggestion
enum class CurrentPhaseType {
```

marcemmers (changes requested):

lib/ocpp/v201/smart_charging.cpp
christopher-davis-afs: This can't be const since we modify its `numberPhases` for FR 49.

lib/ocpp/v201/smart_charging.cpp
christopher-davis-afs: If we change the name of the method, what would you suggest?

marcemmers: I actually don't know without it becoming a mess of a method name. I am generally not in favor of having validations and getter functions modify stuff but we do it more in this library so lets leave it be.

As a compromise you could maybe update the doxygen for the method to reflect that it is actually setting stuff to default if left empty.

lib/ocpp/v201/smart_charging.cpp
marcemmers: Yep you are right, I did not see that yet when making this comment. The suggestion to change it from pointer type to reference still stands though.

christopher-davis-afs:
Is this good to merge?

marcemmers (approved):

drmrd:
Contributes to #361

PR #540, opened 2024-03-26
Adding OCPP 2.0.1 status document

RobertDeLeeuw:
- [X] I have made corresponding changes to the documentation
- [X] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,5 +5,6 @@
 ## Checklist before requesting a review
 - [ ] I have performed a self-review of my code
 - [ ] I have made corresponding changes to the documentation
+- [ ] If OCPP 2.0.1: I have updated the [OCPP 2.0.1 status document](https://github.com/EVerest/libocpp/tree/main/doc/ocpp_201_status.md)
 - [ ] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements
 
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ All documentation and the issue tracking can be found in our main repository her
 ## Support for OCPP 2.0.1
 
 The development of OCPP2.0.1 is in progress. 
-[For details, see the sheet with the current implementation status.](https://docs.google.com/spreadsheets/d/1NXEa7d0pkT4_1L9lZ9RpiLHGPtTKoqwPsKLJPjTZ_vM/edit?usp=sharing)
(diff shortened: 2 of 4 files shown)


marcemmers:
Yesterday we suggested to have some kind of legend of what statuses we have and then not have text anymore there.

Proposal for the list of statuses:
Done :white_check_mark:
In progress :fast_forward: 
Not applicable :no_entry_sign:

Do we need more options?

marcemmers:
Leaving it empty seems logical to me. Keep in mind that the proposed legend is my proposal, we didn't discuss these icons specifically. Only that we wanted some kind of legend.

Pietfried:
I'd drop `In progress`. It removes one level of complexity and is propably harder to keep up to date. We have the issue tracker for this.

I have updated all functional requirements where there was not yet an entry. I assumed that all fullfilled requirements are still fullfilled.

What I noticed is that `n/a` can currently mean the following:
* It is a requirement for the CSMS
* It cannot / is not handled by libocpp but by the libocpp user application / an external library (like libevse-security)

We could differentiate between these these two cases, but I'm ok with using `n/a` for both and use the remark column if it could be unclear.

RobertDeLeeuw:
We could also remove all the CSMS requirements. And make clear we only kept the list of Charging Station relevant requirements

marcemmers:
I agree with @Pietfried that dropping in progress might be a good idea. Even if it is for the fact that someone who starts to work on something needs to open a PR to put it in progress which is not something that will be done often (or ever).

Personally I would not drop the CSMS requirements just to keep it explicit that we don't have to do them. Otherwise you could always get questions about certain requirements if they are applicable. We could of course find another emoji for N/A because of CSMS to differentiate if we think that is needed.

RobertDeLeeuw:
Added description to begin of the page and remove "In Progress" emoji.

christopher-davis-afs (changes requested):
Left comments on what's been done in K01 and what isn't applicable.

Pietfried:
> Left comments on what's been done in K01 and what isn't applicable.

I pushed an update based on your comments. Please check back, feel free to also push updates on this branch and let us know when the update on the smart charging block is complete from your side

christopher-davis-afs:
> > Left comments on what's been done in K01 and what isn't applicable.
> 
> I pushed an update based on your comments. Please check back, feel free to also push updates on this branch and let us know when the update on the smart charging block is complete from your side

I think this covers the progress on SmartCharging 👍

Pietfried (approved):

PR #548, opened 2024-04-02
Add unit tests for sha256 functions

hikinggrass:
## Describe your changes
Unit tests for the sha256 functions as a prerequisite for https://github.com/EVerest/libocpp/pull/535

### DIFF
--- a/tests/v201_utils_tests.cpp
+++ b/tests/v201_utils_tests.cpp
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2024 Pionix GmbH and Contributors to EVerest
+
+#include <gtest/gtest.h>
+#include <ocpp/v201/utils.hpp>
+
+namespace ocpp {
+namespace common {
+
+class V201UtilsTest : public ::testing::Test {
+protected:
+    const std::string empty_input;
+    const std::string short_input = "hello there";
+    const std::string long_input =
+        "hello there hello there hello there hello there hello there hello there hello there hello there "
+        "hello there hello there hello there hello there hello there hello there hello there hello there "
+        "hello there hello there hello there hello there hello there hello there hello there hello there "
+        "hello there hello there hello there hello there hello there hello there hello there hello there";
+    const std::string invalid = "354cacb2d2c45cb28af92ca348ea3a2236ecc48c81c78e0924bf46bd68d9c407";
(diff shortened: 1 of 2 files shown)


tests/v201_utils_tests.cpp
```diff
@@ -0,0 +1,58 @@
...
+    ASSERT_EQ("e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+              ocpp::v201::utils::sha256(empty_input));
+    ASSERT_EQ("12998c017066eb0d2a70b94e6ed3192985855ce390f321bbdb832022888bd251",
+              ocpp::v201::utils::sha256(short_input));
+    ASSERT_EQ("34aa8868354dc8eb0e76fbc8b5f13259094bcdf5688c6a48a2bcb89ed863d441",
+              ocpp::v201::utils::sha256(long_input));
+}
+
+TEST_F(V201UtilsTest, test_invalid_sha256) {
+    ASSERT_NE(invalid, ocpp::v201::utils::sha256(empty_input));
```
marcemmers, line 38: I am not so sure what the purpose is of these `NE` tests, especially if you already tested the valid outputs.

tests/v201_utils_tests.cpp
```diff
@@ -0,0 +1,58 @@
...
+    ASSERT_NE(invalid, ocpp::v201::utils::sha256(short_input));
+    ASSERT_NE(invalid, ocpp::v201::utils::sha256(long_input));
+}
+
+TEST_F(V201UtilsTest, test_valid_generate_token_hash) {
+    ocpp::v201::IdToken valid_iso14443_token = {"ABAD1DEA", ocpp::v201::IdTokenEnum::ISO14443};
```
marcemmers, line 44: Might be interesting to validate that the same token with a different token type results in a different output?

marcemmers (changes requested):

tests/v201_utils_tests.cpp
hikinggrass: You're right, I've removed them

tests/v201_utils_tests.cpp
marcemmers: But now you still have one token with id `valid` and the other `ABAD1DEA` right?

hikinggrass: You're right, I added another test that checks for that :+1:

marcemmers (approved):

Issue #495, opened 2024-02-26
StopTransaction contains transactionId -1 in OCPP 1.6 start offline case


lategoodbye:
### OCPP Version

OCPP1.6

### Describe the bug

We noticed the following issue with our Tarragon platform and EVerest core 2024.01. In case we start an offline transaction and stop it online the stopTransaction contains transactionId -1. But the expected behavior would be that EVerest replace the bogus transactionId with the real value from the previous StartTransaction.Conf. Otherwise the transaction will be running endlessly.

OCPP communication with Steve as backend (shouldn't matter)
```
[INFO ] 2024-02-26 09:14:46,425 de.rwth.idsg.steve.ocpp.ws.WebSocketLogger - [chargeBoxId=lesbos_left, sessionId=5863aadf-4995-59e9-1bda-f2c9e09c9e78] Received: [2,"2e55de9d-3fea-4782-9b57-2bc027976a31","StatusNotification",{"connectorId":1,"errorCode":"NoError","info":"EVConnected","status":"Preparing","timestamp":"2024-02-26T09:12:49.706Z"}]
[INFO ] 2024-02-26 09:14:46,434 de.rwth.idsg.steve.ocpp.ws.WebSocketLogger - [chargeBoxId=lesbos_left, sessionId=5863aadf-4995-59e9-1bda-f2c9e09c9e78] Sending: [3,"2e55de9d-3fea-4782-9b57-2bc027976a31",{}]
[INFO ] 2024-02-26 09:14:46,435 de.rwth.idsg.steve.ocpp.ws.WebSocketLogger - [chargeBoxId=lesbos_left, sessionId=5863aadf-4995-59e9-1bda-f2c9e09c9e78] Received: [2,"941dfb0b-b995-45f5-9833-9dfb3f60dcfb","StartTransaction",{"connectorId":1,"idTag":"DEADBEEF","meterStart":971,"timestamp":"2024-02-26T09:12:50.059Z"}]
[INFO ] 2024-02-26 09:14:46,841 de.rwth.idsg.steve.ocpp.ws.WebSocketLogger - [chargeBoxId=lesbos_left, sessionId=5863aadf-4995-59e9-1bda-f2c9e09c9e78] Sending: [3,"941dfb0b-b995-45f5-9833-9dfb3f60dcfb",{"transactionId":1458,"idTagInfo":{"status":"Accepted","expiryDate":"2024-02-26T10:14:46.831Z"}}]
[...]

PR #551, opened 2024-04-04
OCPP1.6: Improved offline transaction handling

Pietfried:
## Describe your changes
Improved offline transaction handling:
* extended TRANSACTION table in database with START_TRANSACTION_MESSAGE_ID and STOP_TRANSACTION_MESSAGE_ID
* these columns are required in order to process transactionIds received from the CSMS after being offline
* Generating negative transaction ids for transactions to be able to differentiate between multiple offline transactions
* Improved stopping of pending transactions on startup

https://github.com/EVerest/libocpp/issues/495


lategoodbye:
@Pietfried Thanks for the update. Currently i cannot promise that i have the chance to test the fix before 2024.03 is released.

But i like to give some feedback to the implementation. In our propietary stack we also had similiar issues with offline mode. In our transaction database of the OCPP client we decided to store two separate transaction ids (internal & external). At the start of the transaction (regardless on on/offline case) the internal one is generated and stored in the database (external is left empty / NULL). In online case then the StartTransaction.Req is send and in case the StartTransaction.Conf has been received this transaction id is assigned as the external one of the ongoing internal transaction in the database.

This approach has the advantage that the implementation doesn't rely on the assumption that the OCPP CSMS never use negative transaction ids. From our experience at least Virta CSMS use a transaction id = -1in case the start transaction has been rejected (ConcurrentTx).

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -781,12 +790,14 @@ void ChargePointImpl::send_meter_value(int32_t connector, MeterValue meter_value
     req.connectorId = connector;
     std::ostringstream oss;
     oss << "Gathering measurands of connector: " << connector;
+
     if (connector > 0) {
         auto transaction = this->transaction_handler->get_transaction(connector);
-        if (transaction != nullptr && transaction->get_transaction_id() != -1) {
-            auto transaction_id = transaction->get_transaction_id();
+        if (transaction != nullptr and transaction->get_transaction_id().has_value()) {
+            // transactions with a positive transaction id have been acknowledged by CSMS
```
hikinggrass, line 797: Does this comment still apply?

Pietfried, line 797:
```suggestion
```

Pietfried: Nope :)

hikinggrass (approved):
Since the database schema changed we should probably add a migration for the next release (or at least warn about this change)

Issue #542, opened 2024-03-27
Libwebsockets: be able to set an interface or ip address to connect to.


maaikez:
### OCPP Version

OCPP1.6, OCPP2.0.1

### Describe the problem

For the network connection profiles refactoring (issue #408), we want to be able to set an interface the websocket has to use to connect to. It should be able to set an interface or the ip address of the interface. It is possible an interface has two ip addresses (one link local and one 'normal' ip address) so it must be possible to set the ip address the websocket needs to connect to. Also, if you do nothing, the websocket library might just choose to use for example the wired interface to connect while you want to connect via the modem. 

So the feature request is to make sure it is possible to set an ip address of in interface which the websocket will use to connect to the backoffice.

### Describe your solution

In libwebsockets, in the `lws_context_creation_info` struct, you can set the `iface`, which can be an ip address or the interface name. It is available when `LWS_WITH_NETWORK` is defined. However, if you only set this `iface`, it seems the websocket still connects via another interface (tested with a network connection profile set to a modem with a local ip address that is not available via the outside world, so it should not be able to connect in that case). According to the documentation, we should set `bind_iface` in the same struct to `1` to make this work (this should call `setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, vhost->iface, (socklen_t)strlen(vhost->iface)`, found in `unix-sockets.c`). However, if I try that, the websocket can not connect with the backend at all anymore. I did not find out why yet.

maaikez:
It looks like setsockopt (SO_BINDTODEVICE) does not work with an ip address, so I think we should not set that either. It is unclear to me why setting the ip address in `iface` does not work, because I think it uses the 'bind' call underwater, which should just work.

AssemblyJohn:
I see there are two 'iface' definitions, one in 'lws_client_connect_info' and one in 'lws_context_creation_info'. Will start looking into the issue with both IP binding and interface name binding.

Note: as per some posts, the 'SO_BINDTODEVICE' requires elevated permissions and must be supported by the OS.
Note2: CAP_NET_RAW capability is also required

PR #552, opened 2024-04-05
Added support for iface in libwebsockets, added config values, added …

AssemblyJohn:
## Describe your changes

Added the possibility of setting the used network iface for libwebsocket. Updated documentation for usage.

[54](https://github.com/EVerest/libocpp/issues/542)

### DIFF
--- a/README.md
+++ b/README.md
@@ -558,3 +558,13 @@ The old websocket++ implementation has been deprecated. For enabling websocket++
 ```bash
   cmake .. -DLIBOCPP_ENABLE_DEPRECATED_WEBSOCKETPP=ON
 ```
+
+### Support for iface
+
+In order to connect through a custom network iface, a custom internal config variable 'IFace' can be used.
+
+```json
+"Internal": {        
+    "IFace" : "enp43s0"
+}
+```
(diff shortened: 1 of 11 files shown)


marcemmers:
> It has been concluded that in the current form it serves the purpose. However, the config value will be used only as the 'default' for a network profile, if at runtime the connection manager executes an config parameter over-write, that will be the used network interface.

I would suggest to change the naming so this is also clear. Maybe something like DefaultIFace?

AssemblyJohn:
> > It has been concluded that in the current form it serves the purpose. However, the config value will be used only as the 'default' for a network profile, if at runtime the connection manager executes an config parameter over-write, that will be the used network interface.
> 
> I would suggest to change the naming so this is also clear. Maybe something like DefaultIFace?

I am wondering if that would not create further confusion, because there is a system-wide default IFace that is used while creating any socket. Plus, it is an optional parameter, it is not required to be present in the DB, the 'rea' default is 'system-default'.

marcemmers:
I see what you mean that that would be confusing too. Still, once we have the option to change the interface via the new functionality it is also confusing. I don't know how to improve this though so open to suggestions. Otherwise I guess we can leave it like you have it now.

AssemblyJohn:
I'll come back after I can have a look at the network profiles PR and see how it was integrated there.

marcemmers (approved):

Pietfried (approved):

PR #555, opened 2024-04-08
Made libwebsocket usage default

AssemblyJohn:
## Describe your changes

Websocket++ usage has been deprecated and due to future heavy interface refactoring it has been removed from the default compilation. Libwebsocket has been made the default websocket library that is used.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -56,7 +56,7 @@ add_subdirectory(ocpp/v16/messages)
 add_subdirectory(ocpp/v201/messages)
 
 option(LIBOCPP_USE_BOOST_FILESYSTEM "Usage of boost/filesystem.hpp instead of std::filesystem" OFF)
-option(LIBOCPP_ENABLE_LIBWEBSOCKETS "Usage of libwebsockets instead of websockets++" OFF)
+option(LIBOCPP_ENABLE_DEPRECATED_WEBSOCKETPP "Usage of deprecated websocket++ instad of libwebsockets" OFF)
 
 target_include_directories(ocpp
     PUBLIC
@@ -110,10 +110,10 @@ endif()
  
 target_link_libraries(ocpp
     PUBLIC
-        everest::timer
-        websocketpp::websocketpp
+        everest::timer        
         nlohmann_json_schema_validator
         everest::evse_security
+        websocketpp::websocketpp
     PRIVATE
         OpenSSL::SSL
         OpenSSL::Crypto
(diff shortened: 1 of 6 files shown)


lib/ocpp/common/websocket/websocket.cpp
```diff
@@ -5,8 +5,10 @@
 #include <ocpp/common/websocket/websocket.hpp>
 #include <ocpp/v16/types.hpp>
 
-#ifdef LIBOCPP_ENABLE_LIBWEBSOCKETS
 #include <ocpp/common/websocket/websocket_libwebsockets.hpp>
+
+#ifdef LIBOCPP_ENABLE_DEPRECATED_WEBSOCKETPP
+#include <ocpp/common
```
Pietfried, line 11: This doesnt look right

Pietfried (dismissed):

lib/CMakeLists.txt
```diff
@@ -110,10 +110,10 @@ endif()
  
 target_link_libraries(ocpp
     PUBLIC
-        everest::timer
-        websocketpp::websocketpp
+        everest::timer        
         nlohmann_json_schema_validator
         everest::evse_security
+        websocketpp::websocketpp
```
hikinggrass, line 116: That should probably be conditional on LIBOCPP_ENABLE_DEPRECATED_WEBSOCKETPP

hikinggrass: There's still a find_package call in the toplevel CMakeLists.txt as well: https://github.com/EVerest/libocpp/blob/8a9e06962e4beff69eb954f94b0b28c64bfa8314/CMakeLists.txt#L43
That should be conditional too (and have one for libwebsockets instead)

AssemblyJohn: I have done just that, however we have the following problem: Since at the moment the interface refactor is NOT done (in the current state) ALL the code still depends on websocket++ headers. So, I've tried to make it conditional but the compilation failed when the websocket++ library was not added.

The workaround is to leave it inside, until we have the interface refactor that will not make the code dependant on websocket++.

lib/ocpp/common/websocket/websocket.cpp
```diff
@@ -19,14 +17,14 @@ Websocket::Websocket(const WebsocketConnectionOptions& connection_options, std::
...
-    this->websocket = std::make_unique<WebsocketTlsTPM>(connection_options, evse_security);
-#else
+#ifdef LIBOCPP_ENABLE_DEPRECATED_WEBSOCKETPP
     if (connection_options.security_profile <= 1) {
         this->websocket = std::make_unique<WebsocketPlain>(connection_options);
     } else if (connection_options.security_profile >= 2) {
         this->websocket = std::make_unique<WebsocketTLS>(connection_options, evse_security);
     }
+#else
+    this->websocket = std::make_unique<WebsocketTlsTPM>(connection_options, evse_security);
```
Pietfried, line 27: We should rename this class in another PR at some point..

marcemmers (approved):

Issue #450, opened 2024-02-05
Upgrading to a newer libocpp version currently requires a database wipe. There should be upgrade paths for this.


marcemmers:
### Problem:
When we make changes to the SQLite init files we use we can end up breaking compatibility with databases already out there. For example if a table gets an extra column there is no way to detect that purely from the init file. You could do an alter table but then that would fail if the new column already exists leading to complicated init files as well.

Currently the only way to fix these issues is to delete your local database and create a new clean one with the correct tables set.

There should be a way to upgrade to a newer libocpp version without having to delete databases since that just won't do in a production environment.

### Requirements:
These are things I think we should want as requirements for the new functionality:
* From a certain version (1.0.0?) onward there should always be an upgrade path all the more recent versions.
* We must be able to go back to the database we had before trying to apply changes in case an error occurs. This might be done with a simple backup before applying updates.

***Nice to haves:***
* Rollback functionality. You want to be able to go back to an older version from a more recent version.

### Potential ways to implement
* One option would be to use the `user_version` to keep track of which changes we still need to make and then add a list of those changes in code. Kind of like https://github.com/cljoly/rusqlite_migration/tree/master although that is a rust implementation. The principle seems rather simple and easy to implement.
* Leave the SQLite init files as is and use an automatic detection to do the migrations like described here: https://david.rothlis.net/declarative-schema-migration-for-sqlite/
* https://contribute.liquibase.com/extensions-integrations/directory/database-tutorials/sqlite/
  * Needs Java to run so not an option for this project
* https://sqitch.org/
  * Trying to install on my ubuntu machine took up roughly 50mb for all the dependencies. We might be able to strip it down but seems a bit overkill.
* https://atlasgo.io/getting-started
  * Looks to be catered towards cloud bases solutions and applying migrations to databases using CI kind of practices. Not an options for us in my opinion.
* ...
* There might be more options and libraries. Feel free to suggest, I'll try to add some later as well.

marcemmers:
Looking at all the options we have and the discussion we had during the working group meeting of the 7th of February I would suggest to go for a rather simple solution quite similar to the [rusqlite_migration](https://github.com/cljoly/rusqlite_migration/tree/master) library without using separate packages that need to be installed.

#### More in detail:
* We start out with an SQL init file that is used to start the database with. This file should not be changed anymore once this functionality is implemented.
* Every change we want to make to the database is done through migration files. These files are simply more SQL files that can be applied to the database in the right order.
[...]

PR #556, opened 2024-04-08
Add functionality for database migrations

marcemmers:
## Describe your changes

Implemented database migrations according to this comment in the original issue: https://github.com/EVerest/libocpp/issues/450#issuecomment-1933807346

The method is also implemented for 1.6 and 2.0.1 directly in this PR.

There are a few differences and things to consider:
* Making a backup of the database before applying the migrations is up to the user of the library and is not implemented.
  * Since the migrations are applied on construction of the `charge_point`, the backup needs to be made before constructing it.
* The user of the library is responsible for getting the migration files on the target and also making sure that even on a downgrade the files stay on target. This is because the down files from the "newer" version are used to go back to the old database version.
* We are using the SQLite command `ALTER TABLE DROP COLUMN` which is only supported from SQLite version 3.35 onwards. This is available from Yocto version 3.3 hardknott and Ubuntu 22.04 for example.

https://github.com/EVerest/libocpp/issues/450


lib/ocpp/common/database/database_connection.cpp
```diff
@@ -28,14 +33,22 @@ bool DatabaseConnection::open_connection() {
...
 
 bool DatabaseConnection::close_connection() {
-    EVLOG_info << "Closing database connection...";
+    return this->close_connection_internal(false);
+}
+
+bool DatabaseConnection::close_connection_internal(bool force_close) {
+    if (!force_close && this->open_count.fetch_sub(1) != 1) {
+        EVLOG_debug << "Connection should remain open for other users";
+        return true;
```
hikinggrass, line 47: Shouldn't this return false since the closing did not happen?

marcemmers: My idea behind true was that the operation did not fail per se. If I would return false then the user of the function might think it failed.

config/CollectMigrationFiles.cmake
```diff
@@ -0,0 +1,63 @@
+function(COLLECT_MIGRATION_FILES)
```
hikinggrass, line 1: this function name should be lowercase

Pietfried (approved):

hikinggrass (approved):

PR #568, opened 2024-04-15
Removed dependency on websocket++, minimized dependencies

AssemblyJohn:
## Describe your changes

Removed the strong dependency to websocket++ by adding the relevant headers to a 3rd_party folder, and making the library optional based on a cmake option. The hard-coded websocket++ constants were also removed.

Linked core PR: https://github.com/EVerest/everest-core/pull/631

### DIFF
--- a/include/ocpp/common/types.hpp
+++ b/include/ocpp/common/types.hpp
@@ -636,6 +636,41 @@ enum class MessageDirection {
     ChargingStationToCSMS
 };
 
+enum class ConnectionFailedReason {
+    InvalidCSMSCertificate = 0,
+};
+
+///
+/// \brief Reason why a websocket closes its connection
+///
+enum class WebsocketCloseReason : uint8_t {
+    /// Normal closure, meaning that the purpose for which the connection was
+    /// established has been fulfilled.
+    Normal = 1,
+    /// Close the connection with a forced TCP drop.
+    /**
+     * This special value requests that the WebSocket connection be closed by
+     * forcibly dropping the TCP connection. This will leave the other side of
+     * the connection with a broken connection and some expensive timeouts. this
+     * should not be done except in extreme cases or in cases of malicious
+     * remote endpoints.
+     */
+    ForceTcpDrop,
+    /// The endpoint was "going away", such as a server going down or a browser
+    /// navigating away from a page.
+    GoingAway,
+    /// A dummy value to indicate that the connection was closed abnormally.
+    /**
+     * In such a case there was no close frame to extract a value from. This
+     * value is illegal on the wire.
+     */
+    AbnormalClose,
+    /// Indicates that the service is restarted. A client may reconnect and if
+    /// if it chooses to do so, should reconnect using a randomized delay of
+    /// 5-30s
+    ServiceRestart
+};
+
 } // namespace ocpp
 
 #endif
--- a/include/ocpp/common/types.hpp
+++ b/include/ocpp/common/types.hpp
@@ -644,30 +644,11 @@ enum class ConnectionFailedReason {
 /// \brief Reason why a websocket closes its connection
 ///
 enum class WebsocketCloseReason : uint8_t {
-    /// Normal closure, meaning that the purpose for which the connection was
-    /// established has been fulfilled.
+    /// Normal closure
     Normal = 1,
-    /// Close the connection with a forced TCP drop.
-    /**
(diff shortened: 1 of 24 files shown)


lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -715,7 +716,7 @@ void WebsocketTlsTPM::close(websocketpp::close::status::value code, const std::s
     // Notify any message senders that are waiting, since we can't send messages any more
     msg_send_cv.notify_all();
 
-    std::thread closing([this]() { this->closed_callback(websocketpp::close::status::normal); });
+    std::thread closing([this]() { this->closed_callback(WebsocketCloseReason::Normal); });
```
Pietfried, line 719: Shouldnt this use the `code` argument?

include/ocpp/common/websocket/websocket.hpp
```diff
@@ -34,7 +37,7 @@ class Websocket {
     void set_connection_options(const WebsocketConnectionOptions& connection_options);
 
     /// \brief disconnect the websocket
-    void disconnect(websocketpp::close::status::value code);
+    void disconnect(WebsocketCloseReason code);
 
     // \brief reconnects the websocket after the delay
     void reconnect(std::error_code reason, long delay);
```
Pietfried, line 43: Looks like the `std::error_code reason` is not used and can also be removed

Pietfried (changes requested):
Looks good. Added a few remarks / suggestions.

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
AssemblyJohn: This works on a fire-and-forget mode. We just notice the thread that we want to close, and the thread will do the close as soon as possible when it sees fit. We are not waiting for the close in order to get a response. Is there any case in which we care how/when we close? Additional mutex/conditional variables will have to be added in order to block the calling thread until the close is done if required.

AssemblyJohn:
Implemented comments. Some codacity issues are related to some 3rd _party code I wouldn't modify too much. This can't be merged with the required modifications in everest-core.

Pietfried (approved):
looks good from my side. Please wait for @hikinggrass approval as well before merging

AssemblyJohn:
I'd also require a review on: https://github.com/EVerest/everest-core/pull/631

This can't be merged without that.

include/ocpp/common/types.hpp
```diff
@@ -636,6 +636,41 @@ enum class MessageDirection {
...
+     * In such a case there was no close frame to extract a value from. This
+     * value is illegal on the wire.
+     */
+    AbnormalClose,
+    /// Indicates that the service is restarted. A client may reconnect and if
+    /// if it chooses to do so, should reconnect using a randomized delay of
+    /// 5-30s
+    ServiceRestart
+};
+
```
hikinggrass, line 673: Probably a good idea to link that this is coming from https://github.com/rstudio/websocket/blob/2161ee7de8f8782fae4d717c59e1292f5f5d9cf2/src/lib/websocketpp/close.hpp#L56

include/ocpp/common/websocket/websocket_uri.hpp
```diff
@@ -53,10 +54,10 @@ class Uri {
         return uri.str();
     }
 
-    websocketpp::uri get_websocketpp_uri() { // FIXME: wrap needed `websocketpp:uri` functionality inside `Uri`
-        return websocketpp::uri(this->secure, this->host, this->port,
-                                this->path_without_chargepoint_id /* is normalized with ending slash */ +
-                                    this->chargepoint_id);
+    websocketpp_utils::uri get_websocketpp_uri() { // FIXME: wrap needed `websocketpp:uri` functionality inside `Uri`
```
hikinggrass, line 57: Is that comment still relevant?

hikinggrass (changes requested):

include/ocpp/common/websocket/websocket_uri.hpp
AssemblyJohn: I think in the future we should do it manually, without any 3rd party, so in that case it would still be relevant.

Pietfried (approved):
looks good to me except a few nits

hikinggrass (approved):

PR #579, opened 2024-04-18
Drop transaction message from database if delivery attempts are exceeded

hikinggrass:
- [  ] If OCPP 2.0.1: I have updated the [OCPP 2.0.1 status document](https://github.com/EVerest/libocpp/tree/main/doc/ocpp_201_status.md)


Pietfried (approved):

marcemmers (commented):
I would expect the other remaining `remove_transaction_message` in `drop_update_messages_from_transactional_message_queue` to also use the `initial_unique_id` right? Or is it not needed there since those messages have never been attempted to send so they still have their original id in the message?

Other than that, lgtm.

PR #589, opened 2024-04-28
Guard 2.0.1 database transactions with a mutex

hikinggrass:
Otherwise it might happen that we try to insert metervalues into the database while another transaction event is happening. In certain circumstances (for example with multiple EVSEs present) this can lead to a "Could not begin transaction" exception

### DIFF
--- a/include/ocpp/common/database/database_connection.hpp
+++ b/include/ocpp/common/database/database_connection.hpp
@@ -17,7 +17,7 @@ namespace ocpp::common {
 class DatabaseTransactionInterface {
 public:
     /// \brief Destructor of transaction: Will by default rollback unless commit() is called
-    virtual ~DatabaseTransactionInterface();
+    virtual ~DatabaseTransactionInterface() = default;
 
     /// \brief Commits the transaction and release the lock on the database interface
     virtual void commit() = 0;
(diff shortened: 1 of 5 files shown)


lib/ocpp/v201/database_handler.cpp
```diff
@@ -51,6 +51,7 @@ void DatabaseHandler::init_enum_table_inner(const std::string& table_name, const
         throw std::runtime_error("Table does not exist.");
     }
 
+    std::scoped_lock lock(this->database_transaction_mutex);
```
marcemmers, line 54: I would suggest to do this more generically by putting this in the database connection somehow. Apparently you can't have more than 1 transaction per connection so that seems to be the perfect place. Lock on begin transaction and unlock on the commit or rollback.

marcemmers: For the record, we discussed this morning I will pick this one up ;)

include/ocpp/common/database/database_connection.hpp
```diff
@@ -11,6 +12,20 @@
 
 namespace ocpp::common {
 
+/// \brief Helper class for transactions. Will lock the database interface from new transaction until commit() or
+/// rollback() is caller or the object destroyed
```
Pietfried, line 16:
```suggestion
/// rollback() is called or the object destroyed
```

Pietfried (approved):

PR #597, opened 2024-04-30
Support change of a variable on the fly

jannejy:


lib/ocpp/v201/charge_point.cpp
```diff
@@ -808,6 +808,10 @@ void ChargePoint::on_security_event(const CiString<50>& event_type, const std::o
     this->security_event_notification_req(event_type, tech_info, false, false);
 }
 
+void ChargePoint::on_variable_changed(const SetVariableData& set_variable_data) {
```
marcemmers, line 811: Why didn't you make the function public instead of adding a wrapper?

marcemmers (approved):
There is a conflict now so please fix that. Other than that I'm ok with this after your explanation.

PR #601, opened 2024-05-02
Convert another logline to debug in libwebsockets

marcemmers:

### DIFF
--- a/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
+++ b/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
@@ -591,8 +591,8 @@ bool WebsocketTlsTPM::connect() {
         return false;
     }
 
-    EVLOG_info << "Connecting to uri: " << this->connection_options.csms_uri.string()
-               << " with security-profile " << this->connection_options.security_profile
+    EVLOG_info << "Connecting to uri: " << this->connection_options.csms_uri.string() << " with security-profile "
+               << this->connection_options.security_profile
                << (this->connection_options.use_tpm_tls ? " with TPM keys" : "");
 
     // new connection context
--- a/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
+++ b/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
@@ -916,7 +916,7 @@ void WebsocketTlsTPM::on_writable() {
 
         // This message was polled in a previous iteration
         if (message->sent_bytes >= message->payload.length()) {
-            EVLOG_info << "Websocket message fully written, popping processing thread from queue!";
+            EVLOG_debug << "Websocket message fully written, popping processing thread from queue!";
 
             // If we have written all bytes to libwebsockets it means that if we received
             // this writable callback everything is sent over the wire, mark the message
--- a/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
+++ b/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
@@ -668,15 +668,14 @@ bool WebsocketTlsTPM::connect() {
         return !local_data->is_connecting() && EConnectionState::INITIALIZE != local_data->get_state();
     });
 
-    EVLOG_debug << "Connect finalized with state: " << (int)local_data->get_state() << " Timeouted: " << timeouted;
     bool connected = (local_data->get_state() == EConnectionState::CONNECTED);
 
     if (!connected) {
-        EVLOG_debug << "Conn failed, interrupting.";
+        EVLOG_info << "Connect failed with state: " << (int)local_data->get_state() << " Timeouted: " << timeouted;
 
         // If we timeouted the on_conn_fail was not dispatched, since it did not had the chance
         if (timeouted && local_data->get_state() != EConnectionState::ERROR) {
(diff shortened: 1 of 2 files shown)


Pietfried (approved):
I was thinking about that one as well, since it only logs on larger messages, but ok for me to reduce log level here as well

hikinggrass (approved):

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -701,16 +697,13 @@ bool WebsocketTlsTPM::connect() {
 }
 
 void WebsocketTlsTPM::reconnect(long delay) {
-    EVLOG_info << "Attempting TLS TPM reconnect with delay:" << delay;
```
AssemblyJohn, line 704: I would like this to be kept.

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -672,15 +668,15 @@ bool WebsocketTlsTPM::connect() {
         return !local_data->is_connecting() && EConnectionState::INITIALIZE != local_data->get_state();
     });
 
-    EVLOG_info << "Connect finalized with state: " << (int)local_data->get_state() << " Timeouted: " << timeouted;
+    EVLOG_debug << "Connect finalized with state: " << (int)local_data->get_state() << " Timeouted: " << timeouted;
```
AssemblyJohn, line 671: This could be moved at line 675, and be kept as 'info' because it should not happen a lot of times.

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -547,9 +547,9 @@ void WebsocketTlsTPM::client_loop() {
     }
 
     // Print data for debug
-    EVLOG_info << "LWS connect with info "
-               << "port: [" << i.port << "] address: [" << i.address << "] path: [" << i.path << "] protocol: ["
-               << i.protocol << "]";
+    EVLOG_debug << "LWS connect with info "
```
AssemblyJohn, line 550: I would keep this as info, it only happens on a connect should not be too spammy.

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
marcemmers: We print the exact same thing on line 716. If you think it is better I can move remove the line on 716 and leave this one instead?

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
marcemmers: It also seems to happen on every reconnect attempt in case the connection does not work. But that is not the reason I removed this one. We also already print the same information in `Connecting to uri: etc` so I didn't really see a reason to print the same information twice.

Could you elaborate on what you are using from this?

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
AssemblyJohn: I've simply missed that one sorry :). I think this would be better, because if we have a crash or anything in the 'close' this would be the last message printed and would give us a better clue on where the problem might be.

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
AssemblyJohn: Sometimes there are some info missing from the uri (in certain contexts), maybe extend this one a bit with TPM and remove the second one?

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
marcemmers: I will move it to the middle then since it would be a bit strange to print `reconnecting` and immediately after that print `not reconnecting`.

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
marcemmers: I do think the second one is more readable to leave in in the future. I will set this back to info for now, there is already a comment above it that it is for debug anyway so will hopefully be removed sometime in the future.

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -1148,7 +1139,7 @@ int WebsocketTlsTPM::process_callback(void* wsi_ptr, int callback_reason, void*
     } break;
 
     case LWS_CALLBACK_CLIENT_CONNECTION_ERROR:
-        EVLOG_error << "CLIENT_CONNECTION_ERROR: " << (in ? reinterpret_cast<char*>(in) : "(null)");
+        EVLOG_debug << "CLIENT_CONNECTION_ERROR: " << (in ? reinterpret_cast<char*>(in) : "(null)");
```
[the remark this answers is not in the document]
marcemmers:
```suggestion
        EVLOG_error << "CLIENT_CONNECTION_ERROR: " << (in ? reinterpret_cast<char*>(in) : "(null)");
```

AssemblyJohn (approved):

Issue #595, opened 2024-04-29
Graceful exception handling of database operations


Pietfried:
### OCPP Version

OCPP1.6, OCPP2.0.1

### Describe the problem

The database implementation of OCPP1.6 and OCPP2.0.1 throws exceptions that are not always handled by the instance that called the functions that throw. This causes libocpp to crash, where in many cases this is not the desired outcome. 

### Describe your solution

The goal of this issue is to refactor the database implementation and the calling instances to add a more graceful handling of certain exceptions. In many cases, logging the failed operation or a retry mechanism is more applicable than a crash of the application.

PR #602, opened 2024-05-02
Refactor Exception Handling of Database Handlers

Pietfried:
## Describe your changes
Refactor database handlers and consumers of database handlers:
* All public database operations throw exceptions in case the database operation failed
* Added try-catch blocks to all function calls of database operations including appropriate handling of the failed operations

The primary design decision in the development of DatabaseHandler classes is to allow exceptions to propagate to higher level logic rather than catching them internally. By propagating exceptions of database operations, it is ensured the consumer application is fully aware of any issues that occur during database operations. This transparency allows higher-level logic to make informed decisions based on the specific errors encountered.

https://github.com/EVerest/libocpp/issues/595


lib/ocpp/v16/database_handler.cpp
```diff
@@ -377,21 +398,31 @@ void DatabaseHandler::insert_or_update_local_authorization_list_entry(const CiSt
...
+                this->delete_local_authorization_list_entry(authorization_data.idTag.get().c_str());
+            }
+        } catch (const QueryExecutionException& e) {
+            // catch but continue with remaining entries
+            success = false;
         }
     }
+
+    if (!success) {
+        throw QueryExecutionException(this->database->get_error_message());
```
hikinggrass, line 424: Does this really always provide a meaningful error message? I'm not 100% sure here, for example if the first insert_or_update fails and all subsequent once succeed get_error_message (which uses sqlite3_errmsg) may have already been cleared: https://www.sqlite.org/capi3ref.html#sqlite3_errcode

lib/ocpp/v201/database_handler.cpp
```diff
@@ -512,24 +517,28 @@ bool DatabaseHandler::transaction_metervalues_clear(const std::string& transacti
...
-    while (select_stmt->step() == SQLITE_ROW) {
+    int status;
+    while ((status = select_stmt->step()) == SQLITE_ROW) {
         auto row_id = select_stmt->column_int(0);
         delete_stmt->bind_int("@row_id", row_id);
 
         if (delete_stmt->step() != SQLITE_DONE) {
-            EVLOG_error << "Could not delete from table: " << this->database->get_error_message();
+            throw QueryExecutionException(this->database->get_error_message());
+            ;
```
hikinggrass, line 527: superfluous ;

lib/ocpp/v201/database_handler.cpp
```diff
@@ -231,21 +225,30 @@ void DatabaseHandler::insert_or_update_local_authorization_list_entry(const IdTo
...
+                this->delete_local_authorization_list_entry(authorization_data.idToken);
+            }
+        } catch (const QueryExecutionException& e) {
+            // catch but continue with remaining entries
+            success = false;
         }
     }
+
+    if (!success) {
+        throw QueryExecutionException(this->database->get_error_message());
```
hikinggrass, line 250: same concerns regarding the reported error message as in the 1.6 code

hikinggrass (changes requested):
Looks good, just some minor remarks

hikinggrass (approved):

include/ocpp/common/message_queue.hpp
```diff
@@ -242,7 +248,13 @@ template <typename M> class MessageQueue {
             if (remove_next_update_message && element->isTransactionUpdateMessage() &&
                 transaction_message_queue.size() > 1) {
                 EVLOG_debug << "Drop transactional message " << element->initial_unique_id;
-                database_handler->remove_transaction_message(element->initial_unique_id);
+                try {
+                    database_handler->remove_transaction_message(element->initial_unique_id);
+                } catch (const QueryExecutionException& e) {
+                    EVLOG_warning << "Could not delete message from transaction queue: " << e.what();
+                } catch (const std::exception& e) {
```
marcemmers, line 255: Is there a reason we still have these ones? Just as a "catch all" thing to make sure we don't crash I assume?

lib/ocpp/common/database/database_connection.cpp
```diff
@@ -28,8 +29,11 @@ class DatabaseTransaction : public DatabaseTransactionInterface {
     }
 
     void commit() override {
-        this->database.execute_statement("COMMIT TRANSACTION");
+        const auto retval = this->database.execute_statement("COMMIT TRANSACTION");
```
marcemmers, line 32: I am wondering if `execute_statement` should not throw if it fails as well. Maybe that is for another PR then.

include/ocpp/common/message_queue.hpp
Pietfried: Yes I tried to keep the `std::exception` where it was already present

lib/ocpp/common/database/database_connection.cpp
Pietfried: I was thinking the same but I left it for now since its not part of any public api. No strong opinion here...

marcemmers (approved):
Like mentioned on zulip there are still some functions that return bool that don't have to anymore. Other than that, lgtm (y)

marcemmers (approved):

PR #603, opened 2024-05-03
Re-connect attempt on timeout

AssemblyJohn:
## Describe your changes

When there was a connection timeout, a re-connect was not re-attempted. The conn thread did not had the opportunity to dispatch the 'on_conn_fail' callback in the case of a timeout. The dispatch is manually send on a connection timeout, if the error state was not set.

### DIFF
--- a/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
+++ b/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
@@ -556,8 +556,10 @@ void WebsocketTlsTPM::client_loop() {
         // This condition can occur when connecting fails to an IP address
         // retries need to be attempted
         local_data->update_state(EConnectionState::ERROR);
-        conn_cv.notify_one();
         on_conn_fail();
+
+        // Notify conn waiter
+        conn_cv.notify_one();
     } else {
         EVLOG_debug << "Init client loop with ID: " << std::hex << std::this_thread::get_id();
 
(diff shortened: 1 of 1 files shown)


lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -675,6 +677,13 @@ bool WebsocketTlsTPM::connect() {
 
     if (!connected) {
         EVLOG_error << "Conn failed, interrupting.";
+
+        // If we timeouted the on_conn_fail was not dispatched, since it did not had the chance
+        if (timeouted && local_data->get_state() != EConnectionState::ERROR) {
```
james-ctc, line 682: would it be safer to have this code after the check that local_data != nullptr?
e.g.
```
std::shared_ptr<ConnectionData> local_data = conn_data;
if (local_data != nullptr) {
        if (timeouted && local_data->get_state() != EConnectionState::ERROR) {
            EVLOG_error << "Conn failed with timeout, without disconnect dispatch, dispatching manually.";
            on_conn_fail();
        }
        local_data->do_interrupt();
}
```

james-ctc: I think there was a problem where local_data had been reset (i.e. nullptr) and local_data->do_interrupt(); was terminating OCPP.

james-ctc: Originally the code was conn_data->do_interrupt(); which was a problem because another thread could reset conn_data causing the invalid pointer dereference.
This was solved by taking a copy of conn_data (local_data) however while waiting on the condition variable another thread can change conn_data. After the wait is complete conn_data is re-fetched to ensure that do_interrupt() is called on the most up-to-date information.

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -675,13 +677,23 @@ bool WebsocketTlsTPM::connect() {
...
-        conn_data.reset();
+        // Interrupt and drop the connection data
+        local_data->do_interrupt();
+
+        // Also interrupt the latest conenction, if it was set by a parallel thread
+        auto local = conn_data;
+
+        if (local != nullptr) {
+            local->do_interrupt();
+            conn_data.reset();
```
james-ctc, line 695: should conn_data always be reset?

AssemblyJohn: if we checked and it is always null, whould we re-set an already null null pointer? During the execution of this, I don't think that any thread should run in paralel that can somehow set the stat of the conn_data in any way, Have you detected any possibility of that in code?

james-ctc: With multiple threads there is always a possibility of a race condition unless critical sections are protected by mutexes.
e.g.
thread 1: line  961 auto local = conn_data; and then gets swapped out
thread 2: calls connect() ... and updates conn_data line 609 conn_data = local_data; and then gets swapped out

There is a comment in the code that connect() "Will be called from external threads as well"

james-ctc (approved):
looks good to me

Pietfried (approved):

PR #615, opened 2024-05-13
Add thread that executes the callbacks via a queue to synchronize them

marcemmers:

### DIFF
--- a/include/ocpp/common/websocket/websocket_libwebsockets.hpp
+++ b/include/ocpp/common/websocket/websocket_libwebsockets.hpp
@@ -111,7 +111,7 @@ private:
     std::queue<std::function<void()>> deferred_callback_queue;
     std::mutex deferred_callback_mutex;
     std::condition_variable deferred_callback_cv;
-    bool stop_deferred_handler;
+    std::atomic_bool stop_deferred_handler;
 };
 
 } // namespace ocpp
--- a/include/ocpp/common/websocket/websocket_libwebsockets.hpp
+++ b/include/ocpp/common/websocket/websocket_libwebsockets.hpp
@@ -82,9 +82,7 @@ private:
     void handle_deferred_callback_queue();
 
     /// \brief Add a callback to the queue of callbacks to be executed. All will be executed from a single thread
-    void push_deferred_callback(std::function<void()> callback);
-
-    void stop_deferred_callback_handler();
+    void push_deferred_callback(const std::function<void()> &callback);
 
 private:
     std::shared_ptr<EvseSecurity> evse_security;
@@ -109,7 +107,7 @@ private:
     std::condition_variable recv_message_cv;
     std::string recv_buffered_message;
 
-    std::thread deferred_callback_thread;
+    std::unique_ptr<std::thread> deferred_callback_thread;
     std::queue<std::function<void()>> deferred_callback_queue;
     std::mutex deferred_callback_mutex;
     std::condition_variable deferred_callback_cv;
--- a/include/ocpp/common/websocket/websocket_libwebsockets.hpp
+++ b/include/ocpp/common/websocket/websocket_libwebsockets.hpp
@@ -82,7 +82,7 @@ private:
     void handle_deferred_callback_queue();
 
     /// \brief Add a callback to the queue of callbacks to be executed. All will be executed from a single thread
-    void push_deferred_callback(const std::function<void()> &callback);
+    void push_deferred_callback(const std::function<void()>& callback);
 
 private:
     std::shared_ptr<EvseSecurity> evse_security;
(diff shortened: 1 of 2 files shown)


lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -1298,4 +1295,25 @@ int WebsocketTlsTPM::process_callback(void* wsi_ptr, int callback_reason, void*
...
 }
 
+void WebsocketTlsTPM::push_deferred_callback(std::function<void()> callback) {
+    std::scoped_lock tmp_lock(this->deferred_callback_mutex);
+    this->deferred_callback_queue.push(callback);
+    this->deferred_callback_cv.notify_one();
+}
+
+void WebsocketTlsTPM::handle_deferred_callback_queue() {
+    while (true) {
```
AssemblyJohn, line 1305: Does it look OK to always execute this thread? Maybe when we close the connection we should somehow have a way to end the thread and also join it in the destructor, not good practice to leave a thread running in the background.

marcemmers: Ah yes you are right. I am relatively new to threads and expected it to be handled cleanly by the thread destructor but that is not the case. I will add this.

marcemmers: I think it should be fine to leave the thread alive while we are closed right? Otherwise we might open up issues with callbacks not being called etc.

AssemblyJohn: That is correct. The behavior should be as following: 
1. Have the thread always run if there are actions in the queue or if there is a 'live' connection (conn_data.is_valid() or something along those lines)
2. If the connection is closed, and the queue is not empty, still keep the thread running until the queue is depleted
3. If a 'reconnect' is initiated and the thread is already running do NOT start a new thread but re-use the existing thread

hikinggrass: however maybe we want to be able to stop this thread when we shutdown completely, so having this use a member variable instead of just "true" (that could potentially be set from the outside to "false") is probably not a bad idea

marcemmers: I do that using a break statement instead of checking it in the while loop. I can add it/move it to the while as well if you prefer?

hikinggrass: Ah you're right, must have missed that earlier somehow. No than it's fine I would say

hikinggrass (approved):

include/ocpp/common/websocket/websocket_libwebsockets.hpp
```diff
@@ -100,6 +106,12 @@ class WebsocketTlsTPM final : public WebsocketBase {
     std::queue<std::string> recv_message_queue;
     std::condition_variable recv_message_cv;
     std::string recv_buffered_message;
+
+    std::unique_ptr<std::thread> deferred_callback_thread;
+    std::queue<std::function<void()>> deferred_callback_queue;
+    std::mutex deferred_callback_mutex;
+    std::condition_variable deferred_callback_cv;
+    bool stop_deferred_handler;
```
AssemblyJohn, line 114: For the sake of correction, since this variable is used from multiple threads, let's keep the mode we used bools in the .cpp file and use a 'std::atomic_bool' here.

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -1288,4 +1294,31 @@ int WebsocketTlsTPM::process_callback(void* wsi_ptr, int callback_reason, void*
...
+        std::function<void()> callback;
+        {
+            std::unique_lock lock(this->deferred_callback_mutex);
+            this->deferred_callback_cv.wait(
```
AssemblyJohn, line 1308: While the condition is correct, please use a **wait_for** (like the recv_loop) with a timeout of 1 second. It should be enough to free up CPU and still sleep most of the time when we don't have a callback in the queue.

marcemmers: I've changed this but I still don't like it since now the following code also needs to be changed since I can get past the `wait_for` even if there is nothing to do. I am really curious about your experience with the 0.01% of the cases where the notification gets lost. I did not find any similar experiences, the only thing is that people can implement the variable incorrectly or use the wrong mutex and stuff like that.

I'd rather leave it a `wait` instead since that is what we are always doing here.

AssemblyJohn (approved):

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
marcemmers: Any thoughts here @AssemblyJohn

AssemblyJohn: Except the previous discussion I have no further comments.

PR #618, opened 2024-05-14
fix: OCPP message handling on invalid messages

james-ctc:
## Describe your changes

When there is a problem converting a string into an enum std::out_of_range in thrown rather than a JSON exception. A general catch was added so that a CallError is sent (rather than terminating EVerest)

In the specific case of TriggerMessage and ExtendedTriggerMessage the response NotImplemented is sent in the CallResult.

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
 #include "ocpp/common/types.hpp"
+#include <stdexcept>
 #include <thread>
 
 #include <everest/logging.hpp>
@@ -1092,6 +1093,15 @@ void ChargePointImpl::message_callback(const std::string& message) {
         }
     } catch (json::exception& e) {
         EVLOG_error << "JSON exception during handling of message: " << e.what();
+        if (json_message.is_array() && json_message.size() > MESSAGE_ID) {
+            auto call_error = CallError(MessageId(json_message.at(MESSAGE_ID).get<std::string>()), "FormationViolation",
+                                        e.what(), json({}, true));
+            this->send(call_error);
+            this->securityEventNotification(ocpp::security_events::INVALIDMESSAGES, message, true);
+        }
+    } catch (std::out_of_range e) {
+        // std::out_of_range is used when converting strings to enums
+        EVLOG_error << "Out of range exception during handling of message: " << e.what();
         if (json_message.is_array() && json_message.size() > MESSAGE_ID) {
             auto call_error = CallError(MessageId(json_message.at(MESSAGE_ID).get<std::string>()), "FormationViolation",
                                         e.what(), json({}, true));


Pietfried:
Hi @james-ctc , adding the general catch in `message_callback` looks good to me, but I dont understand why we need specific catch handling for the TriggerMessages. Maybe Im missing something but in case there is an `out_of_range` error when constructing the `Call<TriggerMessageRequest>` we should also respond with a CallError rather than responding with `NotSupported` in my opinion.

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -1098,6 +1099,15 @@ void ChargePointImpl::message_callback(const std::string& message) {
             this->send(call_error);
             this->securityEventNotification(ocpp::security_events::INVALIDMESSAGES, message, true);
         }
+    } catch (std::out_of_range e) {
+        // std::out_of_range is used when converting strings to enums
+        EVLOG_error << "JSON exception during handling of message: " << e.what();
```
Pietfried, line 1104:
```suggestion
        EVLOG_error << "Out of range exception during handling of message: " << e.what();
```

james-ctc: OCPP 1.6J specification 7.44. TriggerMessageStatus states:
"NotImplemented Requested notification cannot be sent because it is either not implemented or unknown."
In the specific failure I was investigating there was a trigger message with an unrecognised requestedMessage. The JSON was correctly formatted. Since the MessageTrigger was not in the enumeration it meets the requirement for a NotImplemented status since the request was unknown.

Pietfried: Which one was it? All of them should be supported and in case we get an out of range error it should be because the `requestedMessage` has an invalid value, e.g. is no value defined by the `MessageTrigger` enumeration.

james-ctc:
(moved comment - it didn't relate to the suggested change)

The request was for message not in the enumeration. Hence unknown. Which is specifically mentioned as requiring a `NotImplemented` response rather than CallError.
MessageTrigger has:
- BootNotification
- DiagnosticsStatusNotification
- FirmwareStatusNotification
- Heartbeat
- MeterValues
- StatusNotification

Where a message isn't implemented then `NotImplemeted` is returned.
When it isn't in that list it is unknown. The OCPP spec requires that `NotImplemeted` is returned for both not implemented and not known cases.

were `requestedMessage` missing from the request it should result in a CallError since `requestedMessage` is a required field in the request.

i.e. the test was specifically checking that a request not is the `MessageTrigger` enum correctly returns a status of `NotImplemented` as required by section 7.44. TriggerMessageStatus

"NotImplemented Requested notification cannot be sent because it is either **not implemented** or **unknown**."

Has the requirement not had the text **or unknown** then I would agree that a CallError would be correct.

Pietfried:
> (moved comment - it didn't relate to the suggested change)
> 
> The request was for message not in the enumeration. Hence unknown. Which is specifically mentioned as requiring a `NotImplemented` response rather than CallError. MessageTrigger has:
> 
> * BootNotification
> * DiagnosticsStatusNotification
> * FirmwareStatusNotification
> * Heartbeat
> * MeterValues
> * StatusNotification
> 
> Where a message isn't implemented then `NotImplemeted` is returned. When it isn't in that list it is unknown. The OCPP spec requires that `NotImplemeted` is returned for both not implemented and not known cases.
> 
> were `requestedMessage` missing from the request it should result in a CallError since `requestedMessage` is a required field in the request.
> 
> i.e. the test was specifically checking that a request not in the `MessageTrigger` enum correctly returns a status of `NotImplemented` as required by section 7.44. TriggerMessageStatus
> 
> "NotImplemented Requested notification cannot be sent because it is either **not implemented** or **unknown**."
> 
> Had the requirement not had the text **or unknown** then I would agree that a CallError would be correct.

I understand your reasoning but I don't see how a `requestedMessage` can be considered unknown. Its a finite set of values and all values are implemented by libocpp. The CSMS violates the specification if it uses any other value. 

The OCPP-J 1.6 Specification defines the following error code that can be used as part of a CallError.

> "PropertyConstraintViolation: Payload is syntactically correct but at least one
field contains an invalid value".

I think there are reasons for both options but from a semantic point of view I would rather have a `NotImplementd` separated cleary from and invalid value for `requestedMessage`, so I would rather use the CallError(PropertyConstraintViolation).

james-ctc:
The spec allows for unknown as an option - so I guess we will need to get clarification from the OCA as to how unknown entries should be handled.

james-ctc:
specific case handling for TriggerMessage and ExtendedTriggerMessage removed - CallError will result should a value outside of the enum be received.

Pietfried (approved):

PR #629, opened 2024-05-23
Persist queued changeAvailabilty

bWF0dGhpYXMK:
## Describe your changes
The queued changeAvailabilty.req (which is answered with scheduled) is store also in database, to be reboot safe.


lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -1350,6 +1350,11 @@ void ChargePointImpl::preprocess_change_availability_request(
             }
         }
 
+        // We store the queued request already in the database, if powerloss the "new" state is available
```
Pietfried, line 1353:
```suggestion
        // We store the queued request in the database, so in case of a powerloss the operational state is persisted.
```

Pietfried (approved):

Issue #633, opened 2024-05-23
Security Event notifications sent even when regisration is not accepted


SNSubramanya:
### OCPP Version

OCPP2.0.1

### Describe the bug

According to specs no messages must be sent to the CSMS unless the registration status is Accepted. But it looks like the security event notification is sent regardless of the registration status. 

### To Reproduce

- Boot up CS
- Send BootnotificationRequest to CSMS
- CS received BoorNotification Reposnse with status "Pending" or "Rejected"
- CS sends Security event notification

### Anything else?
This seems to be due to a missing requirement that hasn't been implemented yet (see B03.FR.03). So perhaps the right way to go about this is to modify the message handling in the queue so that no messages are sent while registration status isn't accepted.

PR #637, opened 2024-05-24
Message Queue Refactoring

Pietfried:
## Describe your changes

This PR picks up the issue that OCPP messages could theoretically be sent before the CSMS accepted a BootNotification.req. Only BootNotification.req itself or messages triggered during `Pending` shall be sent before the BootNotification.req was accepted. 

It can and should happen though that messages are queued by the charge point when the BootNotification.req was not yet Accepted. This could e.g. be when:
* ChargePoint is offline
* BootNotification.req was responded by CSMS with `Pending` or `Rejected`

During this time the consumer of libocpp can call functions to initate OCPP messages (`on_session_started` , `on_security_event`, `on_firmware_update_notiifcation`, ...). Especially transaction related messages shall not be discarded but queued and sent when the RegistrationStatus becomes `Accepted`.

Especially for OCPP2.0.1 there was no generic handling in place that avoids these messages from being queued during a `Pending`, `Rejected` state, so there was no generic prevention of sending these messages even when it is not allowed.

__________________________________________

This PR refactors the sending and processing of OCPP messages.

It introduces a `SendAttemptResult` enum that contains the resulting values of an attempt to send/queue a message. Messages shall either be:
* SEND_IMMEDIATELY - in case the RegistrationStatus is `Accepted` OR the MessageType is `BootNotification` OR the message is triggered by `TriggerMessage.req`
* SEND_AFTER_REGISTRATION_STATUS_ACCEPTED - in case RegistrationStatus is not `Accepted` AND the message is transaction-related
* DISCARD(ed) - in case RegistrationStatus is not `Accepted` AND the message is not transaction-related
 
Depending on this result messages that are queued are labled in case they should be `SEND_AFTER_REGISTRATION_STATUS_ACCEPTED`. In case the RegistrationStatus is not `Accepted` other messages like `BootNotification` or messages triggered by a `TriggerMessage.req` are allowed and enabled to jump the queue.

https://github.com/EVerest/libocpp/issues/633


include/ocpp/common/message_queue.hpp
```diff
@@ -342,44 +345,52 @@ template <typename M> class MessageQueue {
...
-                    auto& normal_message = this->normal_message_queue.front();
-                    EVLOG_debug << "normal msg timestamp: " << normal_message->timestamp;
-                    if (normal_message->timestamp <= now) {
-                        EVLOG_debug << "normal message timestamp <= now";
-                        message = normal_message;
-                        queue_type = QueueType::Normal;
-                    } else {
-                        EVLOG_error << "A normal message should not have a timestamp in the future: "
-                                    << normal_message->timestamp << " now: " << now;
+                for (auto normal_message_it = normal_message_queue.begin();
```
marcemmers, line 351: Seems like this searching for a message is kind of duplicated, once for normal and once for transactions. Could be a good candidate to move into a helper function I would say.

Also, first checking for transaction message and then checking the other queue would make sense. Now we always check the normal queue and then possibly overwrite using the transaction queue.

include/ocpp/common/message_queue.hpp
```diff
@@ -498,20 +509,20 @@ template <typename M> class MessageQueue {
     }
 
     /// \brief pushes a new \p call message onto the message queue
-    template <class T> void push(Call<T> call) {
+    template <class T> void push(Call<T> call, const bool stall_until_accepted = false) {
```
marcemmers, line 512: Maybe a bit of a nit but wouldn't true make more sense as the default argument since most messages need to be stalled?

include/ocpp/v201/charge_point.hpp
```diff
@@ -697,12 +705,38 @@ class ChargePoint : public ChargePointInterface, private ocpp::ChargingStationBa
     void handle_data_transfer_req(Call<DataTransferRequest> call);
 
     // general message handling
-    template <class T> bool send(ocpp::Call<T> call) {
-        this->message_queue->push(call);
-        return true;
+    template <class T> bool send(ocpp::Call<T> call, const bool initiated_by_trigger_message = false) {
```
marcemmers, line 708: Can we move the implementation to the cpp file? Only the charge_point needs to call these functions so we don't need to expose the implementation.

include/ocpp/v201/charge_point.hpp
```diff
@@ -697,12 +705,38 @@ class ChargePoint : public ChargePointInterface, private ocpp::ChargingStationBa
...
+
+        switch (send_message_attempt_result) {
+        case SendAttemptResult::SEND_IMMEDIATELY:
+            this->message_queue->push(call);
+            return true;
+        case SendAttemptResult::SEND_AFTER_REGISTRATION_STATUS_ACCEPTED:
+            this->message_queue->push(call, true);
+            return true;
+        case SendAttemptResult::DISCARD:
+        default:
```
a-w50, line 720: Again, no default here!

include/ocpp/v201/charge_point.hpp
a-w50: Yes, this should be possible as this function is only called from within the `charge_point.cpp`.  Having a wrapper class for ocpp::Call<T> which uses type erasure (like `std::variant`), most of the template usage could be removed here.

a-w50 (changes requested):

include/ocpp/v201/charge_point.hpp
marcemmers: I don't necessary mean that the whole template function needs to be removed. More that template implementation does not have to reside in the header file. You can perfectly create the template prototype here and then define the template implementation in the CPP file as long as all the methods that use it have the actual implementation before they use the templated call.

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -301,20 +301,18 @@ void ChargePointImpl::boot_notification(bool initiated_by_trigger_message) {
 }
 
 void ChargePointImpl::clock_aligned_meter_values_sample() {
-    if (this->initialized) {
```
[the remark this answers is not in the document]
Pietfried: It was apparently indicating if a BootNotification.conf(Accepted) was received and used in some parts to check if its allowed to send/queue messages. This became obsolet now. I missed to also remove the property from the class and add it to the PR message

include/ocpp/common/message_queue.hpp
```diff
@@ -303,6 +305,7 @@ template <typename M> class MessageQueue {
         resuming(false),
         running(true),
         new_message(false),
+        is_registration_status_accepted(false),
```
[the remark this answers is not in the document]
Pietfried: It's a queue structure that is solely created to fullfil the requirements of the communication protocol. I get your concern but also think that this is debatable. It's definitely not something that I would like to add as part of this PR.

include/ocpp/common/message_queue.hpp
```diff
@@ -342,44 +345,52 @@ template <typename M> class MessageQueue {
...
+
+                    message = *transaction_message_it;
+                    queue_type = QueueType::Transaction;
+                    selected_transaction_message_it = transaction_message_it; // Save iterator to the selected message
+                    break; // Stop searching once a sendable message is found
                 }
 
                 if (message == nullptr) {
                     EVLOG_debug << "No message in queue ready to be sent yet";
+                    this->new_message = false;
```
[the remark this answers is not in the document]
Pietfried: See https://github.com/EVerest/libocpp/blob/main/include/ocpp/common/message_queue.hpp#L449 : `new_message` was set to false only when `transaction_message_queue` and `normal_message_queue` is empty to allow the cv to wait for the next notify when a new message is added to the queue.

With the changes added we can have a situation where there are messages on the queue, but we dont want to process the messages on the queue because its not allowed to send them, so we want to wait until the next message until we continue processing.

include/ocpp/common/message_queue.hpp
Pietfried: Only in the case the RegistrationStatus is Pending or Rejected messages will be pushed with `true` so most messages will likely be queued with `false`

include/ocpp/common/message_queue.hpp
```diff
@@ -63,6 +63,39 @@ enum class QueueType {
     None,
 };
 
+/// \brief Indicates the transmission priority of a message that is being pushed to the message queue
+enum class MessageTransmissionPriority {
+    SEND_IMMEDIATELY,                        // message can be queued and can be send immediately
+    SEND_AFTER_REGISTRATION_STATUS_ACCEPTED, // message can be queued and shall be send only if registration status is
+                                             // accepted
+    DISCARD                                  // message shall be discarded and not be sent
+};
```
a-w50, line 72: I guess the corresponding text in the PR message needs to be adapted

include/ocpp/common/message_queue.hpp
```diff
@@ -162,6 +191,17 @@ template <typename M> class MessageQueue {
...
+    bool allowed_to_send_message(const std::shared_ptr<ControlMessage<M>> message) {
+        if (message->stall_until_accepted and !this->is_registration_status_accepted) {
+            return false;
+        }
+
+        if (message->timestamp > DateTime()) {
+            return false;
+        }
+        return true;
+    }
```
a-w50, line 203: Again, I would write this as a free function and pass `is_registration_status_accepted` as a parameter.  Furthermore, `ControlMessage<M>` shouldn't be passed as a shared pointer, but by reference.
If written as a free function, you would need to write it as template (because of `ControlMessage<M>`).
Performance wise, `DateTime()` might not be instantiated and evaluated each and every time.

include/ocpp/common/message_queue.hpp
```diff
@@ -340,46 +381,49 @@ template <typename M> class MessageQueue {
                 }
 
                 // prioritize the message with the oldest timestamp
-                auto now = DateTime();
```
a-w50, line 343: Maybe the idea of a single timestamp was not that wrong, you can also pass it inside the lambda.

include/ocpp/common/message_queue.hpp
```diff
@@ -340,46 +381,49 @@ template <typename M> class MessageQueue {
...
+                                     // message from normal message queue is BootNotification, this is prioritized
+                                     if (message->messageType == M::BootNotification) {
+                                         return false;
+                                     }
+                                     // transaction messages is older than normal message, so select transaction message
+                                     if (msg->timestamp <= message->timestamp) {
+                                         return true;
+                                     }
+                                     return false;
+                                 });
```
a-w50, line 417: Due to the high indentation level, the formatting is a bit ugly.  What you can do in this case, and what in general is a good idea for *longer* lambdas, define them beforehand and give them a *self documenting* name like:
```
auto is_transaction_message_available = [&](const std::shared_ptr<ControlMessage<M>>& msg) {
  ...
};
const auto transaction_message_it = std::find_if(std::begin(transaction_message_queue), std::end(transaction_message_queue), is_transaction_message_available);
```
The name `is_transaction_message_available` is of course not completely correct, because it prioritizes some of the *normal* messages.  Also using `msg` and `message` is a bit unfortunate.  Also note, that starting with c++14 you can use aliases in the capture list, so you could have written
`auto ... = [this, &normal_message = message](const ... transaction_message) {}`.

include/ocpp/common/message_queue.hpp
```diff
@@ -671,7 +715,7 @@ template <typename M> class MessageQueue {
                 this->in_flight->message.at(CALL_ACTION).template get<std::string>() + std::string("Response"));
             this->in_flight->promise.set_value(enhanced_message);
 
-            if (this->in_flight->isTransactionMessage()) {
+            if (is_transaction_message(this->in_flight->messageType)) {
```
a-w50, line 718: you could also write an additional overload like:
```
template <typename M> auto is_transaction_message(const ControlMessage<M>& control_message) {
  return is_transaction_message(control_message.messageType);
}
```
and with that you could write here `is_transaction_message(*this->in_flight)`;

lib/ocpp/common/message_queue.cpp
```diff
@@ -7,43 +7,64 @@
...
+    return MessageTransmissionPriority::DISCARD;
 }
 
-template <> bool ControlMessage<v16::MessageType>::isTransactionMessage() const {
-    if (this->messageType == v16::MessageType::StartTransaction ||
-        this->messageType == v16::MessageType::StopTransaction || this->messageType == v16::MessageType::MeterValues ||
-        this->messageType == v16::MessageType::SecurityEventNotification) {
+bool is_transaction_message(const ocpp::v16::MessageType message_type) {
+    if (message_type == v16::MessageType::StartTransaction || message_type == v16::MessageType::StopTransaction ||
+        message_type == v16::MessageType::MeterValues || message_type == v16::MessageType::SecurityEventNotification) {
```
a-w50, line 26: I'd prefer to write here
```
return (message_type == v16::MessageType::StartTransaction || message_type == v16::MessageType::StopTransaction ||
        message_type == v16::MessageType::MeterValues || message_type == v16::MessageType::SecurityEventNotification);
```

lib/ocpp/common/message_queue.cpp
```diff
@@ -7,43 +7,64 @@
...
         return true;
     }
     return false;
 }
 
-template <> bool ControlMessage<v16::MessageType>::isTransactionUpdateMessage() const {
-    return (this->messageType == v16::MessageType::MeterValues);
+bool is_transaction_message(const ocpp::v201::MessageType message_type) {
+    if (message_type == v201::MessageType::TransactionEvent ||
```
a-w50, line 33: same as above

tests/lib/ocpp/common/test_message_queue.cpp
```diff
@@ -120,17 +120,16 @@ std::ostream& operator<<(std::ostream& os, const TestMessageType& message_type)
     return os;
 };
 
-template <> bool ControlMessage<TestMessageType>::isTransactionMessage() const {
-    return this->messageType == TestMessageType::TRANSACTIONAL ||
-           this->messageType == TestMessageType::TRANSACTIONAL_UPDATE;
+bool is_transaction_message(const TestMessageType message_type) {
+    return message_type == TestMessageType::TRANSACTIONAL || message_type == TestMessageType::TRANSACTIONAL_UPDATE;
```
a-w50, line 124: For these kind of expressions it improves readability if you add braces like
```
(message_type == TestMessageType::TRANSACTIONAL) || (message_type == TestMessageType::TRANSACTIONAL_UPDATE)
```
because sometimes the operator precedence is not that obvious

a-w50 (changes requested):
Just some minor things

lib/ocpp/common/message_queue.cpp
```diff
@@ -7,43 +7,58 @@
...
+bool is_transaction_message(const ocpp::v16::MessageType message_type) {
+    return message_type == v16::MessageType::StartTransaction || message_type == v16::MessageType::StopTransaction ||
+           message_type == v16::MessageType::MeterValues || message_type == v16::MessageType::SecurityEventNotification;
 }
 
-template <> bool ControlMessage<v16::MessageType>::isBootNotificationMessage() const {
-    return this->messageType == v16::MessageType::BootNotification;
+bool is_transaction_message(const ocpp::v201::MessageType message_type) {
+    return message_type == v201::MessageType::TransactionEvent ||
+           message_type == v201::MessageType::SecurityEventNotification;
```
a-w50, line 31: would also parenthesize these

a-w50 (approved):
:+1:

include/ocpp/common/message_queue.hpp
```diff
@@ -71,26 +64,78 @@ template <typename M> struct ControlMessage {
...
+    Normal,
+    Transaction,
+    None,
+};
 
-    /// \brief Determine whether message is a BootNotification.
-    bool isBootNotificationMessage() const;
+/// \brief Indicates the transmission priority of a message that is being pushed to the message queue
+enum class MessageTransmissionPriority {
+    SEND_IMMEDIATELY,                        // message can be queued and can be send immediately
```
marcemmers, line 91: Why the all caps? Would be better to keep it consistent I would say and leave them CamelCase like we do with all enum classes.

lib/ocpp/common/message_queue.cpp
```diff
@@ -7,43 +7,59 @@
 
 namespace ocpp {
 
-template <> ControlMessage<v16::MessageType>::ControlMessage(const json& message) {
-    this->message = message.get<json::array_t>();
-    this->messageType = v16::conversions::string_to_messagetype(message.at(CALL_ACTION));
-    this->message_attempts = 0;
-    this->initial_unique_id = this->message[MESSAGE_ID];
-}
+MessageTransmissionPriority get_message_transmission_priority(bool is_boot_notification_message, bool triggered,
```
marcemmers, line 10: Just a heads up, this file does not exist anymore since the v16/v201 compilation change so you'd have to split the changes up too.

marcemmers (approved):
Small review comments and I agree with some comments made by @a-w50 but those might be for later improvements.

hikinggrass (approved):

PR #643, opened 2024-05-29
Clean database close attempt

AssemblyJohn:
## Describe your changes

Added a clean database close in the SQL destructor.

### DIFF
--- a/include/ocpp/v201/device_model_storage_sqlite.hpp
+++ b/include/ocpp/v201/device_model_storage_sqlite.hpp
@@ -8,6 +8,7 @@
 #include <sqlite3.h>
 
 #include <everest/logging.hpp>
+#include <ocpp/common/database/database_connection.hpp>
 #include <ocpp/v201/device_model_storage.hpp>
 
 namespace ocpp {
@@ -16,7 +17,7 @@ namespace v201 {
 class DeviceModelStorageSqlite : public DeviceModelStorage {
 
 private:
-    sqlite3* db;
+    std::unique_ptr<ocpp::common::DatabaseConnectionInterface> db;
 
     int get_component_id(const Component& component_id);
 
@@ -26,6 +27,7 @@ public:
     /// \brief Opens SQLite connection at given \p db_path
     /// \param db_path  path to database
     explicit DeviceModelStorageSqlite(const fs::path& db_path);
+    ~DeviceModelStorageSqlite() = default;
 
     std::map<Component, std::map<Variable, VariableMetaData>> get_device_model() final;
 
--- a/lib/ocpp/v201/device_model_storage_sqlite.cpp
+++ b/lib/ocpp/v201/device_model_storage_sqlite.cpp
@@ -12,7 +12,9 @@ using namespace common;
 namespace v201 {
 
 DeviceModelStorageSqlite::DeviceModelStorageSqlite(const fs::path& db_path) {
-    if (sqlite3_open(db_path.c_str(), &this->db) != SQLITE_OK) {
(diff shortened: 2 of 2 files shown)


lib/ocpp/v201/device_model_storage_sqlite.cpp
```diff
@@ -20,6 +20,12 @@ DeviceModelStorageSqlite::DeviceModelStorageSqlite(const fs::path& db_path) {
     }
 }
 
+DeviceModelStorageSqlite::~DeviceModelStorageSqlite() {
+    if (sqlite3_close(this->db) != SQLITE_OK) {
```
marcemmers, line 24: I would take a look at how its done in `close_connection_internal` in the `database_connection.cpp` file. You could also finalize all open transactions and use `_v2` close.

AssemblyJohn: I think the best way forward would be to refactor this a bit and use the database_connection which already takes care of many of the details that you have mentioned.

marcemmers: Yes I think you are right, just didn't want to blow up the PR like that. In the long run this is better :+1:

AssemblyJohn:
Updated to use the new auto-managed interface.

include/ocpp/v201/device_model_storage_sqlite.hpp
```diff
@@ -26,6 +27,7 @@ class DeviceModelStorageSqlite : public DeviceModelStorage {
     /// \brief Opens SQLite connection at given \p db_path
     /// \param db_path  path to database
     explicit DeviceModelStorageSqlite(const fs::path& db_path);
+    ~DeviceModelStorageSqlite();
```
marcemmers, line 30:
```suggestion
    ~DeviceModelStorageSqlite() = default;
```

This should work too right?

Pietfried (approved):
lgtm

christopher-davis-afs:
The changes here for the `DeviceModelStorageSqlite` class seem to have broken [our tests](https://github.com/EVerest/libocpp/actions/runs/9355354996/job/25750301770?pr=611). It doesn't seem to work with in-memory databases properly?

Are there changes we need to make to [our database setup](https://github.com/EVerest/libocpp/pull/611/files#diff-acd0468b0fc7d05e101c753455aa6e82b0d2efbb9256c1a1fd1891c1e7ca7c40R157-R181)?

christopher-davis-afs:
I found the fix. In `DatabaseConnection::open_connection()` you have the following:

```cpp
    if (this->database_file_path.string().find(":memory:") == std::string::npos and
        !fs::exists(this->database_file_path.parent_path())) {
        fs::create_directories(this->database_file_path.parent_path());
    }
```

[SQLite has multiple ways to initialize in-memory databases](https://www.sqlite.org/inmemorydb.html):

> If two or more distinct but shareable in-memory databases are needed in a single process, then the [mode=memory](https://www.sqlite.org/uri.html#coreqp) query parameter can be used with a [URI filename](https://www.sqlite.org/uri.html) to create a named in-memory database:

The code above only checks for `:memory:` and not `mode=memory`. We need to use `mode=memory` to give the device model database a unique name so we don't mangle the device model db and other dbs. So the code needs to look like this:

```cpp
    if (this->database_file_path.string().find(":memory:") == std::string::npos and
        this->database_file_path.string().find("mode=memory") == std::string::npos and
        !fs::exists(this->database_file_path.parent_path())) {
        fs::create_directories(this->database_file_path.parent_path());
    }
```

AssemblyJohn:
Fix here: https://github.com/EVerest/libocpp/pull/659/files

PR #650, opened 2024-05-30
Replace recv_loop with the new callback handler since they do the same; calling callbacks from a different thread

marcemmers:
## Describe your changes
The recv_loop is used to handle messages from a different thread than the one transmitting them. Sometimes when we receive a message we transmit a response and wait on the result. If we are calling that from the same thread we've created a deadlock.

Recently I introduced a new thread for handling callbacks. The recv_loop is essentially also calling a callback so could make use of the same thread instead of having its own.


marcemmers:
I will close this PR for now. @AssemblyJohn rightfully pointed out that the queue has to be cleared on a disconnect and that would potentially require more logic and complexity than I am trying to remove.

PR #652, opened 2024-05-31
Plug&Charge authorization in case OCSP can not be generated

Pietfried:
## Describe your changes
Without this change, a Plug&Charge authorization request failed when the OCSP could not be generated, e.g. because the contract certificate chain did not contain any OCSP extension. Those contract certificates exist in the real world. This commit changes the behavior in case no OCSP data could be generated. If CentralContractValidationAllowed is true, the certificate is send as part of the Authorize.req instead of rejecting the request directly.

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -2970,7 +2970,15 @@ ocpp::v201::AuthorizeResponse ChargePointImpl::data_transfer_pnc_authorize(
                     forward_to_csms = true;
                 } else {
                     EVLOG_warning << "Online: OCSP data could not be generated";
-                    authorize_response.idTokenInfo.status = ocpp::v201::AuthorizationStatusEnum::Invalid;
+                    if (central_contract_validation_allowed) {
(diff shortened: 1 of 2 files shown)


lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -2970,7 +2970,15 @@ ocpp::v201::AuthorizeResponse ChargePointImpl::data_transfer_pnc_authorize(
                     forward_to_csms = true;
                 } else {
                     EVLOG_warning << "Online: OCSP data could not be generated";
-                    authorize_response.idTokenInfo.status = ocpp::v201::AuthorizationStatusEnum::Invalid;
+                    if (central_contract_validation_allowed) {
+                        EVLOG_info << "Online: OCSP data could not be generated. Pass contract validation to CSMS";
+                        authorize_req.certificate = certificate.value();
+                        forward_to_csms = true;
```
mennodegraaf, line 2976: In the OCPP2.0.1 implementation this is called `forwarded_to_csms`

mennodegraaf (approved):
The changes look very straightforward

hikinggrass (approved):

PR #664, opened 2024-06-07
Add device model interface and use that for testing

marcemmers:

### DIFF
--- a/include/ocpp/v201/device_model.hpp
+++ b/include/ocpp/v201/device_model.hpp
@@ -128,8 +128,8 @@ public:
     ///                        otherwise only non read-only variables can be changed. Defaults to false
     /// \return Result of the requested operation
     virtual SetVariableStatusEnum set_value(const Component& component_id, const Variable& variable_id,
-                                    const AttributeEnum& attribute_enum, const std::string& value,
-                                    const bool allow_read_only = false) = 0;
+                                            const AttributeEnum& attribute_enum, const std::string& value,
+                                            const bool allow_read_only = false) = 0;
 
     /// \brief Sets the variable_id attribute \p value specified by \p component_id , \p variable_id and \p
     /// attribute_enum for read only variables only. Only works on certain allowed components.
@@ -139,13 +139,15 @@ public:
     /// \param value
     /// \return Result of the requested operation
     virtual SetVariableStatusEnum set_read_only_value(const Component& component_id, const Variable& variable_id,
-                                              const AttributeEnum& attribute_enum, const std::string& value) = 0;
+                                                      const AttributeEnum& attribute_enum,
+                                                      const std::string& value) = 0;
 
     /// \brief Gets the VariableMetaData for the given \p component_id and \p variable_id
     /// \param component_id
     /// \param variable_id
     /// \return VariableMetaData or std::nullopt if \p component_id or \p variable_id not present
-    virtual std::optional<VariableMetaData> get_variable_meta_data(const Component& component_id, const Variable& variable_id) = 0;
+    virtual std::optional<VariableMetaData> get_variable_meta_data(const Component& component_id,
+                                                                   const Variable& variable_id) = 0;
 
     /// \brief Gets the ReportData for the specifed filter \p report_base \p component_variables and \p
     /// component_criteria
@@ -161,9 +163,9 @@ public:
     /// \param component_variables
     /// \param component_criteria
     /// \return
-    virtual std::vector<ReportData>
-    get_custom_report_data(const std::optional<std::vector<ComponentVariable>>& component_variables = std::nullopt,
-                           const std::optional<std::vector<ComponentCriterionEnum>>& component_criteria = std::nullopt) = 0;
+    virtual std::vector<ReportData> get_custom_report_data(
+        const std::optional<std::vector<ComponentVariable>>& component_variables = std::nullopt,
+        const std::optional<std::vector<ComponentCriterionEnum>>& component_criteria = std::nullopt) = 0;
 
     /// \brief Check data integrity of the device model provided by the device model data storage:
     /// For "required" variables, assert values exist. Checks might be extended in the future.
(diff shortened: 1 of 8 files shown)


hikinggrass (approved):

include/ocpp/v201/device_model.hpp
```diff
@@ -168,46 +127,118 @@ class DeviceModel {
...
     /// \param report_base
     /// \param component_variables
     /// \param component_criteria
     /// \return
-    std::vector<ReportData>
-    get_custom_report_data(const std::optional<std::vector<ComponentVariable>>& component_variables = std::nullopt,
-                           const std::optional<std::vector<ComponentCriterionEnum>>& component_criteria = std::nullopt);
+    virtual std::vector<ReportData> get_custom_report_data(
+        const std::optional<std::vector<ComponentVariable>>& component_variables = std::nullopt,
+        const std::optional<std::vector<ComponentCriterionEnum>>& component_criteria = std::nullopt) = 0;
```
a-w50, line 168: Is it really necessary to transfer an optional `std::vector` instead of just passing an empty vector?

include/ocpp/v201/device_model.hpp
```diff
@@ -168,46 +127,118 @@ class DeviceModel {
...
+    /// a value for the given parameters can be requested. If it can be requested it will be retrieved from the device
+    /// model storage and the given \p value will be set to the value that was retrieved
+    /// \param component_id
+    /// \param variable_id
+    /// \param attribute_enum
+    /// \param value string reference to value: will be set to requested value if value is present
+    /// \param allow_write_only true to allow a writeOnly value to be read.
+    /// \return GetVariableStatusEnum that indicates the result of the request
+    virtual GetVariableStatusEnum request_value_internal(const Component& component_id, const Variable& variable_id,
+                                                         const AttributeEnum& attribute_enum, std::string& value,
```
a-w50, line 185: Would return the `value` by value, instead of passing by value.

include/ocpp/v201/device_model.hpp
```diff
@@ -168,46 +127,118 @@ class DeviceModel {
...
+/// functionality to support the use cases defined in the functional block Provisioning
+class DeviceModel : public DeviceModelInterface {
+
+private:
+    DeviceModelMap device_model;
+    std::unique_ptr<DeviceModelStorage> storage;
+
+    GetVariableStatusEnum request_value_internal(const Component& component_id, const Variable& variable_id,
+                                                 const AttributeEnum& attribute_enum, std::string& value,
+                                                 bool allow_write_only) override;
```
a-w50, line 199: Depending on whether this `DeviceModel` class should be derived from again or not, it might be better to use `final` instead of `override` here and following.
Furthermore, the `DeviceModel` implementation could be moved into a separate header file if implementation details should not leak into the `DeviceModelInterface`

include/ocpp/v201/smart_charging.hpp
```diff
@@ -47,15 +47,15 @@ std::ostream& operator<<(std::ostream& os, const ProfileValidationResultEnum val
...
+    std::map<int32_t, std::shared_ptr<EvseInterface>>& evses;
 
     std::shared_ptr<ocpp::v201::DatabaseHandler> database_handler;
     // cppcheck-suppress unusedStructMember
     std::map<int32_t, std::vector<ChargingProfile>> charging_profiles;
     std::vector<ChargingProfile> station_wide_charging_profiles;
 
 public:
-    explicit SmartChargingHandler(std::map<int32_t, std::unique_ptr<EvseInterface>>& evses);
+    explicit SmartChargingHandler(std::map<int32_t, std::shared_ptr<EvseInterface>>& evses);
```
a-w50, line 58: Is there any reason why `explicit` is used here?

include/ocpp/v201/smart_charging.hpp
```diff
@@ -47,15 +47,15 @@ std::ostream& operator<<(std::ostream& os, const ProfileValidationResultEnum val
 /// to calculate the composite schedules
 class SmartChargingHandler {
 private:
-    std::map<int32_t, std::unique_ptr<EvseInterface>>& evses;
+    std::map<int32_t, std::shared_ptr<EvseInterface>>& evses;
```
a-w50, line 50: Creating a type alias for this type here like `using EvseMap = std::map<int32_t, std::unique_ptr<EvseInterface>>` and using it might make the code more readable.

a-w50: Is there any reason, why changing from `std::unique_ptr` to `std::shared_ptr`?

a-w50 (commented):
Just some notes/questions

include/ocpp/v201/device_model.hpp
marcemmers: I did not really touch or change the functions, I only made an interface for it. So while the comment might be valid I think it is out of scope for this PR so I will not look into it. I will resolve this conversation and also some others which are similar.

include/ocpp/v201/device_model.hpp
marcemmers: Currently I don't think there is a reason to derive from `DeviceModel` again but I also don't see a reason to prohibit it by using final.

As for the interface. The only reason we are doing that is to be able to mock it. Not really to not leak implementation detail so that is why we have been using the interfaces in the same file as the header.

include/ocpp/v201/smart_charging.hpp
marcemmers: I have another change in progress that gets rid of the `map<int, unique_ptr<EvseInterface>>` thing by using a different class to manage the Evse since I think this map exposes way more than needed. So I don't see the need to change the map right now. It is a good suggestion otherwise.

As for the shared ptr, it's to make mocking easier since the test driver can access the `EvseMock` as well as inject it into the `SmartChargingHandler`. This will probably also change with the above mentioned PR.

include/ocpp/v201/device_model.hpp
a-w50: I see your point.  Still I think, if creating a now an abstract interface for the device model class, it would make sense to rethink some of its interface.  Doing this later on would be at least 3 times the work (abstract interface, implementation and mock).

include/ocpp/v201/smart_charging.hpp
a-w50: I had a similar issue, but usually then the unit test is not written properly (I guess it is about the lifetime right? The unit test just creates evse's and forgets about them - they should be part of the unit tests context - you might want to do assertions on them later on anyway and pass them via `std::ref` do the map).  In general I wouldn't make any compromises on the ownership of objects (shared vs. unique) in the code, just to make it easier to write unit tests.
But depending on the PR you're mentioning, this problem might be gone by then.

marcemmers: For reference, the PR i was talking about is here: https://github.com/EVerest/libocpp/pull/668

I will update this one after that one is merged.

maaikez: Well as you know, I really don't like changing code just for unittests. It makes things more complex to understand and read in my opinion.

PR #539, opened 2024-03-24
[REF] Documentation changed for the wrong name of parameter (issue 422)

thenaserov:
status_notifcation_callback was replaced by component_state_manager to have a correct documentation. Simply the name in the documentation modified! Let me know if there is something missing or misunderstood otherwise kindly provide assistance for successful merge.

## Checklist before requesting a review
- [*] I have performed a self-review of my code
- [*] I have made corresponding changes to the documentation
- [*] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

PR #668, opened 2024-06-14
Add helper class holding the EVSE's

marcemmers:
## Describe your changes

I've noticed the pattern of sharing a references to maps holding unique_ptrs and sharing references to unique_ptrs themselves and though we are exposing to much in that way. The map can be changed most of the time and the ptrs influenced as well.

This change adds a new class that holds the EVSE classes and only exposes them in a way that they are not mutable. This class is then much more easy to pass along to other users of the EVSEs.

I also added a specific exception when trying to access an EVSE with an invalid evse_id.

### DIFF
--- a/include/ocpp/common/custom_iterators.hpp
+++ b/include/ocpp/common/custom_iterators.hpp
@@ -8,28 +8,43 @@
 
 namespace ocpp {
 
+// This file contains the implementation of a couple of custom iterators, enabling the iteration and abstraction of more complex containers.
+// For example we want to be able to iterate a std::vector<std::unique_ptr<T>> and get references to T from the iterator
+//
+// ForwardIterator<T> implements a custom iterator for type T that does not depend on any underlying container.
+//   You can add this for your type by returning ForwardIterator<T> from begin() and end()
+// ForwardIterator can only be constructed by passing a ForwardIteratorBase derived struct in a unique_ptr.
+//   ForwardIteratorBase is an abstract class that allows us to implement container specific functionality.
+// VectorOfUniquePtrIterator is an implementation of the ForwardIteratorBase specifically catered to the example of std::vector<std:unique_ptr<T>>
+//   So to implement the begin() and end() functions for your type, you could do the following:
+//   std::vector<std::unique_ptr<T>> container;
+//   ForwardIterator<T> begin() {
+//       return ForwardIterator<T>(std::make_unique<VectorOfUniquePtrIterator<T>>(container.begin()));
+//   }
+
+/// \brief Abstract struct to implement to enable the use of the ForwardIterator<T>
+template <typename T> struct ForwardIteratorBase {
+    virtual ~ForwardIteratorBase() = default;
+
+    /// \brief Get a reference to the value pointed to by the iterator
+    virtual T& deref() const = 0;
+    /// \brief Increment the iterator once to the next position
+    virtual void advance() = 0;
+    /// \brief Check for equality between this iterator and \p other
+    virtual bool not_equal(const ForwardIteratorBase& other) = 0;
+};
+
 /// \brief Helper struct that allows the of an iterator in an interface, can be implemented using any forward iterator
-template <typename T> struct ForwardIteratorWrapper {
+template <typename T> struct ForwardIterator {
     using iterator_category = std::forward_iterator_tag;
     using difference_type = std::ptrdiff_t;
     using value_type = T;
     using pointer = value_type*;
     using reference = value_type&;
-
-    /// \brief Abstract struct to implement to enable the use of this iterator wrapper
-    struct Base {
-        virtual ~Base() = default;
-
-        /// \brief Get a reference to the value pointed to by the iterator
-        virtual reference deref() const = 0;
-        /// \brief Increment the iterator once to the next position
-        virtual void advance() = 0;
-        /// \brief Check for equality between this iterator and \p other
-        virtual bool not_equal(const Base& other) = 0;
-    };
+    using base = ForwardIteratorBase<T>;
 
     /// \brief Construct a new wrapper using any type that implements the abstract struct Base
-    explicit ForwardIteratorWrapper(std::unique_ptr<Base> it) : it{std::move(it)} {
+    explicit ForwardIterator(std::unique_ptr<base> it) : it{std::move(it)} {
     }
 
     /// \brief Get a reference to the value pointed to by the iterator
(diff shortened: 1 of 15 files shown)


tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -206,29 +183,26 @@ class ChargepointTestFixtureV201 : public testing::Test {
...
     EXPECT_THAT(sut, testing::Eq(ProfileValidationResultEnum::TxProfileMissingTransactionId));
 }
 
-TEST_F(ChargepointTestFixtureV201, K01FR16_IfTxProfileHasEvseIdNotGreaterThanZero_ThenProfileIsInvalid) {
-    auto wrong_evse_id = STATION_WIDE_ID;
-    create_evse_with_id(wrong_evse_id);
-    auto profile = create_charging_profile(DEFAULT_PROFILE_ID, ChargingProfilePurposeEnum::TxProfile,
-                                           create_charge_schedule(ChargingRateUnitEnum::A), uuid());
-    auto sut = handler.validate_tx_profile(profile, *evses[wrong_evse_id]);
+// TODO: Review this test. TxProfileEvseIdNotGreaterThanZero does not make sense since you can't get an EvseInterface& with get_id()==0
```
marcemmers, line 192: @christopher-davis-afs @couryrr-afs This is the PR I was talking about during the WG meeting. 

I had to disable this test since I could not find a logical way to fix it. Do you have any insights here?

Also comments in general are appreciated.

include/ocpp/v201/evse.hpp
```diff
@@ -170,13 +168,13 @@ class Evse : public EvseInterface {
     Evse(const int32_t evse_id, const int32_t number_of_connectors, DeviceModel& device_model,
          std::shared_ptr<DatabaseHandler> database_handler,
          std::shared_ptr<ComponentStateManagerInterface> component_state_manager,
-         const std::function<void(const MeterValue& meter_value, const Transaction& transaction, const int32_t seq_no,
-                                  const std::optional<int32_t> reservation_id)>& transaction_meter_value_req,
-         const std::function<void()> pause_charging_callback);
+         const std::function<void(const MeterValue& meter_value, EnhancedTransaction& transaction)>&
```
marcemmers, line 171: Fix the doxygen of the constructor so https://github.com/EVerest/libocpp/pull/539 can be closed

include/ocpp/v201/evse.hpp
```diff
@@ -30,11 +30,11 @@ class EvseInterface {
 
     /// \brief Returns an OCPP2.0.1 EVSE type
```
maaikez, line 31: Comment is not changed, but function is.

include/ocpp/v201/evse_manager.hpp
```diff
@@ -0,0 +1,68 @@
...
+    std::string msg;
+};
+
+class EvseManagerInterface {
+private:
+    virtual std::vector<std::unique_ptr<EvseInterface>>& get_vector() = 0;
+    virtual const std::vector<std::unique_ptr<EvseInterface>>& get_vector() const = 0;
+
+public:
+    using EvseIterator = VectorOfUniquePtrIterator<EvseInterface>;
```
maaikez, line 33: Why is a const vector with const unique ptrs not possible?
`const std::vector<const std::unique_ptr<const EvseInterface>>` or something?

include/ocpp/v201/evse_manager.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+    virtual ~EvseManagerInterface() = default;
+
+    EvseIterator begin();
```
maaikez, line 37: I am missing documentation (not only here, but for the whole EvseIterator), why you use this and not just a vector. If someone looks at this later, it is very unclear why this is for (and I do not totally understand it either as well, is it really a problem what you are trying to solve here?).

marcemmers: I agree this all needs a bit of documentation so I will add that.

The reason I have done it like this is so that we no longer expose the `unique_ptr` that we use just for storage to the "outside" world.

Initially I was also afraid that if you return the vector others could reset or assign the `unique_ptr`s for example. Returning a `const vector` will take care of that, but you still have to deal with pointers instead of references which I think is not ideal.

Also, now you can use range based for on the `evse_manager` in `charge_point`, otherwise you would have to call `get_vector` or a similar function.

lib/ocpp/v201/evse_manager.cpp
```diff
@@ -0,0 +1,51 @@
...
+                             transaction_meter_value_req,
+                         const std::function<void(int32_t evse_id)>& pause_charging_callback) {
+    evses.reserve(evse_connector_structure.size());
+    for (const auto& [evse_id, connectors] : evse_connector_structure) {
+        evses.push_back(std::make_unique<Evse>(evse_id, connectors, device_model, database_handler,
+                                               component_state_manager, transaction_meter_value_req,
+                                               pause_charging_callback));
+    }
+}
+
```
a-w50, line 22: Dear @marcemmers ,

in general I think it is very fruitful to create some abstract interfaces for the main classes, because testability might be increased and interfaces get more explicit.
Looking at the `EvseManagerInterface`, it is (except for the constructor) a container wrapper.  But despite the `begin/end` functions one could have simply taken a `std::vector`.  To keep things simple, I would also propose to only use a `std::vector` - in case you would like to get some custom iterator behavior, you can still overload `std::begin` and `std::end` to return your custom iterator (as far as I remember this is also the preferred way of customizing containers, as it doesn't couple the logic of how to have a view (iterators) on the container to the container itself).
Note, that this won't work for range-based for loops, because `std::vector` has implemented it's own member functions `begin` and `end` and hence these will be selected while constructing the for loop (see https://en.cppreference.com/w/cpp/language/range-for) - btw. that is exactly the reason why one should prefer the free functions `begin` and `end` vs. member functions.  You could still write a wrapper like `EvseByReference` only for getting your custom iterator and then you could iterate via `for(auto& evse : EvseByReference(evse_map))` ...
An issue, I see with the `EvseManagerInterface` is that it couples the construction of the evse vector to the container itself because this is done in the constructor.  So whenever you pass this container to some other function, you will leak all the construction details.  It might be better here to have a free function, which does exactly the same as the constructor but instead returns the vector.  Furthermore, the `EvseManagerInterface` restricts your flexibility, as you're not able to add/remove evses dynamically (although the `EvseManager` will allow for that by using `get_vector`).
The `is_valid_evse` function for example could also be freed by just accepting the vector and the id and then comparing sizes.  This way, you're decoupling again the responsibility of containing/owning to one class (vector in this case) and the responsibility of creating elements or arguing whether the elements are in a valid state or whatever to another class.  This increases both composability and testability.

Furthermore I see excessive use of `std::shared_ptr` - this is a general issue within libocpp - but as far as I can see, this is not necessary, because the lifetime/ownership relationships are quite clear.  In this case for example, the charge point owns the evses and hence can pass by reference (instead of shared pointer) to the smart charging handler.  Shared pointers would be only necessary if there are valid cases, where the smart charging handler could outlive the charge point - which I think can be ruled out as a valid use case.  The same reasoning applies for most of the other shared pointers.  Although, performance might not be a big issue here, keep in mind that shared pointers need to use locking when constructed/destructed.

marcemmers: Hi @a-w50, thanks for the extensive feedback!

I can see what you mean about me increasing the coupling by forcing the use of vectors for whoever implements the `EvseManagerInterface`. Initially I had it split into a pure interface and then a derived implementation, after that changed it into this state but I agree I shouldn't have done that.

I still think this interface class is better then using a vector or a map or whatever container since my whole point is that I don't think we should allow adding more evse's dynamically. This is a config that should be applied once at startup and it does not make sense to change it on the fly. Allowing this under the guise of flexibility opens up potential other issues because not all the initialization code might be performed for example. Whatever is used by the derived class should not matter, that is implementation detail and should not leak.

As for the shared ptr, I agree and I will change it to share a reference to the interface instead of a shared_ptr.

include/ocpp/common/custom_iterators.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    /// \brief Get a reference to the value pointed to by the iterator
+    virtual T& deref() const = 0;
+    /// \brief Increment the iterator once to the next position
+    virtual void advance() = 0;
+    /// \brief Check for equality between this iterator and \p other
+    virtual bool not_equal(const ForwardIteratorBase& other) = 0;
+};
+
+/// \brief Helper struct that allows the of an iterator in an interface, can be implemented using any forward iterator
```
hikinggrass, line 37: looks like there's something missing between "the" and "of"

hikinggrass (approved):
Just a minor additional comment

PR #670, opened 2024-06-14
Add option to disable compilation of v16 or v201 if you have no need for one of those 2 options

marcemmers:
## Describe your changes

Instead of adding a bunch of layers to the CMake libraries I've used options instead since they are more easily integrated.
Added options `LIBOCPP_ENABLE_V16` and `LIBOCPP_ENABLE_V201` which default to `ON`. Set to `OFF` to disable compilation of one of the parts.

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [ ] If OCPP 2.0.1: I have updated the [OCPP 2.0.1 status document](https://github.com/EVerest/libocpp/tree/main/doc/ocpp_201_status.md)
- [x] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

PR #674, opened 2024-06-15
Unit test refactoring

Pietfried:
## Describe your changes
Refactored unit tests to allow building common + either v16 or v201 unit tests if only one of the is enabled for compilation:
* Moved all unit tests into either common/ v16/ or v201/ directory
* Now we have a single test target `libocpp_unit_tests` instead of multiple ones
* separated message_queue unit tests between v16 and v201
* only building unit tests for v16 or v201 if LIBOCPP_ENABLE_VXX is set

This PR is build upon https://github.com/EVerest/libocpp/pull/670


maaikez:
Maybe not for this pull request but for development it is easier to have an executable test for each test class. Because now you can run all tests or one (at least in my IDE), but not one test class. So I would like to have them all separate, if possible. I know you can run a test class from the terminal quite easily but that's not as nice as having a test executable for each test. But the opinions might differ on this of course.

Other than that, looks ok!

Pietfried:
Since the tests are currently not compiling if you only build v16 or v201 using `-DLIBOCPPP_ENABLE_V16` or `-DLIBOCPPP_ENABLE_V201` this PR only fixes the compilation, restructures the test directories resulting in a single test executable. We can split this up into v16, v201 and common (or even more) test executables in a subsequent PR. This one is now ready for review.

tests/CMakeLists.txt
```diff
@@ -2,30 +2,8 @@ set(MIGRATION_FILES_LOCATION_V16 "${CMAKE_CURRENT_BINARY_DIR}/resources/v16/migr
...
-    GTest::gtest_main
-    pthread
-    )
-
-
-add_test(database_tests database_tests)
-
 add_executable(libocpp_unit_tests)
 
-target_compile_definitions(database_tests
```
maaikez, line 21: The libocpp_unit_tests still pass without those compile definitions? Why were they needed anyway then?

maaikez (approved):

hikinggrass (approved):

PR #682, opened 2024-06-21
Smart Charging: Expand add profile capabilities

couryrr-afs:
## Describe your changes

- Added logic to replace existing non-external constraint charging profiles
- Added logic to verify TX Default charging profile evse placement.
- Added test to cover required callbacks.

### DIFF
--- a/doc/ocpp_201_status.md
+++ b/doc/ocpp_201_status.md
@@ -1234,7 +1234,7 @@ This document contains the status of which OCPP 2.0.1 numbered functional requir
 | K01.FR.02 | 🌐     |                                                                                                 |
 | K01.FR.03 | 🌐 💂  | `TxProfile`s without `transactionId`s are rejected.                                             |
 | K01.FR.04 | ✅     |                                                                                                 |
-| K01.FR.05 |        |                                                                                                 |
+| K01.FR.05 | ✅     |                                                                                                 |
 | K01.FR.06 | 🌐     |                                                                                                 |
 | K01.FR.07 | ⛽️     | Notified through the `signal_set_charging_profiles` callback.                                   |
 | K01.FR.08 | 🌐     | `TxDefaultProfile`s are supported.                                                              |
@@ -1243,8 +1243,8 @@ This document contains the status of which OCPP 2.0.1 numbered functional requir
 | K01.FR.11 |        |                                                                                                 |
 | K01.FR.12 |        |                                                                                                 |
 | K01.FR.13 |        |                                                                                                 |
-| K01.FR.14 |        |                                                                                                 |
-| K01.FR.15 |        |                                                                                                 |
+| K01.FR.14 | ✅     |                                                                                                 |
+| K01.FR.15 | ✅     |                                                                                                 |
 | K01.FR.16 |        |                                                                                                 |
 | K01.FR.17 |        |                                                                                                 |
 | K01.FR.19 |        |                                                                                                 |
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -38,6 +38,7 @@ bool Callbacks::all_callbacks_valid() const {
            this->remote_start_transaction_callback != nullptr and this->is_reservation_for_token_callback != nullptr and
            this->update_firmware_request_callback != nullptr and
            this->security_event_callback != nullptr and
+           this->set_charging_profiles_callback != nullptr and
            (!this->variable_changed_callback.has_value() or this->variable_changed_callback.value() != nullptr) and
            (!this->validate_network_profile_callback.has_value() or
             this->validate_network_profile_callback.value() != nullptr) and
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -37,6 +37,7 @@ bool Callbacks::all_callbacks_valid() const {
            this->get_log_request_callback != nullptr and this->unlock_connector_callback != nullptr and
            this->remote_start_transaction_callback != nullptr and this->is_reservation_for_token_callback != nullptr and
            this->update_firmware_request_callback != nullptr and
+           this->security_event_callback != nullptr and
            (!this->variable_changed_callback.has_value() or this->variable_changed_callback.value() != nullptr) and
            (!this->validate_network_profile_callback.has_value() or
             this->validate_network_profile_callback.value() != nullptr) and
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -36,8 +36,7 @@ bool Callbacks::all_callbacks_valid() const {
            this->connector_effective_operative_status_changed_callback != nullptr and
            this->get_log_request_callback != nullptr and this->unlock_connector_callback != nullptr and
            this->remote_start_transaction_callback != nullptr and this->is_reservation_for_token_callback != nullptr and
-           this->update_firmware_request_callback != nullptr and
-           this->security_event_callback != nullptr and
+           this->update_firmware_request_callback != nullptr and this->security_event_callback != nullptr and
            this->set_charging_profiles_callback != nullptr and
            (!this->variable_changed_callback.has_value() or this->variable_changed_callback.value() != nullptr) and
            (!this->validate_network_profile_callback.has_value() or
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -45,7 +45,21 @@ bool Callbacks::all_callbacks_valid() const {
             this->configure_network_connection_profile_callback.value() != nullptr) and
            (!this->time_sync_callback.has_value() or this->time_sync_callback.value() != nullptr) and
            (!this->boot_notification_callback.has_value() or this->boot_notification_callback.value() != nullptr) and
-           (!this->ocpp_messages_callback.has_value() or this->ocpp_messages_callback.value() != nullptr);
+           (!this->ocpp_messages_callback.has_value() or this->ocpp_messages_callback.value() != nullptr) and
+           (!this->cs_effective_operative_status_changed_callback.has_value() or
+            this->cs_effective_operative_status_changed_callback.value() != nullptr) and
+           (!this->evse_effective_operative_status_changed_callback.has_value() or
+            this->evse_effective_operative_status_changed_callback.value() != nullptr) and
+           (!this->get_customer_information_callback.has_value() or
+            this->get_customer_information_callback.value() != nullptr) and
+           (!this->clear_customer_information_callback.has_value() or
+            this->clear_customer_information_callback.value() != nullptr) and
+           (!this->all_connectors_unavailable_callback.has_value() or
+            this->all_connectors_unavailable_callback.value() != nullptr) and
+           (!this->data_transfer_callback.has_value() or this->data_transfer_callback.value() != nullptr) and
+           (!this->transaction_event_callback.has_value() or this->transaction_event_callback.value() != nullptr) and
+           (!this->transaction_event_response_callback.has_value() or
+            this->transaction_event_response_callback.value() != nullptr);
 }
 
 ChargePoint::ChargePoint(const std::map<int32_t, int32_t>& evse_connector_structure,
--- a/lib/ocpp/v201/smart_charging.cpp
+++ b/lib/ocpp/v201/smart_charging.cpp
@@ -325,6 +325,7 @@ SetChargingProfileResponse SmartChargingHandler::add_profile(int32_t evse_id, Ch
         std::find_if(profile_storage.begin(), profile_storage.end(),
                      [&profile](const ChargingProfile existing_profile) { return profile.id == existing_profile.id; });
 
+    // K01.FR05 - replace non-ChargingStationExternalConstraints profiles if id exists.
     if (profile_with_id != profile_storage.end() &&
         profile_with_id->chargingProfilePurpose != ChargingProfilePurposeEnum::ChargingStationExternalConstraints) {
         *profile_with_id = profile;
--- a/lib/ocpp/v201/smart_charging.cpp
+++ b/lib/ocpp/v201/smart_charging.cpp
@@ -12,10 +12,13 @@
 #include "ocpp/v201/ocpp_types.hpp"
 #include "ocpp/v201/transaction.hpp"
 #include <algorithm>
+#include <exception>
 #include <iterator>
 #include <memory>
 #include <ocpp/v201/smart_charging.hpp>
 #include <optional>
+#include <stdexcept>
+#include <vector>
 
 using namespace std::chrono;
 
@@ -28,6 +31,8 @@ std::string profile_validation_result_to_string(ProfileValidationResultEnum e) {
         return "Valid";
     case ProfileValidationResultEnum::EvseDoesNotExist:
         return "EvseDoesNotExist";
+    case ProfileValidationResultEnum::ExistingChargingStationExternalConstraints:
+        return "ExstingChargingStationExternalConstraints";
     case ProfileValidationResultEnum::InvalidProfileType:
         return "InvalidProfileType";
     case ProfileValidationResultEnum::TxProfileMissingTransactionId:
@@ -111,6 +116,11 @@ ProfileValidationResultEnum SmartChargingHandler::validate_profile(ChargingProfi
         if (result != ProfileValidationResultEnum::Valid) {
             return result;
         }
+    }
+
+    result = verify_no_conflicting_external_constraints_id(profile);
+    if (result != ProfileValidationResultEnum::Valid) {
+        return result;
     }
 
     if (evse_id != STATION_WIDE_ID) {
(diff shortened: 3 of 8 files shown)


couryrr-afs:
@shankari this is ready for your review.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -315,12 +316,31 @@ SmartChargingHandler::validate_profile_schedules(ChargingProfile& profile,
...
+        std::find_if(profile_storage.begin(), profile_storage.end(),
+                     [&profile](const ChargingProfile existing_profile) { return profile.id == existing_profile.id; });
+
+    if (profile_with_id != profile_storage.end() &&
+        profile_with_id->chargingProfilePurpose != ChargingProfilePurposeEnum::ChargingStationExternalConstraints) {
+        *profile_with_id = profile;
     } else {
-        charging_profiles[evse_id].push_back(profile);
+        profile_storage.push_back(profile);
+    }
```
louisg1337, line 333: Just for documentation purposes, I think leaving a comment that this relates to K01.FR.05 could be helpful.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -950,4 +956,75 @@ TEST_F(ChargepointTestFixtureV201, K01_ValidateProfile_IfChargeStationMaxProfile
...
+    auto sut2 = handler.add_profile(DEFAULT_EVSE_ID, profile2);
+
+    auto profiles = handler.get_profiles();
+
+    EXPECT_THAT(sut1.status, testing::Eq(ChargingProfileStatusEnum::Accepted));
+    EXPECT_THAT(sut2.status, testing::Eq(ChargingProfileStatusEnum::Accepted));
+    EXPECT_THAT(profiles, testing::Contains(profile1));
+    EXPECT_THAT(profiles, testing::Contains(profile2));
+}
+
```
louisg1337, line 1029: Not sure if it is necessary, but to be thorough to the K01FR05 tests, it could be worth adding in the opposites of these tests:

1. Same id does **NOT** exist and `ChargingStationExternalContraints`
2. Same id does **NOT** exist and **NO** `ChargingStationExternalContraints`

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -950,4 +956,75 @@ TEST_F(ChargepointTestFixtureV201, K01_ValidateProfile_IfChargeStationMaxProfile
...
+    create_evse_with_id(DEFAULT_EVSE_ID);
+    auto profile = create_charging_profile(DEFAULT_PROFILE_ID, ChargingProfilePurposeEnum::TxDefaultProfile,
+                                           create_charge_schedule(ChargingRateUnitEnum::A), uuid(),
+                                           ChargingProfileKindEnum::Absolute, DEFAULT_STACK_LEVEL);
+
+    auto sut = handler.add_profile(DEFAULT_EVSE_ID, profile);
+
+    EXPECT_THAT(sut.status, testing::Eq(ChargingProfileStatusEnum::Accepted));
+    EXPECT_THAT(handler.get_profiles(), testing::Contains(profile));
+}
```
louisg1337, line 983: Another test here could potentially be helpful:

1. `TxDefaultProfile` and same stack level **DOES** exist

couryrr-afs: This test is unique in the sense that we are only testing that add is placing the profile in the correct location. The flow would normally have the charging profiles going through a validation step before hand which would catch this sort of incorrect state. If you take a look in the tests for `DuplicateTxDefaultProfileFound` that is where this is validated.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
couryrr-afs: For case 1 I believe that is covered with the normal `add_profile` test. Keeping in mind that this is only the ability to add profiles. It does not cover the validation of profiles.

This test is specific to the case of `K01FR05` with that being the lens to look at this through case 2 would not help in catching regressions to `ChargingStationExternalContraints`.

tests/lib/ocpp/v201/test_charge_point.cpp
```diff
@@ -0,0 +1,321 @@
...
+    callbacks.security_event_callback = nullptr;
+
+    EXPECT_FALSE(callbacks.all_callbacks_valid());
+}
+
+TEST_F(ChargePointFixture, K01FR02_CallbacksValidityChecksIfSetChargingProfilesCallbackExists) {
+    configure_callbacks_with_mocks();
+    callbacks.set_charging_profiles_callback = nullptr;
+
+    EXPECT_FALSE(callbacks.all_callbacks_valid());
```
shankari, line 145: is the callback validity actually a test case for `K01FR02`? Why? It doesn't seem to be related to the requirements in the spec.

shankari (approved):
This is a fairly simple change and my comments are primarily focused on improved documentation, so I am approving this. Please do address my comments before sending it off to the community for merging. Documentation clarity is just as important, particularly in open source projects, where contributors can join and leave, as code.

tests/lib/ocpp/v201/test_charge_point.cpp
drmrd: @couryrr-afs: I agree with @shankari that this could use some context. Could you move the `all_callbacks_valid` tests that involve `set_charging_profiles_callback` to the top of these tests (if they're not there already) and include the following comment:

```c++
/*
 * K01.FR.02 states
 *
 *     "The CSMS MAY send a new charging profile for the EVSE that SHALL be used
 *      as a limit schedule for the EV."
 *
 * When using libocpp, a charging station is notified of a new charging profile
 * by means of the set_charging_profiles_callback. In order to ensure that a new
 * profile can be immediately "used as a limit schedule for the EV", a
 * valid set_charging_profiles_callback must be provided.
 *
 * As part of testing that K01.FR.02 is met, we provide the following tests that
 * confirm an OCPP 2.0.1 ChargePoint with smart charging enabled will only
 * consider its collection of callbacks valid if set_charging_profiles_callback
 * is provided.
 */
```

Following those tests, we can add a comment like the following:
```c++
/*
 * For completeness, we also test that all other callbacks are checked by
 * validate_all_profiles.
 */
```

The remaining `validate_all_profiles` tests could immediately follow this comment but with references to K01.FR.02 removed from their names.

drmrd: I could see moving the other tests to a different location in the future but think it's fine to have them here for now. If other validity tests are relevant for functional requirements in the other functional blocks (as is highly likely), they can be moved around as needed. For now, though, it's nice to keep them grouped together.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -33,15 +33,30 @@ bool Callbacks::all_callbacks_valid() const {
            this->connector_effective_operative_status_changed_callback != nullptr and
            this->get_log_request_callback != nullptr and this->unlock_connector_callback != nullptr and
            this->remote_start_transaction_callback != nullptr and this->is_reservation_for_token_callback != nullptr and
-           this->update_firmware_request_callback != nullptr and
+           this->update_firmware_request_callback != nullptr and this->security_event_callback != nullptr and
+           this->set_charging_profiles_callback != nullptr and
```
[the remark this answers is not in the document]
drmrd: @shankari: The `security_event_callback` is defined as non-optional in `charge_point.hpp`:

https://github.com/EVerest/libocpp/blob/77f5b2a6181d94c2e94af790e4c5e61435f0b919/include/ocpp/v201/charge_point.hpp#L146-L152

It looks to me like @gberardi-pillar's change is meant to correct for this.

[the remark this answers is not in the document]
drmrd: That said, the only usage of `security_event_callback` in `libocpp` appears to assume that it's optional:

https://github.com/EVerest/libocpp/blame/587d491589f944c8a99ce8fe56ece612e061f385/lib/ocpp/v201/charge_point.cpp#L1878-L1880

So perhaps it would be better to change the type of `security_event_callback` to `std::optional<std::function>` instead of add it to the validate method?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -33,15 +33,30 @@ bool Callbacks::all_callbacks_valid() const {
...
+           (!this->get_customer_information_callback.has_value() or
+            this->get_customer_information_callback.value() != nullptr) and
+           (!this->clear_customer_information_callback.has_value() or
+            this->clear_customer_information_callback.value() != nullptr) and
+           (!this->all_connectors_unavailable_callback.has_value() or
+            this->all_connectors_unavailable_callback.value() != nullptr) and
+           (!this->data_transfer_callback.has_value() or this->data_transfer_callback.value() != nullptr) and
+           (!this->transaction_event_callback.has_value() or this->transaction_event_callback.value() != nullptr) and
+           (!this->transaction_event_response_callback.has_value() or
+            this->transaction_event_response_callback.value() != nullptr);
```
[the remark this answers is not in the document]
drmrd: These appear to be optional callbacks. @couryrr-afs: I think this hunk should be removed from the PR along with the corresponding unit tests for `all_callbacks_valid`.

lib/ocpp/v201/charge_point.cpp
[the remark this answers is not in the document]
couryrr-afs: I have added these comments.

doc/ocpp_201_status.md
```diff
@@ -1243,8 +1243,8 @@ This document contains the status of which OCPP 2.0.1 numbered functional requir
 | K01.FR.11 |        |                                                                                                 |
 | K01.FR.12 |        |                                                                                                 |
 | K01.FR.13 |        |                                                                                                 |
-| K01.FR.14 |        |                                                                                                 |
-| K01.FR.15 |        |                                                                                                 |
+| K01.FR.14 | ✅     |                                                                                                 |
+| K01.FR.15 | ✅     |                                                                                                 |
```
shankari, line 1247: These are not addressed in this PR, right?

shankari (approved):
One more clarifying question....

doc/ocpp_201_status.md
couryrr-afs: These were addressed with the check [here](https://github.com/EVerest/libocpp/blob/b67951cebbff4de26456e4575e3ff6e84c4630ed/lib/ocpp/v201/smart_charging.cpp#L319)

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -315,12 +316,32 @@ SmartChargingHandler::validate_profile_schedules(ChargingProfile& profile,
...
 }
 
-void SmartChargingHandler::add_profile(int32_t evse_id, ChargingProfile& profile) {
-    if (STATION_WIDE_ID == evse_id) {
-        station_wide_charging_profiles.push_back(profile);
+SetChargingProfileResponse SmartChargingHandler::add_profile(int32_t evse_id, ChargingProfile& profile) {
+    SetChargingProfileResponse response;
+    response.status = ChargingProfileStatusEnum::Accepted;
+
+    auto& profile_storage = (STATION_WIDE_ID == evse_id) ? station_wide_charging_profiles : charging_profiles[evse_id];
```
marcemmers, line 323: Why do we have a separate store for station wide profiles and evse related charging profiles? For this use case it should be a single list I think. This would also be more in line with how the database is going to look like.

K01.FR.05 says that a profile with the same id should replace the existing profile. There is no mention of if the new profile is for the same EVSE etc.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -315,12 +316,32 @@ SmartChargingHandler::validate_profile_schedules(ChargingProfile& profile,
...
+        *profile_with_id = profile;
     } else {
-        charging_profiles[evse_id].push_back(profile);
+        profile_storage.push_back(profile);
+    }
+
+    return response;
+}
+
+std::vector<ChargingProfile> SmartChargingHandler::get_profiles() {
```
marcemmers, line 339: I see that this is used for testing so can I assume this is temporary? I don't see a other reason to expose all the profiles to the outside since this class is responsible for the implementation.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -916,4 +922,73 @@ TEST_F(ChargepointTestFixtureV201, K01_ValidateProfile_IfChargeStationMaxProfile
...
+       K01FR14_IfTxDefaultProfileWithSameStackLevelDoesNotExist_ThenApplyStationWideTxDefaultProfileToAllEvses) {
+
+    auto profile = create_charging_profile(DEFAULT_PROFILE_ID, ChargingProfilePurposeEnum::TxDefaultProfile,
+                                           create_charge_schedule(ChargingRateUnitEnum::A), uuid(),
+                                           ChargingProfileKindEnum::Absolute, DEFAULT_STACK_LEVEL);
+
+    auto sut = handler.add_profile(STATION_WIDE_ID, profile);
+
+    EXPECT_THAT(sut.status, testing::Eq(ChargingProfileStatusEnum::Accepted));
+    EXPECT_THAT(handler.get_profiles(), testing::Contains(profile));
```
marcemmers, line 935: Building on that last comment about `get_profiles()`, I don't see the point in testing it like this. From the test names they seem like good tests but I expect this to be tested using some kind of composite results in the future since that should be the primary interface of this.

marcemmers (commented):
I have some comments/questions. Not sure if there are any actual changes needed but would like some answers at least before approving.

lib/ocpp/v201/smart_charging.cpp
drmrd: > Why do we have a separate store for station wide profiles and evse related charging profiles? For this use case it should be a single list I think. This would also be more in line with how the database is going to look like.

I believe the separate `station_wide_charging_profiles` variable was created for clarity, since various expected behaviors involving profiles with EVSE ID 0 are different than for other EVSE IDs. (@couryrr-afs please correct me if I'm wrong here.)

drmrd: > K01.FR.05 says that a profile with the same id should replace the existing profile. There is no mention of if the new profile is for the same EVSE etc.

+1. If they're not already present, there should be test cases here for ensuring that adding a replacement charging profile (that is, one with the same ID as an existing profile) for a different EVSE than the existing profile results in the new profile being added and the old one being deleted.

gberardi-pillar: > > Why do we have a separate store for station wide profiles and evse related charging profiles? For this use case it should be a single list I think. This would also be more in line with how the database is going to look like.
> 
> I believe the separate `station_wide_charging_profiles` variable was created for clarity, since various expected behaviors involving profiles with EVSE ID 0 are different than for other EVSE IDs. (@couryrr-afs please correct me if I'm wrong here.)

I seem to recall a PR comment asking for them to be separated, although I am having trouble finding it now.

But yes, it was because the concept of a profile on EVSE with ID 0 is handled differently and so the code clarity of separating it explicitly made sense.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
gberardi-pillar: I'm unclear if you are saying that you want the composite results/future for this PR? For now, we have a straightforward and simple way to verify that profiles are being added, and if this behavior can be tested a different way in the future, we can iterate/incrementally build to that state.

lib/ocpp/v201/smart_charging.cpp
gberardi-pillar: I agree that testing behavior is ideal rather than testing implementation. At the moment, we have a way to confirm that adding profiles works as expected. If, in the future, there is a better way to test the behavior without having this interface, we could always revisit, of course.

lib/ocpp/v201/smart_charging.cpp
christopher-davis-afs: K01.FR.05 says:

> When a SetChargingProfileRequest with an already known ChargingProfile.id is received AND the existing ChargingProfile does NOT have chargingProfilePurpose = `ChargingStationExternalConstraints`, The Charging Station SHALL replace the existing ChargingProfile with the one specified.

When [working on K01.FR.05](https://github.com/US-JOET/libocpp/pull/11) it wasn't clear whether or not the charging profile should be replaced when they're in different places. Station-wide profiles are treated differently from EVSE-specific profiles - they're meant to be apply to all EVSEs. K01.FR.14 also says "The Charging Station SHALL apply, but not copy, this profile to all EVSEs" for profiles on EVSE 0. So that implies that station-wide profiles should be independent from the EVSE-specific profiles. This made it seem odd to replace a station-wide profile that applies to all EVSEs with a profile that doesn't or vice-versa. Should those profiles be replaced then? If so, I think it makes sense to have a test case.

marcemmers: As discussed in the working group call, requirement K01.FR.05 is pretty clear that the charging profile id should be unique for the whole charge point. Therefore it indeed should be replaced and I agree adding a test is the best course of action.

lib/ocpp/v201/smart_charging.cpp
marcemmers: Yep understood, no need to block this for now but I think it's good to look at this again in the future.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
marcemmers: I don't expect any changes based on this comment. This was just elaborating on my other comment about exposing the `get_profiles()` function. I agree that we can iterate this in the future once we have some more composite schedule functions.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -916,4 +922,73 @@ TEST_F(ChargepointTestFixtureV201, K01_ValidateProfile_IfChargeStationMaxProfile
...
+    auto sut1 = handler.add_profile(DEFAULT_EVSE_ID, profile1);
+    auto sut2 = handler.add_profile(DEFAULT_EVSE_ID, profile2);
+
+    auto profiles = handler.get_profiles();
+
+    EXPECT_THAT(sut1.status, testing::Eq(ChargingProfileStatusEnum::Accepted));
+    EXPECT_THAT(sut2.status, testing::Eq(ChargingProfileStatusEnum::Accepted));
+    EXPECT_THAT(profiles, testing::Contains(profile1));
+    EXPECT_THAT(profiles, testing::Contains(profile2));
+}
```
couryrr-afs, line 992: Based on the need for profile_ids to be unique. This test seems to be validating an invalid case.

christopher-davis-afs: With the latest update we've removed this test.

lib/ocpp/v201/smart_charging.cpp
christopher-davis-afs: We refactored how we store profiles and simplified the logic, and added new tests to ensure profile IDs are unique in all cases. We also added some new logic and tests to `validate_profile()` in order to ensure we don't replace `ChargingStationExternalConstraints` profiles.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -319,25 +329,39 @@ SmartChargingHandler::validate_profile_schedules(ChargingProfile& profile,
...
+    charging_profiles[evse_id].push_back(profile);
+
     return response;
 }
 
-std::vector<ChargingProfile> SmartChargingHandler::get_profiles() {
-    auto all_profiles = station_wide_charging_profiles;
+std::vector<ChargingProfile> SmartChargingHandler::get_station_wide_profiles() const {
+    std::vector<ChargingProfile> station_wide_profiles;
+    try {
```
marcemmers, line 354: I think we should be able to do this without needing an exception. Maybe check using `.count()` or `.find()` to check if the `STATION_WIDE_ID` is present.

lib/ocpp/v201/smart_charging.cpp
marcemmers: Looks like a good change, makes sense to use the map to also store the station wide profiles.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -8,13 +8,17 @@
...
 #include "ocpp/v201/ocpp_types.hpp"
 #include "ocpp/v201/transaction.hpp"
 #include <algorithm>
+#include <exception>
 #include <iterator>
 #include <memory>
 #include <ocpp/v201/smart_charging.hpp>
 #include <optional>
+#include <stdexcept>
+#include <vector>
```
marcemmers, line 21: Do you really need all of these includes here? `vector` is already used in the header file so it must be included somehow. `<memory>` is directly included in the header file so not needed here.

marcemmers (approved):
Except for 2 small comments this looks good. The comments are relatively minor and I can accept to merge this as is and fix those later if you want to.

lib/ocpp/v201/smart_charging.cpp
christopher-davis-afs: Sorry, clangd adds imports very aggressively. Removed 👍🏾

PR #685, opened 2024-06-26
charge point: improved thread handling

couryrr-afs:
## Describe your changes

- added a start method to `MessageQueue` for creating thread
- called `MessageQueue::start` in `ChargePoint::start`
- added Destructor to `ChargePoint` to handle cleaning up of threads

### DIFF
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -74,7 +74,8 @@ ChargePoint::ChargePoint(const std::map<int32_t, int32_t>& evse_connector_struct
     csr_attempt(1),
     client_certificate_expiration_check_timer([this]() { this->scheduled_check_client_certificate_expiration(); }),
     v2g_certificate_expiration_check_timer([this]() { this->scheduled_check_v2g_certificate_expiration(); }),
-    callbacks(callbacks) {
+    callbacks(callbacks),
+    stop_auth_cache_cleanup_handler(false) {
     // Make sure the received callback struct is completely filled early before we actually start running
     if (!this->callbacks.all_callbacks_valid()) {
         EVLOG_AND_THROW(std::invalid_argument("All non-optional callbacks must be supplied"));
@@ -162,6 +163,7 @@ ChargePoint::ChargePoint(const std::map<int32_t, int32_t>& evse_connector_struct
     // configure logging
     this->configure_message_logging_format(message_log_path);
 
+    // TODO - this should this be dependency injected.
     this->message_queue = std::make_unique<ocpp::MessageQueue<v201::MessageType>>(
         [this](json message) -> bool { return this->websocket->send(message.dump()); },
         MessageQueueConfig{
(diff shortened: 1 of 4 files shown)


couryrr-afs:
@shankari @christopher-davis-afs  @folkengine and @gberardi-pillar  please review

folkengine (approved):

shankari:
@louisg1337 @the-bay-kay can you please do a co-review first?

louisg1337 (approved):

shankari (changes requested):
I understand the basic goals of this PR, and the changes related to splitting the constructor and the init function. However, I think it would be helpful to clarify why the code related to the `auth_cache_cleanup` is needed.

I can see that when cleaning up the ChargePoint, we want to prevent the auth cache cleanup from happening, but **why**? What happens if we don't do this? Why do we want to unlock before checking for the auth cache handler? Do we then re-lock if we don't break? Where?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3603,6 +3616,11 @@ void ChargePoint::cache_cleanup_handler() {
                 EVLOG_debug << "Time based authorization cache cleanup";
             }
             this->auth_cache_cleanup_required = false;
+            lk.unlock();
```
marcemmers, line 3619: There is no need to unlock the lock here since it is running out of scope. That is why there is a "sub" scope here.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3603,6 +3616,11 @@ void ChargePoint::cache_cleanup_handler() {
...
+        }
```
marcemmers, line 3620: You should also add the `stop_auth_cache_cleanup_handler` to the predicate lambda in the `wait_for`. Currently with the notify it will wake up the thread, run the predicate and see that it does not actually need to wake up so it might sleep for another 15 minutes.

couryrr-afs:
> I understand the basic goals of this PR, and the changes related to splitting the constructor and the init function. However, I think it would be helpful to clarify why the code related to the `auth_cache_cleanup` is needed.
> 
> I can see that when cleaning up the ChargePoint, we want to prevent the auth cache cleanup from happening, but **why**? What happens if we don't do this? Why do we want to unlock before checking for the auth cache handler? Do we then re-lock if we don't break? Where?

@shankari the goal was not to stop the auth cache cleanup but to appropriately stop the background thread when the `ChargePoint` destructor is called. We were experiencing some exception state in our test. There was a conversation [here](https://lfenergy.zulipchat.com/#narrow/stream/417676-EVerest.3A-Cloud-communication/topic/ChargePoint.20Initialization.20and.20Architecture.20Discussions/near/445453927). 

The general flow would be once start is called for the `ChargePoint` the `auth_cache_cleanup` thread is ran in the background and waits 15 minutes to run or until it is notified via either of the boolean in the predicate. The addition of `stop_auth_cache_cleanup_handler` is only used in the destructor so once it is set to true the system is essentially shutting down. Additional locking/unlocking should not be needed.

marcemmers (approved):

shankari:
@couryrr-afs thank you for the link to the context! Again, I would encourage you to add that link upfront to the commit message and to the PR description so that future contributors also have access to the "why" of the change. Having a short summary in the PR description would also aid reviewers.

shankari:
I am approving this since it is clearly correct, but I would like an explanation of why we need both the join and the boolean.
My current understanding is:
- exiting with a joinable thread aborts with `terminate called without an active exception`
- so we join the thread in the destructor. but now the program won't exit until the thread exits
- the boolean ensures that the thread will exit from the while loop in this case, so the program will terminate gracefully

A simple explanation like that (assuming it is correct) would go a long way in setting the context for somebody who is debugging the code, or is an intermittent/new contributor, and will help grow the technical community.

shankari (approved):

couryrr-afs:
> I am approving this since it is clearly correct, but I would like an explanation of why we need both the join and the boolean. My current understanding is:
> 
>     * exiting with a joinable thread aborts with `terminate called without an active exception`
> 
>     * so we join the thread in the destructor. but now the program won't exit until the thread exits
> 
>     * the boolean ensures that the thread will exit from the while loop in this case, so the program will terminate gracefully
> 
> 
> A simple explanation like that (assuming it is correct) would go a long way in setting the context for somebody who is debugging the code, or is an intermittent/new contributor, and will help grow the technical community.

@shankari  your understanding is correct. Going forward I will try and add some more details to the PR comments.

Pietfried:
This never starts the message queue for OCPP1.6. 
https://github.com/EVerest/libocpp/pull/691 fixes this

PR #694, opened 2024-07-11
Add wait_for for futures

Pietfried:
## Describe your changes
This PR changes the handling of all futures used by the charge_point.cpp or charge_point_impl.cpp by adding wait_for for futures. 

libocpp consumers (like everest-core have timeouts that might exceed if the changed function do not return within a defined timeout. This occured e.g. for a DataTransfer.req triggered externally which resulted in an everest-framework cmd timeout of 300s.
Without this change, messages might be 'stuck' in the message queue and the time until the message is being processed in the message queue can be very long. This can e.g. happen if connection problems are present and/or the CSMS does not respond to messages as expected or as per spec. Adding the wait_for of 60 seconds makes sure that these functions return within a defined timeout.


lib/ocpp/v201/charge_point.cpp
```diff
@@ -1997,13 +1998,20 @@ AuthorizeResponse ChargePoint::authorize_req(const IdToken id_token, const std::
     req.certificate = certificate;
     req.iso15118CertificateHashData = ocsp_request_data;
 
+    AuthorizeResponse response;
```
marcemmers, line 2001: We should never come here if we are not connected. That still does not mean we might not run into timeout issues in the cases you mentioned so I do think it is valid to have a timeout here.

We should add an additional check for connected here I would say.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3473,6 +3481,12 @@ DataTransferResponse ChargePoint::data_transfer_req(const DataTransferRequest& r
     ocpp::Call<DataTransferRequest> call(request, this->message_queue->createMessageId());
     auto data_transfer_future = this->send_async<DataTransferRequest>(call);
 
+    if (data_transfer_future.wait_for(DEFAULT_WAIT_FOR_FUTURE_TIMEOUT) == std::future_status::timeout) {
```
marcemmers, line 3484: What happens with the message if we cancel it here. There is no point to send it anymore I would say.

Is it an idea to also provide an interface that just sends the message and the response comes from a callback? Or instead of waiting for the future here, we could return a future that the other side could handle whenever there is a response.

marcemmers:
At least in charge_point there are still some `send_async` left that we should also have a look at I think.

maaikez (approved):

lib/ocpp/v201/charge_point.cpp
Pietfried: There are some reasons why we should still send the message:
* If QueueAllMessages is true
* If the caller doesn't  care about the response from the CSMS

I'm open to discuss other handling for the `send_async` mechansim and I like your suggestion, but I would rather not add this as part of this PR, since this does improve the current implementation. I'd propose to put this on the agenda for the next WG call?

Pietfried:
> At least in charge_point there are still some `send_async` left that we should also have a look at I think.

There was missing handling for the future `on_get_15118_ev_certificate_request` so I added this. For the other `send_async` it is never attempted to retrieve the value of the future, so without looking further into it, I was wondering a bit why send_async is even used there.

hikinggrass (approved):

PR #708, opened 2024-07-24
Only log missing variables on debug

marcemmers:
## Describe your changes

When we request an optional variable and the variable is not present we currently print a warning which is strange since the variable is optional with a reason. Changing this to debug still allows to check the logs but by default will not print it. If `get_value` is used on a variable that is not present we will still log and throw anyway.

Also I changed the default for the monitoring rate to 60 seconds to reduce spam (in debug mode) of missing monitoring vairables. Ideally it should not perform monitoring at all when `MonitoringCtrlrEnabled` is false but that is too big of a change for this PR.

### DIFF
--- a/lib/ocpp/v201/device_model.cpp
+++ b/lib/ocpp/v201/device_model.cpp
@@ -203,7 +203,7 @@ GetVariableStatusEnum DeviceModel::request_value_internal(const Component& compo
                                                           bool allow_write_only) {
     const auto component_it = this->device_model.find(component_id);
     if (component_it == this->device_model.end()) {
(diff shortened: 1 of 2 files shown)


AssemblyJohn:
Since there are safety-critical logs in the spec, I would be happier with the other way around, to change your default variable in the config to 60 and leave the default at 1 second. As for the logs, we probably missed them when we made the PR, no problem there.

1.54.1. SetMonitoringLevelRequest
0-Danger
Indicates lives are potentially in danger. Urgent attention
is needed and action should be taken immediately.

PR #725, opened 2024-08-01
Added verifications before calling callbacks

AssemblyJohn:
## Describe your changes

Added checks for functors before an actual call to it.


Pietfried:
Did you understand how it could happen that these callbacks are `nullptr` when they are called?

AssemblyJohn:
No, I could not find any culprit in that case. An initial theory was that the callback captures the 'this' pointer and that somehow becomes stale, but it is improbable, since in that case we would probably have a segfault.

Pietfried (approved):

Issue #369, opened 2023-12-20
Use Case K10: Clear Charging Profile


RobertDeLeeuw:
ClearChargingProfile already implemented for OCPP 1.6. Still needs to be implemented for OCPP 2.0.1.
Hopefully a lot of code can be reused/shared.

Issue #309, opened 2023-12-14
Use Case K09: Get Charging Profiles


drmrd:

drmrd:
My team is actively working on this use case and others in the Smart Charging functional block.

PR #743, opened 2024-08-09
SmartCharging use cases K09 and K10

Pietfried:
## Describe your changes
Implements SmartCharging use cases K09 and K10.
Adds support for the following messages:
* ClearChargingProfile
* GetChargingProfiles
* ReportChargingProfiles 

Tested Cases tested with OCTT:
TC_K_05
TC_K_06
TC_K_07
TC_K_08
TC_K_09 fails because GetCompositeSchedule handling is a prerequisite.
TC_K_24
TC_K_29
TC_K_30
TC_K_31
TC_K_32
TC_K_33
TC_K_34
TC_K_35
TC_K_36

This PR is branched off https://github.com/EVerest/libocpp/pull/711 .
Additional requires some changes in the code generator, which will be applied in this PR: https://github.com/EVerest/libocpp/pull/741 

https://github.com/EVerest/libocpp/issues/309
https://github.com/EVerest/libocpp/issues/369

### DIFF
--- a/include/ocpp/v201/charge_point.hpp
+++ b/include/ocpp/v201/charge_point.hpp
@@ -688,7 +688,7 @@ private:
 
     // Functional Block K: Smart Charging
     void report_charging_profile_req(const int32_t request_id, const int32_t evse_id,
-                                     const ChargingLimitSourceEnum source, const std::vector<ChargingProfile> profiles,
+                                     const ChargingLimitSourceEnum source, const std::vector<ChargingProfile>& profiles,
                                      const bool tbc);
     void report_charging_profile_req(const ReportChargingProfilesRequest& req);
 
--- a/include/ocpp/v201/charge_point.hpp
+++ b/include/ocpp/v201/charge_point.hpp
@@ -29,11 +29,13 @@
 #include <ocpp/v201/messages/CertificateSigned.hpp>
 #include <ocpp/v201/messages/ChangeAvailability.hpp>
 #include <ocpp/v201/messages/ClearCache.hpp>
+#include <ocpp/v201/messages/ClearChargingProfile.hpp>
 #include <ocpp/v201/messages/ClearVariableMonitoring.hpp>
 #include <ocpp/v201/messages/CustomerInformation.hpp>
 #include <ocpp/v201/messages/DataTransfer.hpp>
 #include <ocpp/v201/messages/DeleteCertificate.hpp>
 #include <ocpp/v201/messages/GetBaseReport.hpp>
+#include <ocpp/v201/messages/GetChargingProfiles.hpp>
 #include <ocpp/v201/messages/GetInstalledCertificateIds.hpp>
 #include <ocpp/v201/messages/GetLocalListVersion.hpp>
 #include <ocpp/v201/messages/GetLog.hpp>
@@ -48,6 +50,7 @@
 #include <ocpp/v201/messages/NotifyEvent.hpp>
 #include <ocpp/v201/messages/NotifyMonitoringReport.hpp>
 #include <ocpp/v201/messages/NotifyReport.hpp>
+#include <ocpp/v201/messages/ReportChargingProfiles.hpp>
 #include <ocpp/v201/messages/RequestStartTransaction.hpp>
 #include <ocpp/v201/messages/RequestStopTransaction.hpp>
 #include <ocpp/v201/messages/Reset.hpp>
@@ -683,6 +686,12 @@ private:
     void meter_values_req(const int32_t evse_id, const std::vector<MeterValue>& meter_values,
                           const bool initiated_by_trigger_message = false);
 
+    // Functional Block K: Smart Charging
+    void report_charging_profile_req(const int32_t request_id, const int32_t evse_id,
+                                     const ChargingLimitSourceEnum source, const std::vector<ChargingProfile> profiles,
+                                     const bool tbc);
+    void report_charging_profile_req(const ReportChargingProfilesRequest& req);
+
     // Functional Block N: Diagnostics
     void notify_event_req(const std::vector<EventData>& events);
     void notify_customer_information_req(const std::string& data, const int32_t request_id);
@@ -726,6 +735,8 @@ private:
 
     // Functional Block K: Smart Charging
     void handle_set_charging_profile_req(Call<SetChargingProfileRequest> call);
+    void handle_clear_charging_profile_req(Call<ClearChargingProfileRequest> call);
+    void handle_get_charging_profiles_req(Call<GetChargingProfilesRequest> call);
 
     // Functional Block L: Firmware management
     void handle_firmware_update_req(Call<UpdateFirmwareRequest> call);
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -49,38 +49,15 @@ enum class ProfileValidationResultEnum {
 
 /// \brief This enhances the ChargingProfile type by additional paramaters that are required in the
 /// ReportChargingProfilesRequest (EvseId, ChargingLimitSourceEnum)
-struct ReportedChargingProfile : public ChargingProfile {
+struct ReportedChargingProfile {
+    ChargingProfile profile;
     int32_t evse_id;
     ChargingLimitSourceEnum source;
 
     ReportedChargingProfile(const ChargingProfile& profile, const int32_t evse_id,
                             const ChargingLimitSourceEnum source) :
-        evse_id(evse_id), source(source) {
-        id = profile.id;
-        stackLevel = profile.stackLevel;
-        chargingProfilePurpose = profile.chargingProfilePurpose;
-        chargingProfileKind = profile.chargingProfileKind;
-        chargingSchedule = profile.chargingSchedule;
-        customData = profile.customData;
-        recurrencyKind = profile.recurrencyKind;
-        validFrom = profile.validFrom;
-        validTo = profile.validTo;
-        transactionId = profile.transactionId;
-    };
-
-    /// \brief Gets the original ChargingProfile from this extended type
-    ChargingProfile get_charging_profile() const {
-        return {id,
-                stackLevel,
-                chargingProfilePurpose,
-                chargingProfileKind,
-                chargingSchedule,
-                customData,
-                recurrencyKind,
-                validFrom,
-                validTo,
-                transactionId};
-    };
+        profile(profile), evse_id(evse_id), source(source) {
+    }
 };
 
 namespace conversions {
@@ -107,7 +84,8 @@ public:
 
     virtual ClearChargingProfileResponse clear_profiles(const ClearChargingProfileRequest& request) = 0;
 
-    virtual std::vector<ReportedChargingProfile> get_profiles(const GetChargingProfilesRequest& request) const = 0;
+    virtual std::vector<ReportedChargingProfile>
+    get_reported_profiles(const GetChargingProfilesRequest& request) const = 0;
 };
 
 /// \brief This class handles and maintains incoming ChargingProfiles and contains the logic
@@ -156,7 +134,8 @@ public:
     ///
     /// \brief Gets the charging profiles for the given \p request
     ///
-    std::vector<ReportedChargingProfile> get_profiles(const GetChargingProfilesRequest& request) const override;
+    std::vector<ReportedChargingProfile>
+    get_reported_profiles(const GetChargingProfilesRequest& request) const override;
 
 protected:
     ///
--- a/include/ocpp/v201/smart_charging.hpp
+++ b/include/ocpp/v201/smart_charging.hpp
@@ -11,6 +11,8 @@
 #include <ocpp/v201/device_model.hpp>
 #include <ocpp/v201/evse_manager.hpp>
 #include <ocpp/v201/messages/SetChargingProfile.hpp>
+#include <ocpp/v201/messages/ClearChargingProfile.hpp>
+#include <ocpp/v201/messages/GetChargingProfiles.hpp>
 #include <ocpp/v201/ocpp_enums.hpp>
 #include <ocpp/v201/ocpp_types.hpp>
 #include <ocpp/v201/transaction.hpp>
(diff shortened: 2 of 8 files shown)


include/ocpp/v201/charge_point.hpp
```diff
@@ -683,6 +686,12 @@ class ChargePoint : public ChargePointInterface, private ocpp::ChargingStationBa
     void meter_values_req(const int32_t evse_id, const std::vector<MeterValue>& meter_values,
                           const bool initiated_by_trigger_message = false);
 
+    // Functional Block K: Smart Charging
+    void report_charging_profile_req(const int32_t request_id, const int32_t evse_id,
+                                     const ChargingLimitSourceEnum source, const std::vector<ChargingProfile> profiles,
```
marcemmers, line 691: I think in this case codacy is right and its better to pass as a reference

include/ocpp/v201/smart_charging.hpp
```diff
@@ -45,6 +47,42 @@ enum class ProfileValidationResultEnum {
     DuplicateProfileValidityPeriod
 };
 
+/// \brief This enhances the ChargingProfile type by additional paramaters that are required in the
+/// ReportChargingProfilesRequest (EvseId, ChargingLimitSourceEnum)
+struct ReportedChargingProfile : public ChargingProfile {
```
gberardi-pillar, line 52: Why derive from ChargingProfile? Why not use composition, since this struct seems to do nothing more than provide two other data members, and you provide the ability to get back the original profile anyway. Instead, you can make the ChargingProfile a member.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -120,6 +120,81 @@ std::ostream& operator<<(std::ostream& os, const ProfileValidationResultEnum val
...
+            criteria.chargingProfileId.value().end()) {
+        return true;
+    }
+
+    if (criteria.chargingProfileId.has_value()) {
+        // no profile matches the id, no need to do check match of the criteria
+        return false;
+    }
+
+    if (criteria.chargingLimitSource.has_value() and
```
gberardi-pillar, line 176: Can refactor this into a function:  charging_limit_source_exists_and_is_not_CSO() 

or !charging_limit_source_exists_and_is_CSO()


And do similar refactorings for the rest of the if statements, so that it reads more like:

If (x || y || z || ...) {
   return false;
}

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -436,6 +511,55 @@ std::vector<ChargingProfile> SmartChargingHandler::get_profiles() const {
...
 }
 
+ClearChargingProfileResponse SmartChargingHandler::clear_profiles(const ClearChargingProfileRequest& request) {
+    ClearChargingProfileResponse response;
+
+    const auto profile_id = request.chargingProfileId;
+    const auto criteria = request.chargingProfileCriteria;
+
+    auto any_found_profile = false;
+    // iterate over all profiles
```
gberardi-pillar, line 521: Is this comment necessary?

include/ocpp/v201/smart_charging.hpp
```diff
@@ -106,6 +148,16 @@ class SmartChargingHandler : public SmartChargingHandlerInterface {
...
 
+    ///
+    /// \brief Clears profiles from the system using the given \p request
+    ///
+    ClearChargingProfileResponse clear_profiles(const ClearChargingProfileRequest& request) override;
+
+    ///
+    /// \brief Gets the charging profiles for the given \p request
+    ///
+    std::vector<ReportedChargingProfile> get_profiles(const GetChargingProfilesRequest& request) const override;
```
christopher-davis-afs, line 159: I would name this differently, as we've got `get_profiles()` above that does something completely differently. When reading through the tests, I got confused about what we were doing when we called `get_profiles()`. One name would be `get_reported_profiles()` or `get_matching_profiles()`.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -436,6 +511,55 @@ std::vector<ChargingProfile> SmartChargingHandler::get_profiles() const {
...
+    for (auto& [existing_evse_id, evse_profiles] : charging_profiles) {
+        for (auto it = evse_profiles.begin(); it != evse_profiles.end();) {
+            // K10.FR.04: logical AND between the filters, if not set, the filter does not apply
+            if (profile_matches_clear_criteria(*it, profile_id, criteria, existing_evse_id)) {
+                any_found_profile = true;
+                evse_profiles.erase(it);
```
gberardi-pillar, line 527: I believe this line should be:

it = evse_profiles.erase(it);

Otherwise, it is an invalidated iterator and can't be depended upon to point to the correct part of the vector.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -436,6 +511,55 @@ std::vector<ChargingProfile> SmartChargingHandler::get_profiles() const {
...
+                this->database_handler->delete_charging_profile(it->id);
+            } else {
+                // At least one of the filters did not match
+                ++it;
+            }
+        }
+    }
+
+    if (any_found_profile) {
```
gberardi-pillar, line 536: Can this boolean be removed? Set the response to Unknown by default, and on line 526 set it to Accepted.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -1168,4 +1208,233 @@ TEST_F(ChargepointTestFixtureV201, K01_ValidateAndAdd_AddsValidProfiles) {
...
+    auto periods = create_charging_schedule_periods({0, 1, 2});
+
+    auto profile1 = create_charging_profile(
+        1, ChargingProfilePurposeEnum::ChargingStationMaxProfile,
+        create_charge_schedule(ChargingRateUnitEnum::A, periods, ocpp::DateTime("2024-01-17T17:00:00")));
+    auto profile2 = create_charging_profile(
+        2, ChargingProfilePurposeEnum::TxDefaultProfile,
+        create_charge_schedule(ChargingRateUnitEnum::A, periods, ocpp::DateTime("2024-01-17T17:00:00")));
+
+    auto sut1 = handler.validate_and_add_profile(profile1, 0);
```
gberardi-pillar, line 1221: Replace 0 with STATION_WIDE_ID

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -1168,4 +1208,233 @@ TEST_F(ChargepointTestFixtureV201, K01_ValidateAndAdd_AddsValidProfiles) {
...
+    auto sut2 = handler.validate_and_add_profile(profile2, DEFAULT_EVSE_ID);
+
+    auto profiles = handler.get_profiles();
+    EXPECT_THAT(profiles, testing::SizeIs(2));
+
+    auto reported_profiles = handler.get_profiles(
+        create_get_charging_profile_request(DEFAULT_REQUEST_ID, create_charging_profile_criteria(), 0));
```
gberardi-pillar, line 1228: You can use the constants for the EVSE ID when calling create_get_charging_profile_request() in the tests to make it clear how this line ties back to the setup of the tests.

lib/ocpp/v201/smart_charging.cpp
```diff
@@ -436,6 +511,55 @@ std::vector<ChargingProfile> SmartChargingHandler::get_profiles() const {
...
+    std::vector<ReportedChargingProfile> profiles;
+
+    for (auto& [existing_evse_id, evse_profiles] : charging_profiles) {
+        for (auto it = evse_profiles.begin(); it != evse_profiles.end(); it++) {
+            if (profile_matches_get_criteria(*it, request.chargingProfile, request.evseId, existing_evse_id)) {
+                profiles.push_back(
+                    ReportedChargingProfile(*it, existing_evse_id,
+                                            ChargingLimitSourceEnum::CSO)); // TODO: Add correct source when available
+            }
+        }
```
christopher-davis-afs, line 557: Any particular reason we're using this notation with the iterator `it`? As far as I can tell, within this loop we're not operating on the iterator itself like we do in `clear_profiles()`, so could this be written as:

```c++
    for (auto& [existing_evse_id, evse_profiles] : charging_profiles) {
        for (auto& profile : evse_profiles) {
            if (profile_matches_get_criteria(profile, request.chargingProfile, request.evseId, existing_evse_id)) {
                profiles.push_back(
                    ReportedChargingProfile(profile, existing_evse_id,
                                            ChargingLimitSourceEnum::CSO)); // TODO: Add correct source when available
            }
        }
    }
```

christopher-davis-afs (changes requested):

include/ocpp/v201/smart_charging.hpp
Pietfried: I wouldnt say that the two functions do something completely different. But given that the return type is different I agree that `get_reported_profiles` reads better.

lib/ocpp/v201/smart_charging.cpp
Pietfried: Removed it

lib/ocpp/v201/smart_charging.cpp
Pietfried: I added a templated function `value_exists_and_is_different` to be able to reuse this in `profile_matches_clear_criteria` and `profile_matches_get_criteria`

include/ocpp/v201/charge_point.hpp
```diff
@@ -726,6 +735,8 @@ class ChargePoint : public ChargePointInterface, private ocpp::ChargingStationBa
 
     // Functional Block K: Smart Charging
     void handle_set_charging_profile_req(Call<SetChargingProfileRequest> call);
+    void handle_clear_charging_profile_req(Call<ClearChargingProfileRequest> call);
```
marcemmers, line 738: Why do we pass these by value? We pass by value for all calls but I think it could very well be `const &`

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3207,6 +3232,97 @@ void ChargePoint::handle_set_charging_profile_req(Call<SetChargingProfileRequest
...
+    response.status = ClearChargingProfileStatusEnum::Unknown;
+
+    // K10.FR.06
+    if (msg.chargingProfileCriteria.has_value() and
+        msg.chargingProfileCriteria.value().chargingProfilePurpose.has_value() and
+        msg.chargingProfileCriteria.value().chargingProfilePurpose.value() ==
+            ChargingProfilePurposeEnum::ChargingStationExternalConstraints) {
+        response.statusInfo = StatusInfo();
+        response.statusInfo->reasonCode = "InvalidValue";
+        response.statusInfo->additionalInfo = "ChargingStationExternalConstraintsInClearChargingProfileRequest";
```
marcemmers, line 3248: I think we can make this string a bit more readable with spaces instead of camelcase. There are no rules for this string right?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3207,6 +3232,97 @@ void ChargePoint::handle_set_charging_profile_req(Call<SetChargingProfileRequest
...
+        EVLOG_debug << "Rejecting SetChargingProfileRequest:\n reasonCode: " << response.statusInfo->reasonCode.get()
+                    << "\nadditionalInfo: " << response.statusInfo->additionalInfo->get();
+    } else {
+        response = this->smart_charging_handler->clear_profiles(msg);
```
marcemmers, line 3252: I am a bit worried about the responsibilities of the smart charging handler. We filter out some edge cases but then still hand over the whole message to `clear_profiles` and `get_reported_profiles`. I think we should either move the whole handling of messages there or put an in between layer that handles the full message here. This feels like a kind of halfway solution.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3207,6 +3232,97 @@ void ChargePoint::handle_set_charging_profile_req(Call<SetChargingProfileRequest
...
+}
+
+void ChargePoint::handle_get_charging_profiles_req(Call<GetChargingProfilesRequest> call) {
+    EVLOG_debug << "Received GetChargingProfilesRequest: " << call.msg << "\nwith messageId: " << call.uniqueId;
+    const auto msg = call.msg;
+    GetChargingProfilesResponse response;
+
+    const auto profiles_to_report = this->smart_charging_handler->get_reported_profiles(msg);
+
+    if (not profiles_to_report.empty()) {
```
marcemmers, line 3266: This is a bit confusing for me. I would only to the exit path in the if and then the normal flow outside of the if/else.

lib/ocpp/v201/charge_point.cpp
Pietfried: So far this is based on the behavior of `handle_set_charging_profile_req`.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3207,6 +3232,96 @@ void ChargePoint::handle_set_charging_profile_req(Call<SetChargingProfileRequest
...
+    // required, so we can not report all profiles in a single ReportChargingProfilesRequest. We need to prepare a
+    // single ReportChargingProfilesRequest for each combination of evseId and chargingLimitSource
+    std::set<int32_t> evse_ids;                // will contain all evse_ids of the profiles
+    std::set<ChargingLimitSourceEnum> sources; // will contain all sources of the profiles
+
+    // fill evse_ids and sources sets
+    std::transform(profiles_to_report.begin(), profiles_to_report.end(), std::inserter(evse_ids, evse_ids.end()),
+                   [](const ReportedChargingProfile& profile) { return profile.evse_id; });
+    std::transform(profiles_to_report.begin(), profiles_to_report.end(), std::inserter(sources, sources.end()),
+                   [](const ReportedChargingProfile& profile) { return profile.source; });
```
marcemmers, line 3287: While I do appreciate a good algorithm, this could maybe be simpler with a small for loop like this:
```suggestion
    for(const auto& profile: profiles to report) {
        evse_ids.insert(profile.evse_id);
        sources.insert(profile.source);
    }
```

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3207,6 +3232,96 @@ void ChargePoint::handle_set_charging_profile_req(Call<SetChargingProfileRequest
...
+
+    std::vector<ReportChargingProfilesRequest> requests_to_send;
+
+    for (const auto evse_id : evse_ids) {
+        for (const auto source : sources) {
+            std::vector<ChargingProfile> original_profiles;
+            std::for_each(profiles_to_report.begin(), profiles_to_report.end(),
```
marcemmers, line 3294: Also maybe a normal range based for?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3207,6 +3232,96 @@ void ChargePoint::handle_set_charging_profile_req(Call<SetChargingProfileRequest
...
+                req.requestId = msg.requestId; // K09.FR.01
+                req.evseId = evse_id;
+                req.chargingLimitSource = source;
+                req.chargingProfile = original_profiles;
+                requests_to_send.push_back(req);
+            }
+        }
+    }
+
+    // requests_to_send are ready, send them and define tbc property
```
marcemmers, line 3312: I think you can set tbc default to true and then do `requests_to_send.back().tbc = false`. Then its just iterating all to send.

marcemmers (approved):
In general looks quite good I think. There are some things I would improve but they are more nice to haves and should not change the functionality so up to you to do them or not.

lib/ocpp/v201/charge_point.cpp
marcemmers: For both of these I probably missed it earlier then or it was not as apparent. Maybe we don't need to do anything with it now but we are increasing coupling this way so certainly something to look out for later.

lib/ocpp/v201/charge_point.cpp
christopher-davis-afs: Some thoughts:

* This is modeled after `ProfileValidationResultEnum` I assume. Looking at this and some upcoming code for GetCompositeSchedule, I wonder if we should rename that to be something like `SmartChargingResultEnum` (or have a separate, similar enum)
* Related, the conversion we use essentially just uses the name of the enum. If we want these to be more readable across the board, we should probably change those too.

I think it would be good to standardize this somewhat, and then in cases like this we could generate `statusInfo` from an enum.

Pietfried: I like this idea, probably its better to not rename but have a separate enum for that, since `ProfileValidationResultEnum` is really a result of the validation of the profile. I would not change this as part of this PR (and dont expect e.g. https://github.com/EVerest/libocpp/pull/745 to something for it) but rather add this as part of a subsequent PR.

christopher-davis-afs (approved):

gberardi-pillar (approved):

PR #769, opened 2024-09-03
408 extend network connection profiles

### LINKED ISSUE #408: Extend Network Connection Profiles
To extend network connection profile for a more optional fine-grain control the following **optional** API interfaces need to be added:

1. Set up an interface. A callback can be implemented to set up the interface. For example, if the interface is a modem, it must first be enabled before it is possible to connect to that interface. 
2. Force close the web socket. Can be used to force the web socket to close when the low-level network interface goes down. This is much faster than waiting for a web socket timeout.
3. Switch to a higher priority interface. Normally, if libocpp is connected with a network connection profile, it will not disconnect. However, there may be a situation where libocpp is connected to a profile with priority 2 or lower, and you find out at system level that an interface (with a higher priority) has changed and is now up. In this case, you may want to tell libocpp that the higher priority priority interface is up and that it can try to connect to it. 
4. Connected and disconnected web socket callback. It will provide the network slot in these callbacks so that you can keep the network connection in use (e.g. do not disable the modem) or deactivate it if necessary.

WilcodenBesten:
## Describe your changes

Resolved #408 

- [x ] I have performed a self-review of my code
- [n/a ] If OCPP 2.0.1: I have updated the [OCPP 2.0.1 status document](https://github.com/EVerest/libocpp/tree/main/doc/ocpp_201_status.md)


include/ocpp/v201/ctrlr_component_variables.hpp
```diff
@@ -83,6 +84,7 @@ extern const ComponentVariable& ClientCertificateExpireCheckIntervalSeconds;
 extern const ComponentVariable& MessageQueueSizeThreshold;
 extern const ComponentVariable& MaxMessageSize;
 extern const ComponentVariable& ResumeTransactionsOnBoot;
+extern const ComponentVariable& NetworkConfigTimeout;
```
maaikez, line 87: You added the same variable twice?

WilcodenBesten: My bad 😓

lib/ocpp/v201/connectivity_manager.cpp
```diff
@@ -265,5 +333,97 @@ void ConnectivityManager::next_network_configuration_priority() {
...
+            if (reason != WebsocketCloseReason::ServiceRestart) {
+                if (!next_priority.has_value()) {
+                    this->next_network_configuration_priority();
+                } else {
+                    this->network_configuration_priority = next_priority.value();
+                }
+            }
+            this->start();
+        },
+        WEBSOCKET_INIT_DELAY);
```
maaikez, line 371: Why does the reconnect always have a timeout of 2 seconds?

WilcodenBesten: No sure, this was already present (see deleted line 203in this file), so I just left it as-is.

maaikez: Yes but now you do it always when reconnecting and I don't know if that is necessary. 

@AssemblyJohn or @hikinggrass do you know maybe?

AssemblyJohn: The websocket reconnect interval is calculated in the 'get_reconnect_internal' here: https://github.com/EVerest/libocpp/blob/0c5aac4d8f2a3e3512ea3233e89810b39247b082/lib/ocpp/common/websocket/websocket_base.cpp#L149

maaikez: Yes but was there another reason to add 2 seconds to that, for example because of the old websocket implementation?
the reconnect_backoff_ms is still calculated and used when no connection could be made.

AssemblyJohn: I am sorry, for the initial delay I don't know what is with those 2 seconds init delay. It is not related to the internal websocket functionality that I had to handle. Run a git blame to see who is responsible for that delay code.

maaikez (approved):
Not sure about the 2 second delay like my earlier comment but otherwise it looks fine and it still connects with a CSMS.

lib/ocpp/v201/connectivity_manager.cpp
WilcodenBesten: Git blame results in it being added as part of https://github.com/EVerest/libocpp/pull/129, but I cannot fine any explanation as of why.

include/ocpp/v201/charge_point.hpp
```diff
@@ -932,6 +939,12 @@ class ChargePoint : public ChargePointInterface, private ocpp::ChargingStationBa
     std::vector<CompositeSchedule> get_all_composite_schedules(const int32_t duration,
                                                                const ChargingRateUnitEnum& unit) override;
 
+    std::optional<NetworkConnectionProfile> get_network_connection_profile(const int32_t configuration_slot) override;
```
WilcodenBesten, line 942: Add function description to these new functions

include/ocpp/v201/charge_point.hpp
```diff
@@ -316,6 +337,26 @@ class ChargePointInterface {
     /// \return vector of composite schedules, one for each evse_id including 0.
     virtual std::vector<CompositeSchedule> get_all_composite_schedules(const int32_t duration,
                                                                        const ChargingRateUnitEnum& unit) = 0;
+
+    /// \brief Gets the configured NetworkConnectionProfile based on the given \p configuration_slot . The
+    /// central system uri of the connection options will not contain ws:// or wss:// because this method removes it if
+    /// present. This returns the value from the cached network connection profiles. \param
+    /// network_configuration_priority \return
+    virtual std::optional<NetworkConnectionProfile>
```
marcemmers, line 345: Am I correct that these functions are purely used to get parts of the cached network profiles? Might it be an idea to just create one function that returns a const reference to an object containing all information?

include/ocpp/v201/charge_point.hpp
```diff
@@ -316,6 +337,26 @@ class ChargePointInterface {
...
+    /// \return The priority if the configuration slot exists.
+    ///
+    virtual std::optional<int> get_configuration_slot_priority(const int configuration_slot) = 0;
+
+    /// @brief Get the network connection priorities.
+    /// Each item in the vector contains the configured configuration slots, where the slot with index 0 has the highest
+    /// priority.
+    /// @return The network connection priorities
+    ///
+    virtual const std::vector<std::string>& get_network_connection_priorities() const = 0;
```
marcemmers, line 359: If we are not going to do the combined object, I would at least make this into a vector of integers (keep the parsing inside the library)

lib/ocpp/v201/connectivity_manager.cpp
```diff
@@ -82,13 +85,31 @@ ConnectivityManager::get_network_connection_profile(const int32_t configuration_
...
+                        configuration_slot_string);
+    if (it != network_connection_priorities.end()) {
+        // Index is iterator - begin iterator
+        return it - network_connection_priorities.begin();
+    }
+    return std::nullopt;
+}
+
+const std::vector<std::string>& ConnectivityManager::get_network_connection_priorities() const {
+    return network_connection_priorities;
```
marcemmers, line 102: To continue on that, you might even want to store this as a vector of int instead of string, do the parsing only once. Will help with `get_configuration_slot_priority` as well.

marcemmers, line 102: It's not an official rule but to keep style (also some other places):
```suggestion
    return this->network_connection_priorities;
```

lib/ocpp/v201/connectivity_manager.cpp
```diff
@@ -131,15 +196,50 @@ void ConnectivityManager::init_websocket() {
...
+    auto connection_options = this->get_ws_connection_options(config_slot_int);
+    bool can_use_connection_profile = true;
+
+    if (!network_connection_profile.has_value()) {
+        EVLOG_warning << "No network connection profile configured for " << config_slot_int;
+        can_use_connection_profile = false;
+    } else if (this->configure_network_connection_profile_callback.has_value()) {
+        EVLOG_debug << "Request to configure network connection profile " << config_slot_int;
+
+        std::future<ConfigNetworkResult> config_status = this->configure_network_connection_profile_callback.value()(
```
marcemmers, line 212: Might be good to split this off to a function to handle the callback so this reads a bit cleaner

lib/ocpp/v201/connectivity_manager.cpp
```diff
@@ -121,6 +142,50 @@ bool ConnectivityManager::send_to_websocket(const std::string& message) {
...
+        this->get_network_connection_profile(actual_configuration_slot);
+
+    if (!network_connection_profile.has_value()) {
+        EVLOG_warning << "Network disconnected. No network connection profile configured";
+    } else if ((configuration_slot.has_value() and (configuration_slot.value() == actual_configuration_slot)) or
+               (ocpp_interface.has_value() and
+                (ocpp_interface.value() == network_connection_profile.value().ocppInterface))) {
+        this->disconnect_websocket(ocpp::WebsocketCloseReason::GoingAway);
+        // Since there is no connection anymore: reconnect with the next available network connection profile.
+        reconnect(ocpp::WebsocketCloseReason::GoingAway);
```
jannejy, line 165: After the PR https://github.com/EVerest/libocpp/pull/827 the reason is sent to the callback register_closed_callback. Why don't we take advantage of here and call reconnect from one place? Maybe do we need to think about the new reason for disconnect?

marcemmers (changes requested):

include/ocpp/v201/charge_point.hpp
```diff
@@ -109,6 +108,28 @@ class ChargePointInterface {
...
+    ///
+    /// This is introduced because the websocket can take several minutes to timeout when a network interface becomes
+    /// unavailable, whereas the system can detect this sooner.
+    ///
+    /// \param configuration_slot   The slot of the network connection profile that is disconnected.
+    /// \param ocpp_interface       The interface that is disconnected.
+    ///
+    /// \note At least one of the two params must be provided, otherwise libocpp will not know which interface is down.
+    ///
+    virtual void on_network_disconnected(const std::optional<int32_t> configuration_slot,
```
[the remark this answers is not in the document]
maaikez: There is no situation where we need both?
If that's indeed the case, I agree.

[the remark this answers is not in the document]
WilcodenBesten: No either one of the other, I'll rework this.

include/ocpp/v201/charge_point.hpp
WilcodenBesten: Jep. I guess we can optimize this by caching the configured network priorities based on their configured priority.

lib/ocpp/v201/connectivity_manager.cpp
WilcodenBesten: I'm not entirely sure what you mean or would like me to change here. Could you elaborate a bit more?

lib/ocpp/v201/connectivity_manager.cpp
```diff
@@ -121,6 +142,50 @@ bool ConnectivityManager::send_to_websocket(const std::string& message) {
...
+
+    const std::optional<NetworkConnectionProfile> network_connection_profile_opt =
+        this->get_network_connection_profile(configuration_slot);
+    if (!network_connection_profile_opt.has_value()) {
+        EVLOG_warning << "Could not find network connection profile belonging to configuration slot "
+                      << configuration_slot;
+        return false;
+    }
+
+    reconnect(WebsocketCloseReason::Normal, get_configuration_slot_priority(configuration_slot));
```
jannejy, line 185: Don't we really need here to disconnect websocket first?

WilcodenBesten: You're right that it's not explicitly closed here. The websocket itself also does things based on the actions of this class, so I'll check if it's properly closed when you call this function.

lib/ocpp/v201/connectivity_manager.cpp
jannejy: You can call `reconnect` from the one place (from the `register_closed_callback`), that gives better readability and maintainability. I would use this binding: call `disconnect_websocket` with a reason and in the callback `register_closed_callback` depending on the reason (restart or switch to another network profile), that is now propagated there, call `reconnect`

include/ocpp/v201/charge_point.hpp
WilcodenBesten: Since you also configure the profiles and priorities separately via OCPP, I'm keeping this separate as well. I did rework it so it's no longer a vector of strings.

doc/networkconnectivity/README.md
```diff
@@ -0,0 +1,55 @@
...
+libocpp provides two callbacks for when the websocket is connected and disconnected. It will provide the network slot
+in these callbacks, so you can keep the network connection in use (e.g. not disable the modem), or disable the network connection (example again: disable the modem).
+
+## Sequence diagram
+
+'core' can be read as any application that implements libocpp
+
+For step 9, ping is one way to check if a CSMS is up, but you of course can implement a way to check this yourself.
+
+![Sequence diagram](networkconnectivity_libocpp.png)
```
marcemmers, line 55: We found out yesterday that github supports inline mermaid, so please replace the image and puml files with inline mermaid.

marcemmers (approved):

Issue #781, opened 2024-09-09
Offline behavior for SmartCharging


Pietfried:
### OCPP Version

OCPP1.6, OCPP2.0.1

### Describe the problem

OCPP1.6 requires the following with respect to the offline behavior for Smart Charging 

>If a Charge Point goes offline after having received a transaction-specific charging profile with purpose TxProfile,
then it SHALL continue to use this profile for the duration of the transaction.
If a Charge Point goes offline before a transaction is started or before a transaction-specific charging profile with
purpose TxProfile was received, then it SHALL use the charging profiles that are available.

For OCPP2.0.1 the use cases K06 and K07 basically require the same behavior.

In real world applications it is common that other systems, software or the user take over in case the the charging station is offline. Especially the requirement that the charging station shall continue to use `TxProfile`s when offline leads to the situation the a user has no option to override that value in situations where this might be desired (e.g. the last requried setpoint was 0A or very high, which might have been reasonable under the conditions known at that time). This is an issue for profiles that specify setpoints for longer periods of time or infinitely in the future. The conditions that influence the charging power (local limits, solar power, driver wishes) might have changed when being offline and the changed condition can not be applied for a new setpoint if the charging station is offline.

### Describe your solution

In order to allow other other systems or software to take over in an offline scenario, a feature to ignore certain profiles from the CSMS when offline when calculating the composite schedule could solve the issue described above. 

A configuration variable `IgnoredProfilePurposesOffline` could specify charging profile purposes to be ignored when offline. This should default to being empty. 

### Additional context

Enabling this feature should be used with care and does violate the specification, since the given profiles SHALL be used if present.

Simply ignoring all profiles can be solved by configuring all charging profile purposes within `IgnoredProfilePurposesOffline`. In some cases it might be desired to only ignore `TxProfile`s or e.g. `TxDefaultProfile`s

PR #783, opened 2024-09-10
Feature for SmartCharging Offline Behavior

Pietfried:
## Describe your changes
Added configuration variable to 1.6 and 2.0.1 to allow ignoring profile in case the charging station is offline when the composite schedules are calculated by requests of a libocpp consumer application

https://github.com/EVerest/libocpp/issues/781

### DIFF
--- a/config/v201/component_config/standardized/SmartChargingCtrlr.json
+++ b/config/v201/component_config/standardized/SmartChargingCtrlr.json
@@ -253,6 +253,23 @@
       "description": "Supply voltage of the grid. This value is only used in case a conversion between smart charging amp and watt limits is required"   
     }
   },
+  "IgnoredProfilePurposesOffline": {
+    "variable_name": "IgnoredProfilePurposesOffline",
+      "characteristics": {
+        "valuesList": "ChargingStationMaxProfile,TxDefaultProfile,TxProfile",
+        "supportsMonitoring": true,
+        "dataType": "MemberList"
+      },
+      "attributes": [
+        {
+          "type": "Actual",
+          "mutability": "ReadOnly",
+          "value": ""
+        }
+      ],
+      "description": "Allows configuration of comma seperated list of ChargingProfilePurposes that are ignored in the composite schedule caluclation when offline.",
+      "type": "string"
+  },
   "required": [
     "ChargingProfileMaxStackLevel",
     "ChargingScheduleChargingRateUnit",
--- a/config/v201/component_config/standardized/SmartChargingCtrlr.json
+++ b/config/v201/component_config/standardized/SmartChargingCtrlr.json
@@ -263,7 +263,7 @@
       "attributes": [
         {
           "type": "Actual",
-          "mutability": "ReadOnly",
+          "mutability": "ReadWrite",
           "value": ""
         }
       ],
--- a/lib/ocpp/v16/smart_charging.cpp
+++ b/lib/ocpp/v16/smart_charging.cpp
@@ -396,7 +396,7 @@ SmartChargingHandler::get_valid_profiles(const ocpp::DateTime& start_time, const
         std::lock_guard<std::mutex> lk(charge_point_max_profiles_map_mutex);
 
         if (std::find(std::begin(purposes_to_ignore), std::end(purposes_to_ignore),
-                      ChargingProfilePurposeType::ChargePointMaxProfile) != std::end(purposes_to_ignore)) {
+                      ChargingProfilePurposeType::ChargePointMaxProfile) == std::end(purposes_to_ignore)) {
             for (const auto& [stack_level, profile] : stack_level_charge_point_max_profiles_map) {
                 valid_profiles.push_back(profile);
             }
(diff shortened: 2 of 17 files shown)


lib/ocpp/v16/smart_charging.cpp
```diff
@@ -375,16 +375,19 @@ void SmartChargingHandler::clear_all_profiles() {
...
+                                         const int connector_id,
+                                         const std::set<ChargingProfilePurposeType>& purposes_to_ignore) {
     std::vector<ChargingProfile> valid_profiles;
 
     {
         std::lock_guard<std::mutex> lk(charge_point_max_profiles_map_mutex);
 
-        for (const auto& [stack_level, profile] : stack_level_charge_point_max_profiles_map) {
-            valid_profiles.push_back(profile);
+        if (not purposes_to_ignore.count(ChargingProfilePurposeType::ChargePointMaxProfile)) {
```
james-ctc, line 387: observation
I would consider using EnumFlags here rather than a set.
For a set (or other container based solution) count() has the disadvantage that it doesn't stop on finding the first matching entry. using find() would be better.
Also count() returns a size type rather than a boolean; checking `== 0` or `!= 0` is more appropriate than using `not`.
e.g.
```
if (std::find(
    std::begin(purposes_to_ignore), 
    std::end(purposes_to_ignore), 
    ChargingProfilePurposeType::ChargePointMaxProfile) != std::end(purposes_to_ignore)) {...}
```
or
```
if (auto search = purposes_to_ignore.find(ChargingProfilePurposeType::ChargePointMaxProfile); search != purposes_to_ignore.end()) {...}
```

lib/ocpp/v201/charge_point.cpp
```diff
@@ -4058,21 +4061,30 @@ ChargePoint::set_variables(const std::vector<SetVariableData>& set_variable_data
 }
 
 GetCompositeScheduleResponse ChargePoint::get_composite_schedule(const GetCompositeScheduleRequest& request) {
-    return this->get_composite_schedule_internal(request);
+    std::set<ChargingProfilePurposeEnum> purposes_to_ignore = utils::get_purposes_to_ignore(
+        this->device_model->get_optional_value<std::string>(ControllerComponentVariables::IgnoredProfilePurposesOffline)
+            .value_or(""),
+        this->is_offline());
+    return this->get_composite_schedule_internal(request, purposes_to_ignore);
```
james-ctc, line 4068: is there an EVerest style guide covering the use of `this->`?
it is often needed in lambdas but here it shouldn't be needed.

lib/ocpp/v201/utils.cpp
```diff
@@ -183,6 +183,27 @@ bool is_critical(const std::string& security_event) {
...
 }
 
+std::set<ChargingProfilePurposeEnum> get_purposes_to_ignore(const std::string& csl, const bool is_offline) {
+    if (not is_offline) {
+        return {};
+    }
+
+    if (csl.empty()) {
+        return {};
+    }
```
james-ctc, line 193: could be simplified:
```
if (!is_offline() || csl.empty()) {
  return {};
}
```
Some coding standards recommend avoiding having multiple return statements (MISRA goes further allowing a single return statement).

MISRA is generally looking for
```
std::set<ChargingProfilePurposeEnum> result{};
if (is_offline && !csl.empty()) {
    ... main for loop ...
}
return result;
```

james-ctc (commented):
a few questions and suggestions.
looks okay to me.

config/v201/component_config/standardized/SmartChargingCtrlr.json
```diff
@@ -180,6 +180,23 @@
       "type": "string"
     }
   },
+  "IgnoredProfilePurposesOffline": {
+    "variable_name": "IgnoredProfilePurposesOffline",
+      "characteristics": {
+        "valuesList": "ChargingStationMaxProfile,TxDefaultProfile,TxProfile",
```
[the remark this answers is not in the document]
Pietfried: I don't think that the CSMS should be able to configure ignoring this purpose, especially in an offline case. I don't see a valid use case for this.

lib/ocpp/v201/charge_point.cpp
Pietfried: There is no style guide yet. Im coming from java (where `this` is required) and personally like it to explicitly mark the members.

config/v201/component_config/standardized/SmartChargingCtrlr.json
```diff
@@ -180,6 +180,23 @@
...
+        "supportsMonitoring": true,
+        "dataType": "MemberList"
+      },
+      "attributes": [
+        {
+          "type": "Actual",
+          "mutability": "ReadOnly",
```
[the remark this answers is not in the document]
Pietfried: I changed it to `ReadWrite`

lib/ocpp/v16/smart_charging.cpp
```diff
@@ -375,16 +375,20 @@ void SmartChargingHandler::clear_all_profiles() {
...
+                                         const int connector_id,
+                                         const std::set<ChargingProfilePurposeType>& purposes_to_ignore) {
     std::vector<ChargingProfile> valid_profiles;
 
     {
         std::lock_guard<std::mutex> lk(charge_point_max_profiles_map_mutex);
 
-        for (const auto& [stack_level, profile] : stack_level_charge_point_max_profiles_map) {
-            valid_profiles.push_back(profile);
+        if (std::find(std::begin(purposes_to_ignore), std::end(purposes_to_ignore),
```
james-ctc, line 387: I would be tempted to use std::any_of since access to the found element isn't needed.

james-ctc (approved):

hikinggrass (approved):

Issue #787, opened 2024-09-12
Rename charging profile callbacks


Pietfried:
### OCPP Version

OCPP1.6, OCPP2.0.1

### Describe the problem

Looking at how the `signal_set_charging_profiles_callback` and also `set_charging_profiles_callback` (of ocpp201) I would suggest to change the name of these variables to better represent their purpose. A libocpp user should get notified in case profiles are added / removed which leads to a change of the composite schedule. This can happen when charging profiles are added by `SetChargingProfile.req` but also when transactions stop (because TxProfiles) might be removed.

### Describe your solution

 A more appropriate name would e.g. be `charging_profiles_updated_callback`.

thenaserov:
Rename the callbacks to charging_profiles_updated_callback. This name better reflects the actual purpose of the callback, which is to notify the user whenever the charging profiles are updated, regardless of the source of the update.

Rename the Callbacks:
Replace signal_set_charging_profiles_callback with charging_profiles_updated_callback.
Replace set_charging_profiles_callback with charging_profiles_updated_callback.

Update Documentation:
Ensure that the documentation clearly explains that the charging_profiles_updated_callback is triggered whenever there is a change in the charging profiles, including additions, removals, or modifications.

Backward Compatibility:
If the library is already in use, consider deprecating the old names and providing a transition period where both names are supported. This will give users time to update their code.

Testing:
After renaming, thoroughly test the library to ensure that the callback is triggered correctly in all scenarios where charging profiles are updated.

PR #788, opened 2024-09-12
Notify consumer about charging profile update when a transaction stops

Pietfried:
## Describe your changes
This PR adds executing `signal_set_charging_profiles_callback` when a transaction is stopped to notify the consumer. The stopped transaction might have removed previously reported TxProfiles (as part of the composite schedule) which could influence the load balancing / charging power of the EVSEs.

Remark: Looking at how the `signal_set_charging_profiles_callback` and also `set_charging_profiles_callback` (of ocpp201) I would suggest to change the name of these variables to better represent their purpose. A libocpp user should get notified in case profiles are added / removed which leads to a change of the composite schedule. This can happen when charging profiles are added by `SetChargingProfile.req` but also when transactions stop (because TxProfiles) might be removed. A more appropriate name would e.g. be `charging_profiles_updated_callback`.

I covered that in [this issue](https://github.com/EVerest/libocpp/issues/787).

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -2340,6 +2340,11 @@ void ChargePointImpl::handleClearChargingProfileRequest(ocpp::Call<ClearCharging
 
     ocpp::CallResult<ClearChargingProfileResponse> call_result(response, call.uniqueId);
     this->send<ClearChargingProfileResponse>(call_result);
+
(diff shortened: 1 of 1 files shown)


maaikez (approved):
It would be nice if the callback was called only when the charging profile is indeed changed. You said you would make a ticket for it, so for now it's ok I think.

maaikez:
looks good.

hikinggrass (approved):

PR #790, opened 2024-09-12
OCPP 2.0.1: Refactor Smart Charging Persistence

### LINKED ISSUE #713: OCPP 2.0.1 - Use database for storage of ChargingProfiles
### OCPP Version

OCPP2.0.1

### Describe the problem

During the development of v2.0.1 - K01 SmartCharging we followed the pattern established in v1.6 where valid profiles were stored in both an in-memory data structure and a sqlite database. After this the database is only used when the system power cycles to read the profile from the database back into memory. This was useful during the early stages of development. Looking at this now it seems to add complexity to the code to keep the two storage mechanisms in sync and duplicates data amongst memory and hdd space.

### Describe your solution

Remove the in-memory data structure and read and write to the database for all operations.

### Additional context

_No response_

christopher-davis-afs:
## Describe your changes

Reworks how we handle the storage and loading of charging profiles, making the database the sole source of truth. 

* Adds the ChargingLimitSource to the database, allowing us to store and retrieve the source that we received a profile from
* Fixes the storage of the ChargingProfilePurpose in the database
* Removes our in-memory map of profiles, and makes us store and retrieve profiles using only the database
* Adds the transaction ID of profiles to the database
* Adds functions to the database handler to make querying easier without many super-specific functions
* Query the database for cases where we were previously iterating over all profiles
* Re-implement K09 and K10 using the database, fixing a bug with how we handled the ChargingLimitSource
* Clear invalid profiles when rebooting

Partially includes https://github.com/EVerest/libocpp/pull/782 in order to resolve an issue with clashing test fixture names that resulted in a segfault. We also include a change to the list of dependencies to rename the `json_schema_validator` dep, as using the wrong username generated some confusion.

Resolves #713 for v201


christopher-davis-afs:
@shankari Ready for review :)

include/ocpp/v201/database_handler.hpp
```diff
@@ -172,16 +172,37 @@ class DatabaseHandler : public common::DatabaseHandlerCommon {
     /// charging profiles
 
     /// \brief Inserts or updates the given \p profile to CHARGING_PROFILES table
-    void insert_or_update_charging_profile(const int evse_id, const v201::ChargingProfile& profile);
+    void insert_or_update_charging_profile(
+        const int evse_id, const v201::ChargingProfile& profile,
+        const ChargingLimitSourceEnum charging_limit_source = ChargingLimitSourceEnum::CSO);
```
shankari, line 177: @marcemmers @Pietfried where did these enums come from? I see that they were added in https://github.com/EVerest/libocpp/commit/398f0c5e3465d49a27524d708a8e8281461e69fa but I don't see these particular strings defined in the OCPP spec.

If they are, can you point me to where they are defined?
If not, can you clarify where the enums are intended to be used? For example, I don't understand the difference between "CSO" and "SO", or what "Other" will be used for

shankari (approved):
I have not reviewed the DB handler changes in detail - I have just skimmed over them. These are the low-level non-blocking changes. Please see my subsequent review about high level issues.

shankari (changes requested):
Oh wait, I forgot that this was the in-memory DB change, for which I have lots of high-level comments. I had noted them when I looked at the high level, but got forgot when I came back to finish the review and got caught up in the minutia.

> ... where valid profiles were stored in both an in-memory data structure and a sqlite database. After this the database is only used when the system power cycles to read the profile from the database back into memory. This was useful during the early stages of development. Looking at this now it seems to add complexity to the code to keep the two storage mechanisms in sync and duplicates data amongst memory and hdd space.

I understand the benefit of the change in terms of simiplicity and maintainability, but there are also potential cons in terms of performance since I/O is slower than in-memory accesses. Have you done a performance evaluation of this change
? Is it on the critical path during handshakes (I assume yes, at least for the ISO handshake)? What, if anything, is the impact of this change on the timing of
the handshake, particularly since the station is designed to be run as an embedded systems?

Please summarize the design discussions around this decision

> Partially includes #782 in order to resolve an issue with clashing test fixture names that resulted in a segfault.

Please link to the related issue for the record.

lib/ocpp/v201/database_handler.cpp
```diff
@@ -733,26 +743,213 @@ void DatabaseHandler::insert_or_update_charging_profile(const int evse_id, const
...
     }
+
+    return stmt->changes() > 0;
+}
+
+bool DatabaseHandler::clear_charging_profiles() {
+    return this->database->clear_table("CHARGING_PROFILES");
+}
+
+bool DatabaseHandler::clear_charging_profiles_matching_criteria(const std::optional<int32_t> profile_id,
```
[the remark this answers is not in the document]
christopher-davis-afs: The criteria for `get_charging_profiles_matching_criteria ()` and `clear_charging_profiles_matching_criteria()` are differnt types with different requirements.

lib/ocpp/v201/database_handler.cpp
```diff
@@ -733,26 +743,213 @@ void DatabaseHandler::insert_or_update_charging_profile(const int evse_id, const
...
+            stmt->column_int(0),                                                    // EVSE ID
+            json::parse(stmt->column_text(1)),                                      // profile
+            conversions::string_to_charging_limit_source_enum(stmt->column_text(2)) // source
+        });
+    }
+
+    return results;
+}
+
+std::vector<v201::ChargingProfile> DatabaseHandler::get_charging_profiles_for_evse(const int evse_id) {
```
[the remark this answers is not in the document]
christopher-davis-afs: This is a helper function for getting all of the profiles for an EVSE, and is not related to K09 or K10 which deal with profiles matching a set of criteria.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
```diff
@@ -1184,6 +1187,39 @@ TEST_F(ChargepointTestFixtureV201,
...
+
+    auto profile = create_charging_profile(
+        DEFAULT_PROFILE_ID, ChargingProfilePurposeEnum::TxProfile,
+        create_charge_schedule(ChargingRateUnitEnum::A, periods, ocpp::DateTime("2024-01-17T17:00:00")), DEFAULT_TX_ID);
+
+    auto response = handler.validate_and_add_profile(profile, DEFAULT_EVSE_ID, charging_limit_source);
+    EXPECT_THAT(response.status, testing::Eq(ChargingProfileStatusEnum::Accepted));
+
+    auto sut = database_handler->get_charging_limit_source_for_profile(DEFAULT_PROFILE_ID);
+    EXPECT_THAT(sut, ChargingLimitSourceEnum::SO);
```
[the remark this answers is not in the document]
christopher-davis-afs: In the database handler tests we test that the different sources are stored; CSO is the default, so this higher level test on the SmartChargingHandler tests that when we pass in a different source to `validate_and_add_profile()`, we get that same source back out.

include/ocpp/v201/database_handler.hpp
christopher-davis-afs: `ChargingLimitSourceEnumType` is defined in the OCPP-201_part2_specification_edition2.pdf file, on page 409 (3.11 ChargingLimitSourceEnumType)

dependencies.yaml
```diff
@@ -3,7 +3,7 @@ nlohmann_json:
   git: https://github.com/nlohmann/json
   git_tag: v3.11.2
   options: ["JSON_BuildTests OFF", "JSON_MultipleHeaders ON"]
-nlohmann_json_schema_validator:
+pboettch_json_schema_validator:
```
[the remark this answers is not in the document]
christopher-davis-afs: The URL hadn't changed - I don't think there was ever a `json_schema_validator` under nlohmann, the tag was just mislabeled.

christopher-davis-afs:
We have not done a full performance evaluation, but we discussed this with the community on a previous working group call, and asked about performance as a reason for the existing cache. The consensus we came to was that performance should not be an issue. We've asked for the maintainers to provide additional input on this here.

> Please link to the related issue for the record.

By issue I did not mean GitHub issue. Rather, a problem occurred while trying to rebase this branch that caused a segfault specific to this branch. Applying the changes from #782 resolved the problem.

marcemmers:
Glad to see we are simplifying this and moving to a single storage location!

I am on the road and did not do a full review of the code yet. One thing I would ask is to combine the migration files into a single file. There is not going to be a difference functionality wise but it just keeps it a bit cleaner in my opinion.

lib/ocpp/v201/database_handler.cpp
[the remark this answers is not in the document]
shankari: Can you articulate what these differences are?

For example:
- I look at the implementation of `clear_charging_profiles_matching_criteria` and it checks for evse_id, stack level, chargingProfilePurpose....
- I look at the implementation of `get_charging_profiles_matching_criteria` and it checks for evse_id, charging_profile_id, charging_profile_purpose,...

Maybe this is just a refactor to an internal method, but it seems like we should be able to avoid repeating checks. I am willing to make this be a non-blocking comment, but given that this is already a cleanup change, want to see if we can just handle this now...

shankari:
I am not able to resolve comments, so I have added 👍 to comments that I agree with. There is only one left, and I am willing to make it non-blocking if it is complex to fix.

lib/ocpp/v201/database_handler.cpp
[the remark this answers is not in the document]
christopher-davis-afs: `get_charging_profiles_matching_criteria()` takes in an optional `evse_id` and `ChargingProfileCriterion` type. `ChargingProfileCriterion` expects a list of profile IDs, the charging limit source, the purpose, and the stackLevel. The function's return type is also very specific to the use case of K09, returning a tuple of the evse_id, profile, and the source.

`clear_charging_profiles_matching_criteria()` takes in an optional `profile_id` and an optional `ClearChargingProfile` type. This type expects an optional evse ID, purpose, and stackLevel. We would need to convert the types we receive in here to what `get_charging_profiles_matching_criteria()` expects.

We do not need the extra information that `get` provides in clear, we just want to delete the matching profiles.

In addition, one of the goals of this change set was to avoid iteration and extra queries where possible. This proposed change would mean allocating a list of profiles unnecessarily, iterating over it, and then for each profile creating a new delete query instead of deleting all profiles in one query without iteration.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -4041,17 +4042,15 @@ void ChargePoint::execute_change_availability_request(ChangeAvailabilityRequest
 }
 
 // K01.27 - load profiles from database
```
marcemmers, line 4044: Comment is outdated

lib/ocpp/v201/charge_point.cpp
```diff
@@ -209,8 +209,8 @@ void ChargePoint::start(BootReasonEnum bootreason) {
     // get transaction messages from db (if there are any) so they can be sent again.
     this->message_queue->get_persisted_messages_from_db();
     this->boot_notification_req(bootreason);
-    // K01.27 - call load_charging_profiles when system boots
-    this->load_charging_profiles();
+    // call clear_invalid_charging_profiles when system boots
```
marcemmers, line 212: This comment is only describing what is happening. It's more useful if you add the why, otherwise I think its better to remove the comment and let the code do the talking ;)

include/ocpp/v201/database_handler.hpp
```diff
@@ -178,16 +178,40 @@ class DatabaseHandler : public common::DatabaseHandlerCommon {
...
     /// \brief Deletes all profiles from table CHARGING_PROFILES
-    void clear_charging_profiles();
+    bool clear_charging_profiles();
+
+    /// \brief Deletes all profiles from table CHARGING_PROFILES matching \p profile_id or \p criteria
+    bool clear_charging_profiles_matching_criteria(const std::optional<int32_t> profile_id,
+                                                   const std::optional<ClearChargingProfile>& criteria);
+
+    /// \brief Get all profiles from table CHARGING_PROFILES matching \p profile_id or \p criteria
+    std::vector<std::tuple<int, v201::ChargingProfile, ChargingLimitSourceEnum>>
```
marcemmers, line 199: It might be better to make a helper struct for this return value so that you can actually name the int for example.

include/ocpp/v201/database_handler.hpp
```diff
@@ -178,16 +178,40 @@ class DatabaseHandler : public common::DatabaseHandlerCommon {
...
+    /// \brief Retrieves the charging profiles stored on \p evse_id
+    std::vector<v201::ChargingProfile> get_charging_profiles_for_evse(const int evse_id);
 
     /// \brief Retrieves all ChargingProfiles
+    std::vector<v201::ChargingProfile> get_all_charging_profiles();
+
+    /// \brief Retrieves all ChargingProfiles grouped by EVSE ID
     virtual std::map<int32_t, std::vector<v201::ChargingProfile>> get_all_charging_profiles_group_by_evse();
+
+    ChargingLimitSourceEnum get_charging_limit_source_for_profile(const int profile_id);
```
marcemmers, line 212: I haven't seen the use of this yet, but it seems this could already be returned with the actual profile when we retrieve that right?

include/ocpp/v201/database_handler.hpp
```diff
@@ -178,16 +178,40 @@ class DatabaseHandler : public common::DatabaseHandlerCommon {
...
+
+    std::unique_ptr<common::SQLiteStatementInterface> new_statement(const std::string sql);
```
marcemmers, line 214: Can this one be removed again? Does not seem to be used a lot if at all.

marcemmers (changes requested):
I can't finish the review today so here are some comments already. Still have to look at the database implementation, smart charging and the unit tests.

lib/ocpp/v201/charge_point.cpp
christopher-davis-afs: Removed this and the other comment you mentioned

include/ocpp/v201/database_handler.hpp
christopher-davis-afs: We do; This is a leftover from an earlier stage in this set of commits. Removed it.

include/ocpp/v201/database_handler.hpp
christopher-davis-afs: We use this quite a bit currently within SmartChargingHandler. There's no easy way to remove it without having many very specific functions on the database handler. Long-term it would be better to use the repository pattern for the database handler, which would mean having a class that wraps the handler and provides operations that are specific to, for example, the charging profiles table.

lib/ocpp/v201/database_handler.cpp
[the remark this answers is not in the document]
shankari: The first argument is not very convincing to me, but the second

> In addition, one of the goals of this change set was to avoid iteration and extra queries where possible. This proposed change would mean allocating a list of profiles unnecessarily, iterating over it, and then for each profile creating a new delete query instead of deleting all profiles in one query without iteration.

is a pretty good argument

LGTM!

shankari (approved):

include/ocpp/v201/database_handler.hpp
marcemmers: I haven't heard of the repository pattern before. If I google it it seems like this class is already kind of doing that, except for the fact that this does the operations for all tables. Instead using the repository pattern we would make a class for each table (or logical grouping thereof), is that correct?

I missed the use in SmartChargingHandler btw so this is fine to leave in for now.

include/ocpp/v201/database_handler.hpp
marcemmers: That would improve it a bit, but I still think a struct is easier to read and maintain. Is there a reason you would like to stick to the tuple?

christopher-davis-afs: Mainly: It seems odd to add a struct for one specific use case.

It seems I still left the ReportedChargingProfiles struct in tact, so I'll re-use that.

include/ocpp/v201/database_handler.hpp
christopher-davis-afs: That would be correct, yes

include/ocpp/v201/database_handler.hpp
christopher-davis-afs: Err, that might need to be moved somewhere else because of a recursive import. Where would be a good place to put that struct?

marcemmers: We are probably having some other issues when we are running into recursive imports. But that's way out of scope here. Maybe `include/ocpp/v201/types.hpp` is a good spot for it?

christopher-davis-afs: Moved ReportedChargingProfiles and updated the rest of the code to use it :)

christopher-davis-afs:
> One thing I would ask is to combine the migration files into a single file.

Done :)

marcemmers (approved):

marcemmers:
Ah I missed those 2 codacy results. There are 2 strings passed by value that should be passed by reference. If you fix those and let me know, I will merge this ;)

christopher-davis-afs:
@marcemmers fixed those :)

Issue #721, opened 2024-07-31
Uniform function for splitting a string


Pietfried:
### OCPP Version

OCPP1.6, OCPP2.0.1

### Describe the problem

Currently there are two functions to split a string in libocpp:
`get_vector_from_csv` and `split_string`. 

### Describe your solution

The goal of this issue is to remove `get_vector_from_csv` and its occurences and replace it with `split_string`.

See: https://github.com/EVerest/libocpp/pull/707#discussion_r1695188396

PR #791, opened 2024-09-12
refactor: Replace get_vector_from_csv with split_string

mlitre:
Removed references to get_vector_from_csv by split_string since this last method is more extensible / reusable.

Issue #721 

### DIFF
--- a/include/ocpp/common/utils.hpp
+++ b/include/ocpp/common/utils.hpp
@@ -12,8 +12,6 @@ namespace ocpp {
 /// \brief Case insensitive compare for a case insensitive (Ci)String
 bool iequals(const std::string& lhs, const std::string rhs);
 
-std::vector<std::string> get_vector_from_csv(const std::string& csv_str);
-
 bool is_integer(const std::string& value);
 std::tuple<bool, int> is_positive_integer(const std::string& value);
 bool is_decimal_number(const std::string& value);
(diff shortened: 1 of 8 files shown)


maaikez (commented):
Maybe you can also remove the function completely? Or is it still useful?

mlitre:
> Maybe you can also remove the function completely? Or is it still useful?

I did not remove to avoid a breaking change. If I'm not mistaken, I saw that this was a public utility so any lib user would get issues if they were using the old method. I do agree that it would be simpler to just remove the function. If that is preferred I can go ahead and do that.

maaikez:
> > Maybe you can also remove the function completely? Or is it still useful?
> 
> I did not remove to avoid a breaking change. If I'm not mistaken, I saw that this was a public utility so any lib user would get issues if they were using the old method. I do agree that it would be simpler to just remove the function. If that is preferred I can go ahead and do that.

Yes I understand what you say. 
I'm thinking then we can never remove such functions, and for this particular case, there is a replacement. So I would opt for removing it, otherwise other people will use it later again. 

What do you think?

mlitre:
@maaikez fair enough it seems to me to be a pretty low impact change. I'll go ahead and remove the function.

mlitre:
@maaikez I've gone ahead and removed the method. Let me know if there are other things I missed in terms of code or process for merging changes / contributing. Thanks!

maaikez:
It looks good. Did you somehow check if things still work (except from the unittests)?

mlitre:
> It looks good. Did you somehow check if things still work (except from the unittests)?

I have not done much extra testing to be honest, but if necessary I can run some.

PR #795, opened 2024-09-16
Add critical and timestamp parameters to on_security_event functions

hikinggrass:
This adds "critical" and "timestamp" parameters to the on_security_event functions for OCPP 1.6 and 2.0.1

Criticality is by default determined based on the 1.6 security whitepaper, but can be overwritten externally

By default the current datetime is used for the timestamp, but this can also be overwritten externally (for example if the event occured earlier)


maaikez (approved):
Looks good

PR #801, opened 2024-09-25
Fix missing forward slash in URI if not filled in network profile

marcemmers:
## Describe your changes

If the slash was missing in the CSMS URL we would just append the identity without the slash. This makes sure we add the slash if needed. Also adds unit tests for this function.


maaikez (approved):
Ah nice, I also encountered this at the plugfest for 1.6 but it looks like it is solved for everything with this pull request. Thanks!

hikinggrass (approved):

PR #817, opened 2024-09-30
Remote start and stop could fail in core, we should be able to report that

marcemmers:
## Describe your changes
Our implementation could result in a failure to start or stop and there was no option to report that to the CSMS. With this change we force the core to provide a result. It will still only get called if libocpp thinks it is accepted.

### DIFF
--- a/include/ocpp/v201/charge_point_callbacks.hpp
+++ b/include/ocpp/v201/charge_point_callbacks.hpp
@@ -33,7 +33,8 @@ struct Callbacks {
     std::function<bool(const std::optional<const int32_t> evse_id, const ResetEnum& reset_type)>
         is_reset_allowed_callback;
     std::function<void(const std::optional<const int32_t> evse_id, const ResetEnum& reset_type)> reset_callback;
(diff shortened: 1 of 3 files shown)


Pietfried (approved):
We have only discussed but not taken any action yet to include `statusInfo` information more in all kinds of responses. For this case this probably results in a different return type `RequestStartTransactionResponse` / `RequestStoptTransactionResponse`. 

For this case you can add it as part of this PR or we do this in a subsequent PR (which we need anyways), in which we address other callbacks and internal derivation of statusInfo if possible. I'm approving this and leave that up to you.

marcemmers:
> We have only discussed but not taken any action yet to include `statusInfo` information more in all kinds of responses. For this case this probably results in a different return type `RequestStartTransactionResponse` / `RequestStoptTransactionResponse`.
> 
> For this case you can add it as part of this PR or we do this in a subsequent PR (which we need anyways), in which we address other callbacks and internal derivation of statusInfo if possible. I'm approving this and leave that up to you.

I'd rather first determine what we are going to do exactly and then change these again to the final result. Otherwise chances are high that it needs breaking changes in the future again anyway.

PR #822, opened 2024-10-06
Handle of invalid datetime when format check fails

hikinggrass:


lib/ocpp/common/types.cpp
```diff
@@ -53,8 +43,7 @@ void DateTimeImpl::from_rfc3339(const std::string& timepoint_str) {
             in.seekg(0);
             in >> date::parse("%FT%T", this->timepoint);
             if (in.fail()) {
-                EVLOG_error << "Timepoint string parsing failed. Could not convert: \"" << timepoint_str
-                            << "\" into DateTime.";
+                throw TimePointParseException(timepoint_str);
```
maaikez, line 46: Hmm now you added a throw where the software did not crash in the past. Did you check everywhere this is used and provide a try / catch around it if needed?

jannejy: Well, the `std::string` constructor calls only when json is converted to the `DateTime` type or in database conversions. In other places the constructor is not (should not be) called. So then it should be caught either by `TimePointParseException` or `std::exception`, that is done. From my side and for 2.0.1, I don't see the missing places. @hikinggrass can approve as well for 1.6

include/ocpp/common/types.hpp
```diff
@@ -33,6 +33,23 @@ struct Message {
...
 };
 
+/// \brief Exception used when DateTime class is initialized by invalid timepoint string.
+class TimePointParseException : public std::exception {
+public:
+    explicit TimePointParseException(const std::string& timepoint_str) :
+        msg{"Timepoint string parsing failed. Could not convert: \"" + timepoint_str + "\" into DateTime."} {
+    }
+
+    ~TimePointParseException() noexcept override = default;
```
marcemmers, line 43: Don't think we need to explicitly define this if it is a default destructor

jannejy: Just copied from other exception classes here, but indeed, the compiler will generate the same destructor for us

maaikez (approved):

PR #827, opened 2024-10-08
Fix reconnect mechanism when basic auth password is set

jannejy:


maaikez:
Looks good, but I am not sure if this has other implications. @hikinggrass what do you think?

marcemmers:
@maaikez To me it just looks like we are now actually passing on the argument as it was intended. Looking at the call flow there should be no other implications.

maaikez (approved):

hikinggrass (approved):

PR #831, opened 2024-10-11
K01 ChargePoint refactorings

gberardi-pillar:
This feature branch includes work that updates the ChargePoint constructor to take references instead of shared pointers as parameters, as well as provides a MessageQueueInterface to allow for the creation of a MessageQueueMock, which will make it possible to run tests faster as well as write and maintain tests more easily. 

### DIFF
--- a/include/ocpp/common/message_queue.hpp
+++ b/include/ocpp/common/message_queue.hpp
@@ -164,8 +164,57 @@ bool allowed_to_send_message(const ControlMessage<M>& message, const DateTime& t
     return true;
 }
 
+template <typename M> class MessageQueueInterface {
+public:
+    virtual ~MessageQueueInterface() {
+    }
+    virtual void start() = 0;
+    virtual void reset_next_message_to_send() = 0;
+    virtual void get_persisted_messages_from_db(bool ignore_security_event_notifications = false) = 0;
+
+    /// \brief pushes a new \p call message onto the message queue
+    template <class T> void push(Call<T> call, const bool stall_until_accepted = false) {
+        json call_json = call;
+        this->push(call_json, stall_until_accepted);
+    }
+    virtual void push(const json& message, const bool stall_until_accepted = false) = 0;
+    template <class T> void push(CallResult<T> call_result) {
+        json call_result_json = call_result;
+        this->push_call_result(call_result_json, call_result.uniqueId);
+    }
+    virtual void push_call_result(const json& call_result_json, const MessageId& unique_id) = 0;
+    virtual void push(CallError call_error) = 0;
+    template <class T> std::future<EnhancedMessage<M>> push_async(Call<T> call) {
+        auto message = std::make_shared<ControlMessage<M>>(call);
+        return push_async_internal(message);
+    }
+    virtual std::future<EnhancedMessage<M>> push_async_internal(std::shared_ptr<ControlMessage<M>> message) = 0;
+    virtual EnhancedMessage<M> receive(std::string_view message) = 0;
+    virtual void reset_in_flight() = 0;
+    virtual void handle_call_result(EnhancedMessage<M>& enhanced_message) = 0;
+    virtual void handle_timeout_or_callerror(const std::optional<EnhancedMessage<M>>& enhanced_message_opt) = 0;
+    virtual void stop() = 0;
+    virtual void pause() = 0;
+    virtual void resume(std::chrono::seconds delay_on_reconnect) = 0;
+    virtual void set_registration_status_accepted() = 0;
+    virtual bool is_transaction_message_queue_empty() = 0;
+    virtual bool contains_transaction_messages(const CiString<36> transaction_id) = 0;
+    virtual bool contains_stop_transaction_message(const int32_t transaction_id) = 0;
+    virtual void update_transaction_message_attempts(const int transaction_message_attempts) = 0;
+    virtual void update_transaction_message_retry_interval(const int transaction_message_retry_interval) = 0;
+    virtual void update_message_timeout(const int timeout) = 0;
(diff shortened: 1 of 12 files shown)


include/ocpp/common/message_queue.hpp
```diff
@@ -661,16 +701,16 @@ template <typename M> class MessageQueue {
     }
 
     /// \brief Sends a new \p call_result message over the websocket
-    template <class T> void push(CallResult<T> call_result) {
+    void push_call_result(const json& call_result_json, const MessageId& unique_id) override {
```
shankari, line 704: so we don't need this to be a template any more because we just separate out the structure?

include/ocpp/common/message_queue.hpp
```diff
@@ -966,77 +1004,59 @@ template <typename M> class MessageQueue {
...
-    bool contains_transaction_messages(const CiString<36> transaction_id) {
-        std::lock_guard<std::recursive_mutex> lk(this->message_mutex);
-        for (const auto control_message : this->transaction_message_queue) {
-            if (control_message->messageType == v201::MessageType::TransactionEvent) {
-                v201::TransactionEventRequest req = control_message->message.at(CALL_PAYLOAD);
-                if (req.transactionInfo.transactionId == transaction_id) {
-                    return true;
-                }
-            }
-        }
```
shankari, line 991: why is this removed?
The related change is
https://github.com/US-JOET/libocpp/commit/b72a6f601f62c8cdb80f6ffdd74f613ad12da4b4

which does not have any explanations

include/ocpp/common/message_queue.hpp
```diff
@@ -966,77 +1004,59 @@ template <typename M> class MessageQueue {
...
-    bool contains_stop_transaction_message(const int32_t transaction_id) {
-        std::lock_guard<std::recursive_mutex> lk(this->message_mutex);
-        for (const auto control_message : this->transaction_message_queue) {
-            if (control_message->messageType == v16::MessageType::StopTransaction) {
-                v16::StopTransactionRequest req = control_message->message.at(CALL_PAYLOAD);
-                if (req.transactionId == transaction_id) {
-                    return true;
-                }
-            }
-        }
```
shankari, line 1004: Ditto

tests/lib/ocpp/v201/test_composite_schedule.cpp
shankari, line 1: The changes in this file don't appear to be related to the message queue.
Could you add a summary of the related changes?

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
shankari, line 1: Ditto. This doesn't seem to be related to the message queue at all, and I don't understand the motivation for the change.

shankari (changes requested):
Again, a few high-level questions, for the record, that should be easy to answer.

include/ocpp/common/message_queue.hpp
gberardi-pillar: It was not removed. It was moved into the v201 MessageQueue implementation (see lib/ocpp/v201/message_queue.cpp) since this logic is specific to v201::MessageType. Similarly with the v16::MessageType implementation which is in lib/ocpp/v16/message_queue.cpp.

tests/lib/ocpp/v201/test_composite_schedule.cpp
gberardi-pillar: I'm not sure which changes you are referring to? 

The changes in this file are all related to the MessageQueue, specifically using a MessageQueueMock in tests which is what these changes make possible.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
gberardi-pillar: The other changes in this PR are related to making it possible to pass in parameters, such as the database_handler, to ChargePoint as references rather than as pointers.

include/ocpp/common/message_queue.hpp
gberardi-pillar: The template part is now in the interface, which delegates to the derived class implementation, which does not need a template.

Basically, in C++ you cannot have a pure virtual function that is templated, yet we wanted a MessageQueueInterface to allow for the ability to mock it in tests. So some of the logic was moved into the interface template functions in order to create and pass the non-templated parameters to the derived implementation.

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
shankari: In that case, it would have been helpful to flag that this PR contains two sets of unrelated changes 😄  and include the motivation for the second set of changes as well.

tests/lib/ocpp/v201/test_composite_schedule.cpp
shankari: These changes
https://github.com/EVerest/libocpp/pull/831/files#diff-a117871b4a0e09ac30845bccb5123506c8c180dfa71f9c8abc8c61ae374dcdcbR207

which are clearly not related to the `MessageQueue`

tests/lib/ocpp/v201/test_smart_charging_handler.cpp
gberardi-pillar: The PR description says "This feature branch includes work that updates the ChargePoint constructor to take references instead of shared pointers as parameters, as well as provides a MessageQueueInterface to allow for the creation of a MessageQueueMock,"

Is there something more I should have said to indicate that this includes two separate changes in a single PR?

shankari (approved):

hikinggrass:
Closed in favor of #849 which will refactor the message dispatching. We will include parts of this PR either there or in another PR

Issue #387, opened 2023-12-28
Replace WebSocket++ with libwebsockets


hikinggrass:
A gradual switch from WebSocket++ to libwebsockets has begun in the last weeks of 2023. This will allow us to implement a few additional features that proved difficult to realize with the current implementation. Since this touches an essential part of the libocpp library, the switch will be done gradually and carefully. First a proof-of-concept implementation for TPM integration was realized, this will be followed by additional implementations for the plain and TLS websocket functionality. These will exist as a opt-in compile-time option at first, to facilitate testing. Once the new implementations have been sufficiently tested it will become the default and the old implementation will be deprecated and then removed at a later date.

- [x] Add initial libwebsockets based implementation for TLS with TPM support as a proving ground #320 
- [x] Make libwebsockets integration compile time selectable
- [x] Refactor websocket interface to hide WebSocket++ implementation details
- [x] Implement plain and TLS websocket functionality with libwebsockets
- [x] Provide compile time selection between WebSocket++ and libwebsockets backend
- [x] Test and verify that both implementations behave the same
- [x] Deprecate the WebSocket++ backend
- [x] Remove WebSocket++ as dependency

Pietfried:
@AssemblyJohn please finish this up by removing websocketpp from libocpp. Please also rename the class `WebsocketTlsTPM` since it also covers non tls/tpm connections.

AssemblyJohn:
Solved with: https://github.com/EVerest/libocpp/pull/846

PR #846, opened 2024-10-24
Removed websocket++ as a dependency

AssemblyJohn:
## Describe your changes

Removed websocket++ from the cmake files and source files.

Closes: https://github.com/EVerest/libocpp/issues/387#event-14817871734


AssemblyJohn:
I'd require some ideas on the new websocket class implementation name, since we already have a file called 'websocket.cpp' & a class called 'Websocket'.

Pietfried:
> I'd require some ideas on the new websocket class implementation name (from file 'websocket_libwebsockets.cpp' and name 'WebsocketTlsTPM'), since we already have a file called 'websocket.cpp' & a class called 'Websocket'.

The filename `websocket_libwebsockets` is fine. The class name would then be `WebsocketLibwebsockets`.

CMakeLists.txt
```diff
@@ -25,6 +25,10 @@ if((NOT LIBOCPP_ENABLE_V16) AND (NOT LIBOCPP_ENABLE_V201))
     message(FATAL_ERROR "At least one of LIBOCPP_ENABLE_V16 and LIBOCPP_ENABLE_V201 needs to be ON")
 endif()
 
+if(LIBOCPP_ENABLE_DEPRECATED_WEBSOCKETPP)
+    message(FATAL_ERROR "Websocket++ has been deprecated")
```
marcemmers, line 29: It's no longer deprecated but its removed now right?

AssemblyJohn: I was wondering, if I should even let that inside there. But I thought that if someone has a config that has that option on, maybe a error should be issued.

marcemmers: Yeah I understand keeping the fatal error, I would just change the message a bit since deprecated means you can still use it, and you can't use this anymore.

AssemblyJohn: Changed it to 'removed' from 'deprecated', should be more clear.

marcemmers:
There are still some websocketpp header files present that should be removed.

Also, can't we just get rid of the `Websocket` class itself since it is just a wrapper over `WebocketLibwebsockets` now that only adds a few log lines. Then we no longer have this question of naming either. Just name `WebocketLibwebsockets` `Websocket` and we are done.

AssemblyJohn:
> There are still some websocketpp header files present that should be removed. Both in the `include/common/websocket` and in `3rd_party`.
> 
> Also, can't we just get rid of the `Websocket` class itself since it is just a wrapper over `WebocketLibwebsockets` now that only adds a few log lines. Then we no longer have this question of naming either. Just name `WebocketLibwebsockets` `Websocket` and we are done.

Removed the unused headers, thanks for pointing it out! The 3rd_party is required for some URI validation. I don't exclude in the future to write our own code to take care of that. As for Websocket/WebsocketLibwebsockets, it's better to remain like this since we might add an alternate library instead of libwebsockets. Or at list that is how it was designed in the first place.

Pietfried (approved):

hikinggrass (approved):

PR #849, opened 2024-10-28
WIP: Refactor of message dispatching

Pietfried:
## Describe your changes

Large refactor of message handling:
* Transfered message dispatching to separate class in order to reuse it in functional blocks
* Added new class MessageDispatcherInterface that can dispatch OCPP Call, CallResult, CallError
* MessageDispatcherInterface became member of ChargePoint classes. The templated methods for sending/dispatching Call, CallResult, CallError have been moved from ChargePoint classes to MessageDispatcher
* Added example implementation of one functional block DataTransfer including two tests
* More functional blocks can follow in subsequent PRs


doc/message_dispatching.puml
```diff
@@ -0,0 +1,52 @@
+
+@startuml
+
+interface MessageDispatcherInterface {
+    +dispatch_call(const json& call, bool triggered = false)
+    +dispatch_call_async(const json& call, bool triggered = false): std::future<EnhancedMessage<T>>
+    +dispatch_call_result(const json& call_result)
+    +dispatch_call_error(const json& call_error)
+    +new_message_id(): MessageId
```
marcemmers, line 9: Do we even need to bother with message IDs for users of this layer? The layer should be able to take care of it itself right?

Pietfried: Yes, this functionality does not necessarily have to be part of this interface

marcemmers: Looking at it, I think we might want to move the `createMessageId` functionality out of the message queue, it can be a free function. Then on construction of a `Call<T>` type we can generate a message id internally and essentially hide this value.

doc/message_dispatching.puml
```diff
@@ -0,0 +1,52 @@
...
+class v201::MessageDispatcher {
+    - MessageQueue& message_queue
+    - DeviceModel& device_model
+    - ConnectivityManager& connectivity_manager
+    - RegistrationStatusEnum& registration_status
+}
+
+interface v201::DataTransferInterface {
+    +data_transfer_req(request: DataTransferRequest): std::optional<DataTransferResponse>
+    +handle_data_transfer_req(call: Call<DataTransferRequest>)
```
marcemmers, line 27: Should we replace this with a `handle_message` function that takes an `EnhancedMessage<v201::MessageType>`?

For this class there is only one message to handle but ideally that should not change the API for these functional blocks. If we later have to add a message to handle we can do it purely in the functional block

doc/message_dispatching.puml
```diff
@@ -0,0 +1,52 @@
+
+@startuml
+
+interface MessageDispatcherInterface {
```
marcemmers, line 4: This could also be named `IHandleMessage` if we really want haha

lib/ocpp/v201/charge_point.cpp
```diff
@@ -75,6 +75,11 @@ ChargePoint::ChargePoint(const std::map<int32_t, int32_t>& evse_connector_struct
     }
 
     initialize(evse_connector_structure, message_log_path);
+
+    this->message_dispatcher =
```
marcemmers, line 79: Might want to check here if `message_queue` exists because this constructor is sometimes called with nullptr.

tests/lib/ocpp/v201/functional_blocks/test_data_transfer.cpp
```diff
@@ -0,0 +1,52 @@
...
+// Test when the data_transfer_callback is not set, expecting DataTransferStatusEnum::UnknownVendorId
+TEST(DataTransferTest, HandleDataTransferReq_NoCallback) {
+    MockMessageDispatcher mock_dispatcher;
+    DataTransfer data_transfer(mock_dispatcher, std::nullopt); // No callback
+
+    DataTransferRequest request;
+    request.vendorId = "test_vendor";
+    ocpp::Call<DataTransferRequest> call(request, "unique_id_123");
+
+    EXPECT_CALL(mock_dispatcher, dispatch_call_result(_)).WillOnce(Invoke([](const json& call_result) {
```
marcemmers, line 22: I feel like this is going to be worth investing some time in making these expectations nicer. Way outside of scope for now but would be smart to do before we make more of these tests.

I don't think we really care whether `dispatch_call` or `dispatch_call_async` is called for example, as long as one of those gets called for tests where that is a thing.

Also I would like to create custom matchers to set expectations on specified fields of structs so you don't have to do the invoke construction.

tests/lib/ocpp/v201/functional_blocks/test_data_transfer.cpp
```diff
@@ -0,0 +1,52 @@
...
+        auto response = call_result[ocpp::CALLRESULT_PAYLOAD].get<DataTransferResponse>();
+        EXPECT_EQ(response.status, DataTransferStatusEnum::Accepted);
+    }));
+
+    data_transfer.handle_data_transfer_req(call);
+}
```
marcemmers, line 52: Missing tests for `data_transfer_req` in here.

marcemmers (commented):
In general I think this looks good. Some suggestions or improvements we might want to discuss.

PR #852, opened 2024-10-31
Support hashing directories of certificates

jannejy:


3rd_party/cert_rehash/c_rehash.hpp
```diff
@@ -0,0 +1,310 @@
+/* c_rehash.c - Create hash symlinks for certificates
```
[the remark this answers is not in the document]
jannejy: Also, it's worth looking together with the PR in `evse-security`: https://github.com/EVerest/libevse-security/pull/96

AssemblyJohn:
In general it looks good to me. I like the fact that there's a conditional to check if we're using a directory or a file, however see my comment in the evse-security PR: https://github.com/EVerest/libevse-security/pull/96#issuecomment-2467446037

If it handles that case in security code too, I'll have another look and approve.

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -370,11 +371,16 @@ bool WebsocketLibwebsockets::tls_init(SSL_CTX* ctx, const std::string& path_chai
     }
 
     if (this->evse_security->is_ca_certificate_installed(ocpp::CaCertificateType::CSMS)) {
-        std::string ca_csms = this->evse_security->get_verify_file(ocpp::CaCertificateType::CSMS);
+        std::string ca_csms = this->evse_security->get_verify_location(ocpp::CaCertificateType::CSMS);
```
AssemblyJohn, line 374: I see here that the `get_verify_location` is used, if based on the config this is a file, the file should be used, else the directory, extra checks have to be placed here to use the proper location/file based on the configuration.

jannejy: There is below the check if this is a directory or file. What do you mean?

AssemblyJohn: The problem with this approach is that it uses a fundamentally different approach than `get_verify_file`. The `get_verify_file` searches inside the whole directory, for a single root certificate that is valid. For example if we are using a folder 'certificates/ca/mf' and in it we have MF_ROOT1.pem and MF_ROOT2.pem, the `get_verify_file` will return the certificates/ca/mf/MF_ROOT2.pem presuming that it will be newer. On the other hand if we use `get_verify_location` it will return either certificates/ca/mf/MF_ROOT1.pem or  'certificates/ca/mf' based on the configuration if we use a directory.

The behavior is very different. 

What we need here is some flag and/or config value (I am can't remember if one already exists) that either uses the old `get_verify_file` with it's behavior or the new `get_verify_location` with it's particular behavior.

jannejy: For the file, the behavior is not changed at all, as both methods are returning a path to a file. 

If we are using a directory, we just let the openssl library choose the proper one, not based on the assumptions, which certificate is newer. For example, we have one certificate for one server, it is working, ok. We added another one for another server. If we want to connect to the first server, we won't, as this newest certificate is for another server. This way, we burden the end-user to do extra actions, such as deleting the second certificate.

Why don't we want to pass to the openssl library the choice which certificates we are using for the particular server?

AssemblyJohn: The point is valid, however I do need to double check if there can be issues when using c_rehash on some environments, or other workarounds that have been applied in places. If I can get a confirmation that those are not likely to cause issues, I'll approve, since your approach seems better.

Pietfried:
@marcemmers @jannejy 
https://github.com/EVerest/libevse-security/releases/tag/v0.9.2

AssemblyJohn (approved):
Looks good in the current state!

Pietfried (approved):

PR #854, opened 2024-10-31
Feature/327 use case h01 reservation

maaikez:
## Describe your changes
Changes to implement reservations for ocpp 2.0.1. 


include/ocpp/v201/charge_point.hpp
```diff
@@ -567,6 +579,27 @@ class ChargePoint : public ChargePointInterface, private ocpp::ChargingStationBa
...
+    /// \brief Get connector status.
+    ///
+    /// This will search if there is a connector on this evse with status 'Available'. It will search through all
+    /// connectors and return on the first connector that is 'Available'.
+    ///
+    /// \param evse_id          The evse id.
+    /// \param connector_type   The connector type.
+    /// \return std::nullopt if connector type is given, but does not exist. Otherwise the connector status.
+    ///
+    std::optional<ConnectorStatusEnum> get_available_connector_or_status(const uint32_t evse_id,
```
marcemmers, line 600: Purely looking at this function from the header it is a bit confusing. I think you are just checking if an evse has a connector available and optionally only check for a specific connector type?

If so, I would rename and change the documentation a bit. The connector_type argument is then a sort of optional filter.

include/ocpp/v201/charge_point_callbacks.hpp
```diff
@@ -151,5 +151,14 @@ struct Callbacks {
     std::optional<std::function<void(const RunningCost& running_cost, const uint32_t number_of_decimals,
                                      std::optional<std::string> currency_code)>>
         set_running_cost_callback;
+
+    /// \brief Callback function is called when a reservation request is received from the CSMS
+    std::optional<std::function<ReserveNowStatusEnum(
```
marcemmers, line 156: Any reason why you don't just forward the request like we do with the other callbacks? Saves a bunch of arguments ;)

lib/ocpp/v201/charge_point.cpp
```diff
@@ -1982,6 +1997,88 @@ void ChargePoint::set_evse_connectors_unavailable(EvseInterface& evse, bool pers
...
+ChargePoint::get_available_connector_or_status(const uint32_t evse_id, std::optional<ConnectorEnum> connector_type) {
+    EvseInterface* evse;
+    try {
+        evse = &evse_manager->get_evse(static_cast<int32_t>(evse_id));
+    } catch (const EvseOutOfRangeException&) {
+        EVLOG_error << "Evse id " << evse_id << " is not a valid evse id.";
+        return std::nullopt;
+    }
+
+    bool type_found = false;
```
marcemmers, line 2035: I would move the implementation of this to the `Evse` class, minus the evse_id argument ofcourse.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3296,6 +3393,132 @@ void ChargePoint::handle_heartbeat_response(CallResult<HeartbeatResponse> call)
...
+    ReserveNowResponse response;
+    response.status = ReserveNowStatusEnum::Rejected;
+    if (!this->callbacks.reserve_now_callback.has_value() ||
+        !this->device_model->get_optional_value<bool>(ControllerComponentVariables::ReservationCtrlrAvailable)
+             .value_or(false) ||
+        !this->device_model->get_optional_value<bool>(ControllerComponentVariables::ReservationCtrlrEnabled)
+             .value_or(false)) {
+        // Reservation not available / implemented, return 'Rejected'.
+        // H01.FR.01
+        EVLOG_info << "Receiving a reservation request, but reservation is not implemented.";
```
marcemmers, line 3406: if `ReservationCtrlrEnabled` is false it might be implemented, just not enabled.

Also, for most of these it would be nice if we can add the error to the call result as well for the CSMS to see what is actually happening.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3296,6 +3393,132 @@ void ChargePoint::handle_heartbeat_response(CallResult<HeartbeatResponse> call)
...
+    }
+
+    // Check if we need a specific evse id during a reservation and if that is the case, if we recevied an evse id.
+    const ReserveNowRequest request = call.msg;
+    if (!request.evseId.has_value() &&
+        !this->device_model->get_optional_value<bool>(ControllerComponentVariables::ReservationCtrlrNonEvseSpecific)
+             .value_or(false)) {
+        // H01.FR.19
+        EVLOG_warning
+            << "Trying to make a reservation, but no evse id was given while it should be sent in the request.";
```
marcemmers, line 3419: + `when ReservationCtrlrNonEvseSpecific is disabled`

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3296,6 +3393,132 @@ void ChargePoint::handle_heartbeat_response(CallResult<HeartbeatResponse> call)
...
+        this->send<ReserveNowResponse>(call_result);
+        return;
+    }
+
+    const std::optional<int32_t> evse_id = request.evseId;
+
+    ConnectorStatusEnum connector_status = ConnectorStatusEnum::Unavailable;
+
+    if (evse_id.has_value()) {
+        if (evse_id.value() < 0 || !evse_manager->does_evse_exist(evse_id.value())) {
```
marcemmers, line 3430: I see I missed some checks in the evse manager. The smaller than 0 check should definitely be there (also in the `get_evse` functions. Then `evse_id.value() < 0` would not be necessary anymore

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3296,6 +3393,132 @@ void ChargePoint::handle_heartbeat_response(CallResult<HeartbeatResponse> call)
...
+        EVLOG_warning
+            << "Trying to make a reservation, but no evse id was given while it should be sent in the request.";
+        const ocpp::CallResult<ReserveNowResponse> call_result(response, call.uniqueId);
+        this->send<ReserveNowResponse>(call_result);
+        return;
+    }
+
+    const std::optional<int32_t> evse_id = request.evseId;
+
+    ConnectorStatusEnum connector_status = ConnectorStatusEnum::Unavailable;
```
marcemmers, line 3427: You don't need to store this since you never use it anywhere

include/ocpp/v201/charge_point.hpp
maaikez: Ok I changed it a bit, can you take a look if that's better (have to commit still).

include/ocpp/v201/charge_point_callbacks.hpp
maaikez: Hmm I kept it quite the same as in OCPP 1.6. I don't know why the choice there was to do it this way. Better change it then?

marcemmers: I would prefer to keep the interface for the 2.0.1 implementation consistent over certain functionality shared between 1.6 and 2.0.1 so yeah I would change it. Just my 2 cents ofcourse.

lib/ocpp/v201/charge_point.cpp
maaikez: It is just shorter and better readable in my opinion.

include/ocpp/v201/charge_point_callbacks.hpp
maaikez: Well it is consistent with 1.6. 

@Pietfried  do you have a preference?

lib/ocpp/v201/charge_point.cpp
marcemmers: I'm not sure I follow. You set the `connector_status` value here and in other places, but you never read it except for in line 3407. Where you always set the value the line before so this variable could be limited to that if statement.

maaikez: Ah thought it was about evse_id.

You're right.

include/ocpp/v201/charge_point_callbacks.hpp
maaikez: Talked to Piet, I will change it to the Reservation type (give me some time :)).

maaikez: done!

include/ocpp/v201/charge_point.hpp
```diff
@@ -711,6 +737,11 @@ class ChargePoint : public ChargePointInterface, private ocpp::ChargingStationBa
     void handle_change_availability_req(Call<ChangeAvailabilityRequest> call);
     void handle_heartbeat_response(CallResult<HeartbeatResponse> call);
 
+    // Function Block H: Reservations
+    void handle_reserve_now_request(Call<ReserveNowRequest> call);
+    void handle_cancel_reservation_callback(Call<CancelReservationRequest> call);
+    void send_reserve_now_rejected_response(const MessageId& unique_id, const std::string& status_info);
```
marcemmers, line 743: I see we haven't used this anywhere yet in ChargePoint but the status_info could be a string_view. You can change it if you want but it's fine either way.

marcemmers (approved):

include/ocpp/v201/charge_point.hpp
Pietfried: This also looks a little bit confusing to me. Looking at how this function is used internally, it looks like `bool is_connector_available(const int32_t evse_id, std::optional<ConnectorEnum> connector_type)` provides everything you need. This would search for an available connector at the given `evse_id` using the given `connector_type` filter. Apart from that it looks like we need a function `bool does_connector_exist(const int32_t evse_id, const ConnectorEnum connector_type)` to check if the given `connector_type` belongs to any evse. I will mark the places in `charge_point.cpp` where I think these function shall be used.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3296,6 +3324,145 @@ void ChargePoint::handle_heartbeat_response(CallResult<HeartbeatResponse> call)
...
+
+    if (evse_id.has_value()) {
+        if (!evse_manager->does_evse_exist(evse_id.value())) {
+            EVLOG_error << "Trying to make a reservation, but evse " << evse_id.value() << " is not a valid evse id.";
+            send_reserve_now_rejected_response(call.uniqueId, "Evse id does not exist");
+            return;
+        }
+
+        // Check if there is a connector available for this evse id.
+        std::optional<ConnectorStatusEnum> status =
```
Pietfried, line 3378: use `does_connector_exist` here.

lib/ocpp/v201/charge_point.cpp
```diff
@@ -3296,6 +3324,145 @@ void ChargePoint::handle_heartbeat_response(CallResult<HeartbeatResponse> call)
...
+        const uint64_t number_of_evses = evse_manager->get_number_of_evses();
+        if (number_of_evses <= 0) {
+            send_reserve_now_rejected_response(call.uniqueId, "No evse's found in charging station");
+            EVLOG_error << "Trying to make a reservation, but number of evse's is 0";
+            return;
+        }
+
+        bool status_found = false;
+        for (uint64_t i = 1; i <= number_of_evses; i++) {
+            std::optional<ConnectorStatusEnum> status = get_connector_status(i, request.connectorType);
```
Pietfried, line 3399: use `is_connector_available` here

lib/ocpp/v201/charge_point.cpp
maaikez: why?

lib/ocpp/v201/charge_point.cpp
maaikez: But then it can also be occupied or reserved.

lib/ocpp/v201/evse_manager.cpp
```diff
@@ -30,21 +30,21 @@ EvseManager::EvseIterator EvseManager::end() {
...
-const EvseInterface& EvseManager::get_evse(int32_t id) const {
-    if (id == 0 or id > this->evses.size()) {
+const EvseInterface& EvseManager::get_evse(const int32_t id) const {
+    if (id <= 0 or id > this->evses.size()) {
         throw EvseOutOfRangeException(id);
     }
     return *this->evses.at(id - 1);
 }
 
-bool EvseManager::does_evse_exist(int32_t id) const {
```
maaikez, line 46: The name of this function is a bit strange. In principle, evse id 0 does not exist so I added that check. But composite schedule relies on this that evse id 0 does exist. So maybe we should change the name or add an extra parameter here? @Pietfried ?

include/ocpp/v201/charge_point.hpp
maaikez: Ok, can you (@marcemmers , @Pietfried ) take a look if this is what you mean and if it is more readable now?

lib/ocpp/v201/evse_manager.cpp
marcemmers: The fact that with smart charging profiles they (mis)use evse 0 should be covered in that part of the code I think. Now you can ask the EvseManager if evse 0 exists and it will say yes. Then please give me evse 0 and it will throw :thinking:

include/ocpp/v201/charge_point.hpp
marcemmers: Maybe `get_connector_status` can be a bit more explicit when it does not find an available connector. Instead of returning the state of one of the connectors I would return the first or the last or at least a defined connector status.

lib/ocpp/v201/evse_manager.cpp
maaikez: Exactly, that's strange isn't it?
So it's not for this pull request but shall I make a ticket maybe?

include/ocpp/v201/charge_point.hpp
maaikez: It does return the status of the last found connector with the given connector type. But if no connector with that type is found, it returns a nullopt. Should that be changed? Because then there is a status returned of a connector that does not even exist?

I changed the function a bit by the way, if the connector type is unknown, it is also included in the search.

lib/ocpp/v201/evse_manager.cpp
marcemmers: Well you could change this to be `id > 0 && etc` and then fix the issues in smart charging as well, should be a small fix. Or indeed in a different ticket if you prefer that.

include/ocpp/v201/charge_point.hpp
marcemmers: Ah sorry, I meant this part `If is no 'Available' connector, it will return the status of one of the connectors.` which makes it seem like it just picks a connector to return the status for.

lib/ocpp/v201/evse_manager.cpp
maaikez: Yes I created a bug report, as the function is also called from another function and I think to be sure, we have to test smart charging if everything is still fine. 

https://github.com/EVerest/libocpp/issues/868

lib/ocpp/v201/evse.cpp
```diff
@@ -121,6 +205,25 @@ void Evse::delete_database_transaction() {
...
+    ComponentVariable connector_cv = ConnectorComponentVariables::get_component_variable(
+        this->evse_id, connector_id, ConnectorComponentVariables::Type);
+
+    const std::optional<std::string> connector_type =
+        this->device_model.get_optional_value<std::string>(connector_cv, AttributeEnum::Actual);
+    if (!connector_type.has_value()) {
+        return std::nullopt;
+    }
+
+    return conversions::string_to_connector_enum(connector_type.value());
```
Pietfried, line 224:
```suggestion
try {
    return conversions::string_to_connector_enum(connector_type.value());
}
catch (const StringToEnumException& e) {
    EVLOG_warning << "Could not convert to ConnectorEnum: " << connector_type.value();
    return std::nullopt;
}
```

Pietfried (approved):

Pietfried:
@maaikez we could also tick the Reservation functional block in the `README.md` as part of this PR :)

PR #866, opened 2024-11-13
Fixed bug setting NetworkConfigurationPriority

Pietfried:
## Describe your changes

Fixed an issue when setting the `NetworkConfigurationPriority`. During the validation, every entry of the new priority list was checked using the `connectivity_manager`s cached slots. The slots of the `connectivity manager` are only initialized at startup. A new profile could have been send using a `SetNetworkProfile.req` by the CSMS at runtime, so when validating the `NetworkConfigurationPriority` we shall not use the cached `connectivity_manager` slots but the `NetworkConnectionProfiles` variable stored in the device model, since only this variable contains the new entry.

### DIFF
--- a/lib/ocpp/v201/charge_point.cpp
+++ b/lib/ocpp/v201/charge_point.cpp
@@ -1855,10 +1855,11 @@ bool ChargePoint::validate_set_variable(const SetVariableData& set_variable_data
         const auto network_configuration_priorities = ocpp::split_string(set_variable_data.attributeValue.get(), ',');
         const auto active_security_profile =
             this->device_model->get_value<int>(ControllerComponentVariables::SecurityProfile);
-        const auto network_connection_profiles = json::parse(
-            this->device_model->get_value<std::string>(ControllerComponentVariables::NetworkConnectionProfiles));
-        for (const auto configuration_slot : network_configuration_priorities) {
-            try {
+
+        try {
+            const auto network_connection_profiles = json::parse(
+                this->device_model->get_value<std::string>(ControllerComponentVariables::NetworkConnectionProfiles));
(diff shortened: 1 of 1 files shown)


lib/ocpp/v201/charge_point.cpp
```diff
@@ -1855,16 +1855,22 @@ bool ChargePoint::validate_set_variable(const SetVariableData& set_variable_data
         const auto network_configuration_priorities = ocpp::split_string(set_variable_data.attributeValue.get(), ',');
         const auto active_security_profile =
             this->device_model->get_value<int>(ControllerComponentVariables::SecurityProfile);
+        const auto network_connection_profiles = json::parse(
```
hikinggrass, line 1858: This could potentially throw, probably best to catch it here, log a warning/error and return false

lib/ocpp/v201/charge_point.cpp
```diff
@@ -1887,6 +1893,8 @@ bool ChargePoint::validate_set_variable(const SetVariableData& set_variable_data
             } catch (const std::invalid_argument& e) {
                 EVLOG_warning << "NetworkConfigurationPriority is not an integer: " << configuration_slot;
                 return false;
+            } catch (const json::exception& e) {
+                EVLOG_warning << "Could not parse data of SetNetworkProfileRequest: " << e.what();
```
hikinggrass, line 1897: Should this return true or rather false because something went wrong?

lib/ocpp/v201/charge_point.cpp
Pietfried: Addressed this by widening the scope of the try-catch

hikinggrass (approved):

PR #870, opened 2024-11-18
Libwebsockets (usage) improvements

marcemmers:
## Describe your changes

We encountered deadlocks in the libwebsockters layer when we were often restarting the connectivity manager.
The issue is due to 2 threads calling `join()` at the same time.

I've added the proper `joinable()` checks and also removed the destruction and construction of libwebsockets every new config. This is no longer needed now we have one implementation that handles both secure and unsecure websockets.

### DIFF
--- a/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
+++ b/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
@@ -700,6 +700,9 @@ bool WebsocketLibwebsockets::connect() {
         return false;
     }
 
+    // Clear shutting down so we allow to reconnect again as well
+    this->shutting_down = false;
+
     EVLOG_info << "Connecting to uri: " << this->connection_options.csms_uri.string() << " with security-profile "
                << this->connection_options.security_profile
                << (this->connection_options.use_tpm_tls ? " with TPM keys" : "");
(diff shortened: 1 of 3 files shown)


lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -713,17 +717,23 @@ bool WebsocketLibwebsockets::connect() {
...
+            request_write();
+            this->websocket_thread->join();
+            this->websocket_thread.reset();
+        }
 
-    if (this->recv_message_thread) {
-        // Awake the receiving message thread to finish
-        recv_message_cv.notify_one();
-        this->recv_message_thread->join();
+        if (this->recv_message_thread && this->websocket_thread->joinable()) {
```
Pietfried, line 731:
```suggestion
        if (this->recv_message_thread && this->recv_message_thread->joinable()) {
```

marcemmers: Thanks, good find. Too quick with copying :speak_no_evil:

Pietfried: Actually caught this by running this PR with the [tests added here](https://github.com/EVerest/everest-core/pull/959) - seems to be valuable ;)

Pietfried (approved):

PR #871, opened 2024-11-18
Feature/data transfer functional block

Pietfried:
## Describe your changes
* Added message handler interface to be implemented by specific functional blocks
* Moved DataTransfer functionality to DataTransfer functional block using the targeted design. Added test cases for the new functional block

### DIFF
--- a/include/ocpp/v201/functional_blocks/data_transfer.hpp
+++ b/include/ocpp/v201/functional_blocks/data_transfer.hpp
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#pragma once
+
+#include <ocpp/v201/message_dispatcher.hpp>
+#include <ocpp/v201/messages/DataTransfer.hpp>
+
+namespace ocpp {
+namespace v201 {
+
+class DataTransferInterface {
+
+public:
+    /// \brief Sends a DataTransfer.req message to the CSMS using the given parameters
+    /// \param vendorId
+    /// \param messageId
+    /// \param data
+    /// \return DataTransferResponse containing the result from CSMS
+    virtual std::optional<DataTransferResponse> data_transfer_req(const CiString<255>& vendorId,
+                                                                  const std::optional<CiString<50>>& messageId,
+                                                                  const std::optional<json>& data) = 0;
+
+    /// \brief Sends a DataTransfer.req message to the CSMS using the given \p request
+    /// \param request message shall be sent to the CSMS
+    /// \return DataTransferResponse containing the result from CSMS. In case no response is received from the CSMS
+    /// because the message timed out or the charging station is offline, std::nullopt is returned
+    virtual std::optional<DataTransferResponse> data_transfer_req(const DataTransferRequest& request) = 0;
+
+    /// \brief Handles the given DataTransfer.req \p call by the CSMS by responding with a CallResult
+    virtual void handle_data_transfer_req(Call<DataTransferRequest> call) = 0;
+};
+
+class DataTransfer : public DataTransferInterface {
+
+private:
+    MessageDispatcherInterface<MessageType>& message_dispatcher;
+    std::optional<std::function<DataTransferResponse(const DataTransferRequest& request)>> data_transfer_callback;
+    std::chrono::seconds response_timeout;
+    std::function<bool()> is_websocket_connected;
+
+public:
+    DataTransfer(MessageDispatcherInterface<MessageType>& message_dispatcher,
+                 const std::optional<std::function<DataTransferResponse(const DataTransferRequest& request)>>&
+                     data_transfer_callback,
+                 const std::function<bool()> is_websocket_connected, const std::chrono::seconds response_timeout) :
+        message_dispatcher(message_dispatcher),
+        data_transfer_callback(data_transfer_callback),
+        is_websocket_connected(is_websocket_connected),
+        response_timeout(response_timeout){};
+
+    void handle_data_transfer_req(Call<DataTransferRequest> call) override;
+
+    std::optional<DataTransferResponse> data_transfer_req(const CiString<255>& vendorId,
+                                                          const std::optional<CiString<50>>& messageId,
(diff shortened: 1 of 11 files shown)


include/ocpp/v201/functional_blocks/data_transfer.hpp
```diff
@@ -0,0 +1,62 @@
...
+                                                                  const std::optional<json>& data) = 0;
+
+    /// \brief Sends a DataTransfer.req message to the CSMS using the given \p request
+    /// \param request message shall be sent to the CSMS
+    /// \return DataTransferResponse containing the result from CSMS. In case no response is received from the CSMS
+    /// because the message timed out or the charging station is offline, std::nullopt is returned
+    virtual std::optional<DataTransferResponse> data_transfer_req(const DataTransferRequest& request) = 0;
+
+    /// \brief Handles the given DataTransfer.req \p call by the CSMS by responding with a CallResult
+    virtual void handle_data_transfer_req(Call<DataTransferRequest> call) = 0;
```
marcemmers, line 31: For this functional block not really necessary but for the others it would be nicer if this would just be one call I think. Something like `handle_message()` that internally splits again based on the message type just like the charge_point `handle_message`.

marcemmers: Should probably then pass along the `const EnhancedMessage<v201::MessageType>& message`

Pietfried: Ah right, I should have added a comment for this since you marked that already in the other PR. I will add it here and definitely agree for other functional blocks, but in the end that is handled/defined by every functional block individually, since the `handle_message` function is not defined in any general interface but in the interface of every functional block.

lib/ocpp/v201/functional_blocks/data_transfer.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+std::optional<DataTransferResponse> DataTransfer::data_transfer_req(const DataTransferRequest& request) {
+    DataTransferResponse response;
+    response.status = DataTransferStatusEnum::Rejected;
+
+    ocpp::Call<DataTransferRequest> call(request);
+    auto data_transfer_future = this->message_dispatcher.dispatch_call_async(call);
+
+    if (not this->is_websocket_connected()) {
```
marcemmers, line 44: I don't really like that we have to pass this check along here. Especially since we also have `enhanced_message.offline` as a value that we return. Would be nice if we can remove the is connected callback and just use the other behavior instead.

If we really do need to have information if we are connected, maybe its better to abstract it into the message dispatcher instead since all blocks will have that anyway.

include/ocpp/v201/functional_blocks/data_transfer.hpp
marcemmers: Ah ok no worries. Yes it does not matter that much but thought it would be nice for uniformity ;)

marcemmers:
Overall looks really good, I think we are heading in the right direction with this! Also good to see at least 100% line coverage on the new functional block files :+1:

include/ocpp/v201/functional_blocks/data_transfer.hpp
maaikez: I wanted to say the same, it would be nice to have a handle_message in every functional block.

maaikez (approved):
Looks good. I would like to have a 'handle_message' function (or something) that handles all messages of this functional block so we can get rid of all those functions in chargepoint. I think that's cleaner.

include/ocpp/v201/functional_blocks/data_transfer.hpp
Pietfried: I changed this to a `handle_message` function and found it reasonable to add an interface for it. Using that interface would be more of a convention than something that we can enforce when implementing other functional blocks but it provides at least some soft guidance.

This change goes beyond the scope of the PR, but since we are in the the message dispatching / handling is tightly related to the first functional block implementing it, i would be in favor of merging this as part of this PR.

lib/ocpp/v201/functional_blocks/data_transfer.cpp
Pietfried: I agree, I was just trying to move the functionality to the functional block and provide everything that is required. I think we can indeed just make use of `enhanced_message.offline` here.

Pietfried:
I added the changes discussed above, so it is again ready for review

marcemmers (approved):

lib/ocpp/v201/charge_point.cpp
```diff
@@ -4044,65 +4054,14 @@ void ChargePoint::handle_clear_display_message(const Call<ClearDisplayMessageReq
...
 std::optional<DataTransferResponse> ChargePoint::data_transfer_req(const CiString<255>& vendorId,
                                                                    const std::optional<CiString<50>>& messageId,
                                                                    const std::optional<json>& data) {
-    DataTransferRequest req;
-    req.vendorId = vendorId;
-    req.messageId = messageId;
-    req.data = data;
-
-    return this->data_transfer_req(req);
+    return this->data_transfer->data_transfer_req(vendorId, messageId, data);
```
maaikez, line 4060: Is this function still necessary then?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -4044,65 +4054,14 @@ void ChargePoint::handle_clear_display_message(const Call<ClearDisplayMessageReq
...
-            this->message_dispatcher->dispatch_call_error(call_error);
-            return std::nullopt;
-        }
-    }
-    if (enhanced_message.offline) {
-        return std::nullopt;
-    }
-
-    return response;
+    return this->data_transfer->data_transfer_req(request);
```
maaikez, line 4064: Same for this one, is it still necessary?

lib/ocpp/v201/charge_point.cpp
Pietfried: Yes since it doesnt handle a `DataTransfer.req` from the CSMS but sends a `DataTransfer.req` to the CSMS

PR #872, opened 2024-11-19
Fix bug in ClearChargingProfile.req

Pietfried:
## Describe your changes
Fixed issue that other charging profiles were cleared even when a profile id was specified

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -2359,8 +2359,6 @@ void ChargePointImpl::handleGetCompositeScheduleRequest(ocpp::Call<GetCompositeS
 void ChargePointImpl::handleClearChargingProfileRequest(ocpp::Call<ClearChargingProfileRequest> call) {
     EVLOG_debug << "Received ClearChargingProfileRequest: " << call.msg << "\nwith messageId: " << call.uniqueId;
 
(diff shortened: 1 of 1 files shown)


lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -2372,8 +2370,8 @@ void ChargePointImpl::handleClearChargingProfileRequest(ocpp::Call<ClearCharging
                this->smart_charging_handler->clear_all_profiles_with_filter(
                    call.msg.id, call.msg.connectorId, call.msg.stackLevel, call.msg.chargingProfilePurpose, true)) {
         response.status = ClearChargingProfileStatus::Accepted;
-
-    } else if (this->smart_charging_handler->clear_all_profiles_with_filter(
+    } else if (!call.msg.id and
+               this->smart_charging_handler->clear_all_profiles_with_filter(
                    call.msg.id, call.msg.connectorId, call.msg.stackLevel, call.msg.chargingProfilePurpose, false)) {
```
hikinggrass, line 2375: since `call.msg.id` must be std::nullopt (because of the`!call.msg.id` in the if) at this point maybe you can directly pass `std::nullopt for `call.msg.if` here which would make this a bit more readable

hikinggrass (approved):

PR #874, opened 2024-11-20
Connectivity manager refactoring

marcemmers:
## Describe your changes

We refactored the connectivity manager a bit. We tried to make the interface simpler by just having `connect()` and `disconnect()` calls to control the manager instead of having 4.

This way we also are able to have slightly more control over how libocpp starts (with `start_connecting`) and be able to trigger a connection using a specific network profile.

### DIFF
--- a/lib/ocpp/v201/connectivity_manager.cpp
+++ b/lib/ocpp/v201/connectivity_manager.cpp
@@ -201,6 +201,43 @@ bool ConnectivityManager::on_try_switch_network_connection_profile(const int32_t
     return true;
 }
 
+void ConnectivityManager::check_cache_for_invalid_security_profiles() {
+    const auto security_level = this->device_model.get_value<int>(ControllerComponentVariables::SecurityProfile);
+
+    if (this->last_security_level == security_level) {
+        return;
+    }
+    this->last_security_level = security_level;
+
+    auto before_slot = this->get_active_network_configuration_slot();
+
+    EVLOG_info << "Before cleanup";
+    for (auto slot: this->network_connection_priorities) {
+        EVLOG_info << "Slot " << slot << " sec level: " << this->get_network_connection_profile(slot)->securityProfile;
+    }
+
+    auto is_lower_security_level = [this, security_level](const int slot) {
+        const auto opt_profile = this->get_network_connection_profile(slot);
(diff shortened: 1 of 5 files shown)


lib/ocpp/v201/connectivity_manager.cpp
```diff
@@ -85,198 +88,141 @@ ConnectivityManager::get_network_connection_profile(const int32_t configuration_
...
-                .value_or(default_network_config_timeout_seconds);
-
-        std::future_status status = config_status.wait_for(std::chrono::seconds(config_timeout));
-
-        switch (status) {
-        case std::future_status::deferred:
-        case std::future_status::timeout: {
-            EVLOG_warning << "Timeout configuring config slot: " << config_slot_int;
+        if (std::optional<std::string> interface_address =
+                get_network_configuration_interface(configuration_slot_to_set, network_connection_profile.value());
```
Pietfried, line 189: This confused me a little bit.

`get_network_configuration_interface` makes use of the `configure_network_connection_profile_callback` but `get_network_configuration_interface` does not return the response of `configure_network_connection_profile_callback` though, it just returns an `interface_address`. So if `configure_network_connection_profile_callback` returns `{"sucess": true, interface_string: "std::nullopt"}` no connection attempt will be made although that should probably happen, just without specifying the iface.

lib/ocpp/v201/connectivity_manager.cpp
```diff
@@ -85,198 +88,141 @@ ConnectivityManager::get_network_connection_profile(const int32_t configuration_
...
+            interface_address.has_value()) {
+            connection_options->iface = interface_address;
+        } else {
             can_use_connection_profile = false;
```
Pietfried, line 193: Id prefer to add a log that mentions that the profile for this slot can not be used.

Pietfried (changes requested):
Looks good and more clean :+1: 

Just a couple of remarks where id like to see some changes.

lib/ocpp/v201/connectivity_manager.cpp
marcemmers: That is a fair point. Renamed the function a little bit and allowed it to be used with nullopt as interface.

Pietfried (approved):

PR #878, opened 2024-11-27
Feature/958 reservation all connectors reserved

maaikez:
## Describe your changes
Update status document for reservations

See other related pull requests:
https://github.com/EVerest/everest-core/pull/972
https://github.com/EVerest/everest-utils/pull/171

### DIFF
--- a/lib/ocpp/v201/charge_point_callbacks.cpp
+++ b/lib/ocpp/v201/charge_point_callbacks.cpp
@@ -75,20 +75,20 @@ bool Callbacks::all_callbacks_valid(std::shared_ptr<DeviceModel> device_model) c
                 valid = false;
             }
         }
-    }
 
-    if (valid) {
         if (device_model->get_optional_value<bool>(ControllerComponentVariables::ReservationCtrlrAvailable)
                 .value_or(false)) {
             if (!this->reserve_now_callback.has_value() or this->reserve_now_callback == nullptr) {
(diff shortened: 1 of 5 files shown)


lib/ocpp/v201/charge_point_callbacks.cpp
```diff
@@ -77,6 +77,22 @@ bool Callbacks::all_callbacks_valid(std::shared_ptr<DeviceModel> device_model) c
...
+            }
+
+            if (!this->cancel_reservation_callback.has_value() or this->cancel_reservation_callback == nullptr) {
+                EVLOG_error
+                    << "Reservation is set to 'Available' and 'Enabled' in device model, but cancel_reservation "
+                       "callback is not implemented";
+            }
+        }
+    }
+
```
Pietfried, line 95: This looks like it could be moved inside the upper `if (valid)` and it does not set the `valid` variable to false in case of the error logs

Pietfried (changes requested):

lib/ocpp/v201/charge_point_callbacks.cpp
maaikez: Ah good catch

I put it in this `if (valid)` so it will not continue if the previous were not valid. But maybe it is a good idea to have a message of all not valid callbacks instead of fixing one and getting an error about the next. So changed it accordingly as you asked indeed :).

Pietfried (approved):

PR #886, opened 2024-11-28
Add option to select multiple ocpp versions to libwebsockets and connectivity_manager

marcemmers:
## Describe your changes

We can now fill in an array in connectivity options and libwebsockets will forward that to the CSMS. The CSMS must pick a protocol and we return that to the connectivity manager and charge point.

### DIFF
--- a/include/ocpp/common/types.hpp
+++ b/include/ocpp/common/types.hpp
@@ -610,7 +610,9 @@ std::ostream& operator<<(std::ostream& os, const CertificateHashDataChain& k);
 
 enum class OcppProtocolVersion {
     v16,
-    v201
+    v201,
+    v21,
+    Unknown,
 };
 
 namespace conversions {
--- a/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
+++ b/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
@@ -233,7 +233,10 @@ static bool verify_csms_cn(const std::string& hostname, bool preverified, const 
 
 WebsocketLibwebsockets::WebsocketLibwebsockets(const WebsocketConnectionOptions& connection_options,
                                                std::shared_ptr<EvseSecurity> evse_security) :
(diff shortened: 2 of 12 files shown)


include/ocpp/common/types.hpp
```diff
@@ -610,7 +610,9 @@ std::ostream& operator<<(std::ostream& os, const CertificateHashDataChain& k);
 
 enum class OcppProtocolVersion {
     v16,
-    v201
+    v201,
+    v21,
+    Unknown,
```
marcemmers, line 615: I am not quite sure I like adding 'Unknown' to this list. If we don't then I would have to pass optionals through all callbacks to signify if we don't get a valid result from the CSMS.

What would you guys prefer?

Pietfried: I think it is fine to add it here

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -638,6 +650,15 @@ void WebsocketLibwebsockets::client_loop() {
 
     auto& uri = this->connection_options.csms_uri;
 
+    std::string ocpp_versions;
+    for (bool first = true; auto version : this->connection_options.ocpp_versions) {
```
Pietfried, line 654: Range based for loops with initializer is only available in C++20, we are still on C++17.

marcemmers: Yeah I saw, didn't know this and my compiler just worked. I'll fix it ;)

marcemmers:
```suggestion
    bool first = true;
    for (auto version : this->connection_options.ocpp_versions) {
```

maaikez (approved):
I tested it with octt: migrate to new connection profile and change security profile. It connects fine and it also connects to another backend. So for me it looks all fine.

PR #889, opened 2024-11-28
Defer websocket connect()

Pietfried:
## Describe your changes
The `start()` function of the charge_point could be blocking until a connection timeout occured when connection the websocket for the first time. This change defers the call to the connect() function so that the start() function does not block anymore.

Additionally introduced thread_mutex protecting access to threads of libwebsockets when the ~WebsocketLibwebsockets destructor was called it attempted to acquire a lock for the connection_mutex and joins the websocket_thread . But the websocket_thread runs the client_loop that also requires the lock when calling on_conn_fail, so the client_loop which could result in a deadlock

### DIFF
--- a/lib/ocpp/v201/connectivity_manager.cpp
+++ b/lib/ocpp/v201/connectivity_manager.cpp
@@ -135,7 +135,7 @@ void ConnectivityManager::connect(std::optional<int32_t> configuration_slot_opt)
         // After the websocket gets closed a reconnect will be triggered
         this->websocket->disconnect(WebsocketCloseReason::ServiceRestart);
     } else {
-        this->try_connect_websocket();
+        this->websocket_timer.timeout([this] { this->try_connect_websocket(); }, WEBSOCKET_INIT_DELAY);
     }
 }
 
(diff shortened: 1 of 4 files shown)


lib/ocpp/v201/connectivity_manager.cpp
```diff
@@ -135,7 +135,7 @@ void ConnectivityManager::connect(std::optional<int32_t> configuration_slot_opt)
         // After the websocket gets closed a reconnect will be triggered
         this->websocket->disconnect(WebsocketCloseReason::ServiceRestart);
     } else {
-        this->try_connect_websocket();
+        this->websocket_timer.timeout([this] { this->try_connect_websocket(); }, std::chrono::seconds(0));
```
Pietfried, line 138: @marcemmers I used `std::chrono::seconds(0)` instead of `WEBSOCKET_INIT_DELAY` because the `WEBSOCKET_INIT_DELAY` was only introduced to solve an issue with OCTT tests reconnects as far as I remember.

marcemmers (approved):

lib/ocpp/v201/connectivity_manager.cpp
marcemmers: Yes I don't think there would be a point delaying there. Although I think we should probably revisit misusing a timer for this at all, that would be out of scope for this.

Pietfried:
@marcemmers please have another looks at the recent changes

Pietfried:
> @marcemmers please have another looks at the recent changes

Hi @marcemmers , we are looking into it and considering some more changes to the `WebsocketLibwebsockets` class, so you can wait with the review for this one

Pietfried:
Closed in favor of https://github.com/EVerest/libocpp/pull/896

PR #902, opened 2024-12-06
1.6: Support multiple temperature measurements

hikinggrass:

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -890,16 +890,21 @@ std::optional<MeterValue> ChargePointImpl::get_latest_meter_value(int32_t connec
 
             case Measurand::Temperature: {
                 // temperature
-                const auto temperature = measurement.temperature_C;
-                if (temperature) {
+                for (const auto temperature : measurement.temperature_C) {
                     sample.unit.emplace(UnitOfMeasure::Celsius);
-                    if (temperature.value().location.has_value()) {
-                        sample.location.emplace(conversions::string_to_location(temperature.value().location.value()));
-                    } else {
-                        sample.location.emplace(Location::EV);
+                    if (temperature.location.has_value()) {
+                        try {
+                            sample.location.emplace(conversions::string_to_location(temperature.location.value()));
+                        } catch (const StringToEnumException& e) {
+                            EVLOG_warning << "Could not convert string: " << temperature.location.value()
+                                          << " to Location";
+                        }
                     }
-                    sample.value = ocpp::conversions::double_to_string(temperature.value().value);
-                } else {
+                    sample.value = ocpp::conversions::double_to_string(temperature.value);
+                    filtered_meter_value.sampledValue.push_back(sample);
+                    sample = {};
+                }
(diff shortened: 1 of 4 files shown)


lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -890,16 +890,21 @@ std::optional<MeterValue> ChargePointImpl::get_latest_meter_value(int32_t connec
...
-                    sample.value = ocpp::conversions::double_to_string(temperature.value().value);
-                } else {
+                    sample.value = ocpp::conversions::double_to_string(temperature.value);
+                    filtered_meter_value.sampledValue.push_back(sample);
+                    sample = {};
+                }
+                if (measurement.temperature_C.empty()) {
                     EVLOG_debug << "Measurement does not contain temperature_C configured measurand";
                 }
                 break;
```
Pietfried, line 910:
```suggestion
                }
                if (measurement.temperature_C.empty()) {
                    EVLOG_debug << "Measurement does not contain temperature_C configured measurand";
                }
                sample = {};
                break;
```

Pietfried (changes requested):
We should also add `Temperature` to the `SupportedMeasurands` defaults in the Internal.json schema

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -890,16 +890,21 @@ std::optional<MeterValue> ChargePointImpl::get_latest_meter_value(int32_t connec
...
-                    if (temperature.value().location.has_value()) {
-                        sample.location.emplace(conversions::string_to_location(temperature.value().location.value()));
-                    } else {
-                        sample.location.emplace(Location::EV);
+                    if (temperature.location.has_value()) {
+                        try {
+                            sample.location.emplace(conversions::string_to_location(temperature.location.value()));
+                        } catch (const StringToEnumException& e) {
+                            EVLOG_warning << "Could not convert string: " << temperature.location.value()
+                                          << " to Location";
```
Pietfried, line 900: Id rather set this to `EVLOG_debug` since it will repeately be printed if the location cant be converted for every meter value

lib/ocpp/v16/charge_point_impl.cpp
hikinggrass: Woudn't that make it undefined what happens in the next iteration of the loop? For example we have one temperature reading with a location, followed by one without. If we only reset outside the loop we would carry the location to every sample without a location.
(btw. this whole function needs to be refactored at some point, it's become a bit large and repetitive)

hikinggrass:
> We should also add `Temperature` to the `SupportedMeasurands` defaults in the Internal.json schema

done

lib/ocpp/v16/charge_point_impl.cpp
Pietfried: Ah I see, but we need to set `context` and `format` and `measurand` for the `sample` again if we reset it with `= {}` or we could simply only do `sample.location.reset()`

hikinggrass: Ah totally missed this, changed it

Pietfried (approved):

PR #911, opened 2024-12-12
Censor write only variables logging and added a new callback to sanitize any logging that would be passed to the existing message_callback

WilcodenBesten:
## Describe your changes
1. Added DeviceModel::get_mutability so the mutability of a component variable can be checked before the value change is logged. This prevents write-only varaiables to be logged


WilcodenBesten:
The log line here: https://github.com/EVerest/libocpp/blob/main/lib/ocpp/common/websocket/websocket_libwebsockets.cpp#L996 can still leak secrets as it's not filtered. You could argue that it's debug so that's not needed, but better safe then sorry imo. 

Would love to here you're opinion on that.

include/ocpp/common/ocpp_logging.hpp
```diff
@@ -60,6 +60,7 @@ class MessageLogging {
     std::filesystem::path security_log_file;
     std::ofstream security_log_os;
     std::mutex output_file_mutex;
+    std::function<std::string(const std::string& message)> sanitize_message_callback;
```
hikinggrass, line 63: As discussed in the last WG meeting, lets drop this callback from this PR and just merge the write only variable logging changes. I'll follow up with a generic solution

hikinggrass (changes requested):

include/ocpp/common/ocpp_logging.hpp
marcemmers: I removed the sanitizing so it should be ok like this

lib/ocpp/v201/charge_point.cpp
```diff
@@ -518,6 +518,7 @@ void ChargePoint::configure_message_logging_format(const std::string& message_lo
     bool log_security = log_formats.find("security") != log_formats.npos;
     bool session_logging = log_formats.find("session_logging") != log_formats.npos;
     bool message_callback = log_formats.find("callback") != log_formats.npos;
+    std::function<std::string(const std::string& message)> sanitize_ocpp_messages_callback = nullptr;
```
hikinggrass, line 521: That's not needed anymore

marcemmers:
```suggestion
```

Yep, missed that

hikinggrass (approved):

PR #913, opened 2024-12-13
Block possibility of using a stale reference in SafeQueue

AssemblyJohn:
Fixed a possible dangling stale reference usage when using 'front' that returns a reference to the front member in the SafeQueue. The reference can be 'pop'-ed from another thread that will result in accessing stale data.

### DIFF
--- a/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
+++ b/lib/ocpp/common/websocket/websocket_libwebsockets.cpp
@@ -1515,13 +1515,13 @@ void WebsocketLibwebsockets::on_conn_writable() {
 
     // Execute while we have messages that were polled
     while (true) {
-        WebsocketMessage* message = nullptr;
+        std::shared_ptr<WebsocketMessage> message;
 
         // Break if we have en empty queue
         if (message_queue.empty())
             break;
 
-        message = message_queue.front().get();
+        message = message_queue.front();
 
         if (message == nullptr) {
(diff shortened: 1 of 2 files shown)


lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -1515,13 +1515,13 @@ void WebsocketLibwebsockets::on_conn_writable() {
...
     while (true) {
-        WebsocketMessage* message = nullptr;
+        std::shared_ptr<WebsocketMessage> message;
 
         // Break if we have en empty queue
         if (message_queue.empty())
             break;
 
-        message = message_queue.front().get();
+        message = message_queue.front();
```
a-w50, line 1524: just use auto here and drop the line 1518

lib/ocpp/common/websocket/websocket_libwebsockets.cpp
```diff
@@ -1548,7 +1548,7 @@ void WebsocketLibwebsockets::on_conn_writable() {
         // Poll a single message
         EVLOG_debug << "Client writable, sending message part!";
 
-        WebsocketMessage* message = message_queue.front().get();
+        std::shared_ptr<WebsocketMessage> message = message_queue.front();
```
a-w50, line 1551: same here

a-w50 (approved):

Pietfried (approved):

PR #917, opened 2024-12-16
Apply validation on the Custom part on ChangeConfiguration.req

Matthias-NIDEC:
## Describe your changes
The json which is validated against the Custom profile, is used based on the config.json. Otherwise, it misses keys and throws an exception.
- [x ] I have performed a self-review of my code
- [ x] I have made corresponding changes to the documentation

### DIFF
--- a/lib/ocpp/v16/charge_point_configuration.cpp
+++ b/lib/ocpp/v16/charge_point_configuration.cpp
@@ -2953,9 +2953,9 @@ ConfigurationStatus ChargePointConfiguration::setCustomKey(CiString<50> key, CiS
 
         // validate the updated key against the schema
         Schemas schema(custom_schema);
-        json model;
-        model[key] = new_value;
-        schema.get_validator()->validate(model); // throws exception on error
+        json modelUnderTest=config["Custom"];
+        modelUnderTest[key] = new_value;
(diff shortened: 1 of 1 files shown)


lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -2941,9 +2941,9 @@ ConfigurationStatus ChargePointConfiguration::setCustomKey(CiString<50> key, CiS
 
         // validate the updated key against the schema
         Schemas schema(custom_schema);
-        json model;
-        model[key] = new_value;
-        schema.get_validator()->validate(model); // throws exception on error
+        json modelUnderTest=config["Custom"];
+        modelUnderTest[key] = new_value;
+        schema.get_validator()->validate(modelUnderTest); // throws exception on error
```
Pietfried, line 2946:
```suggestion
        json model=config["Custom"];
        model[key] = new_value;
        schema.get_validator()->validate(model); // throws exception on error
```

Pietfried (changes requested):
Thanks, looks good! In Libocpp we use `snake_case` for variables. Please see the suggested change.

Pietfried (approved):

PR #918, opened 2024-12-16
Add interfaces and mocks for connectivity manager and ocsp updater.

maaikez:
## Describe your changes

Add interfaces and mocks for the connectivity manager and ocsp updater.

This is needed to be able to create a function block for security. 


Pietfried (changes requested):
Looks good! You can drop the `virtual` keyword in the derived classes `ConnectivityManager` and `OcspUpdater` since its inherited from the interface. This also more in line with the existing interfaces we have

Pietfried (approved):

PR #934, opened 2025-01-10
Change cmake test file so it is  possible to add separate unit test executables

maaikez:
## Describe your changes

Until now libocpp always used one big test executable. All tests were linked against the whole libocpp library. That is not always necessary and desirable. With some cmake changes, it is possible to add separate test executables.

I changed the utils_tests so it is created as a separate target instead of inside the default libocpp_unit_tests target, just to make it work and to have an example of how you can do this.


tests/CMakeLists.txt
```diff
@@ -69,16 +137,42 @@ add_custom_command(TARGET libocpp_unit_tests POST_BUILD
...
+    target_compile_definitions(${TEST_NAME} PRIVATE ${TEST_COMPILE_DEFINITIONS})
+    target_compile_definitions(${TEST_NAME} PRIVATE
+        MIGRATION_FILE_VERSION_V16=${MIGRATION_FILE_VERSION_V16}
+        MIGRATION_FILE_VERSION_V201=${MIGRATION_FILE_VERSION_V201}
+        MIGRATION_DEVICE_MODEL_FILE_VERSION_V201=${MIGRATION_DEVICE_MODEL_FILE_VERSION_V201})
+    target_compile_options(${TEST_NAME} PRIVATE ${TEST_COMPILE_OPTIONS})
+    target_compile_features(${TEST_NAME} PUBLIC ${TEST_COMPILE_FEATURES})
+    target_include_directories(${TEST_NAME} PRIVATE ${TEST_INCLUDE_DIRECTORIES})
+    target_link_libraries(${TEST_NAME} PRIVATE ${LIBOCPP_TEST_DEFAULT_LINK_LIBRARIES})
+    message("Add test ${TEST_NAME}")
```
hikinggrass, line 171: This could also use `append_coverage_compiler_flags_to_target(${TEST_NAME})` for coverage flags so we can remove the "global" `append_coverage_compiler_flags()` from here: https://github.com/EVerest/libocpp/blob/850bd796e8d4d6984486ef323d1db5cbbddefd80/CMakeLists.txt#L36C5-L36C35

tests/CMakeLists.txt
hikinggrass, line 1: In general we should refactor this a bit to move some of the setup code to functions in a dedicated .cmake, but that doesn't have to happen in this PR

hikinggrass (changes requested):

tests/libocpp-unittests.cmake
```diff
@@ -0,0 +1,104 @@
...
+                                      ${LIBOCPP_LIB_PATH}/ocpp/v201/init_device_model_db.cpp
+                                      ${LIBOCPP_LIB_PATH}/ocpp/v201/device_model_storage_sqlite.cpp
+                                      ${LIBOCPP_LIB_PATH}/ocpp/v201/utils.cpp
+)
+
+function(add_libocpp_unittest)
+    set(oneValueArgs NAME PATH)
+    set(options)
+    set(multiValueArgs)
+    cmake_parse_arguments(arg_libocpp_unittest "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
```
hikinggrass, line 84: this could just be `arg` to make the following code more readable

hikinggrass (approved):

PR #944, opened 2025-01-20
Move functional block O: DisplayMessage

Pietfried:
## Describe your changes
This PR moves the functionality of functional block O. that was previously implemented as part of the ChargePoint class into a dedicated class.

### DIFF
--- a/include/ocpp/v201/functional_blocks/display_message.hpp
+++ b/include/ocpp/v201/functional_blocks/display_message.hpp
@@ -40,7 +40,7 @@ typedef std::function<ClearDisplayMessageResponse(const ClearDisplayMessageReque
 
 class DisplayMessageInterface : public MessageHandlerInterface {
 public:
-    virtual ~DisplayMessageInterface() {};
+    virtual ~DisplayMessageInterface(){};
 };
 
 class DisplayMessageBlock : public DisplayMessageInterface {
--- a/include/ocpp/v201/functional_blocks/display_message.hpp
+++ b/include/ocpp/v201/functional_blocks/display_message.hpp
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+
+#pragma once
+
+#include <ocpp/v201/evse_manager.hpp>
+#include <ocpp/v201/message_dispatcher.hpp>
+#include <ocpp/v201/message_handler.hpp>
+#include <ocpp/v201/messages/ClearDisplayMessage.hpp>
+#include <ocpp/v201/messages/GetDisplayMessages.hpp>
+#include <ocpp/v201/messages/SetDisplayMessage.hpp>
+
+namespace ocpp::v201 {
+
+///
+/// \brief Convert message content from OCPP spec to DisplayMessageContent.
+/// \param message_content  The struct to convert.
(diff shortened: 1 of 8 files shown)


lib/ocpp/v201/charge_point.cpp
```diff
@@ -1199,6 +1194,24 @@ void ChargePoint::initialize(const std::map<int32_t, int32_t>& evse_connector_st
             this->callbacks.is_reservation_for_token_callback);
     }
 
+    if (device_model->get_optional_value<bool>(ControllerComponentVariables::DisplayMessageCtrlrAvailable)
+            .value_or(false)) {
+        if (this->callbacks.get_display_message_callback.has_value() and
+            this->callbacks.set_display_message_callback.has_value() and
+            this->callbacks.clear_display_message_callback.has_value()) {
```
maaikez, line 1201: Those checks are already done in charge_point_callbacks.cpp, I wonder if we should do them twice? I think you only have to check if DisplayMessageCtrlrAvailable is true here. The check is done in the constructor of ChargePoint and will throw there if the callbacks do not have a value.

include/ocpp/v201/functional_blocks/display_message.hpp
```diff
@@ -0,0 +1,69 @@
...
+/// \param display_message  The struct to convert.
+/// \return The converted struct.
+///
+std::optional<MessageInfo> display_message_to_message_info_type(const DisplayMessage& display_message);
+///
+/// \brief Convert message info from OCPP to DisplayMessage.
+/// \param message_info The struct to convert.
+/// \return The converted struct.
+///
+DisplayMessage message_info_to_display_message(const MessageInfo& message_info);
```
maaikez, line 32: Did you make those free functions (and not static) to be able to test them? Just wondering.

include/ocpp/v201/functional_blocks/display_message.hpp
```diff
@@ -0,0 +1,69 @@
...
+    SetDisplayMessageCallback;
+typedef std::function<ClearDisplayMessageResponse(const ClearDisplayMessageRequest& request)>
+    ClearDisplayMessageCallback;
+
+class DisplayMessageInterface : public MessageHandlerInterface {
+public:
+    virtual ~DisplayMessageInterface(){};
+};
+
+class DisplayMessageBlock : public DisplayMessageInterface {
```
maaikez, line 46: I see you call the class 'DisplayMessageBlock'. All the other classes in functional blocks to not have the 'Block' in the names tail. Shouldn't we keep it the same everywhere? So either remove 'Block' here or add it to the other functional blocks as well.

maaikez (changes requested):

include/ocpp/v201/functional_blocks/display_message.hpp
Pietfried: yes it allows for easier unit testing although these functions are quite straight forward and could also be tested indirectly

include/ocpp/v201/functional_blocks/display_message.hpp
Pietfried: There is already an autogenerated `ocpp::v201::DisplayMessage` type, which makes finding a proper solution a little awkward..

maaikez (approved):

PR #966, opened 2025-01-29
Catch exception when assiging String<> type with too long string

marcemmers:
## Describe your changes

Fixes a crash when the MessageId string received is longer than 36 characters. Instead of crashing we now send an RpcFrameworkError.

### DIFF
--- a/include/ocpp/common/string.hpp
+++ b/include/ocpp/common/string.hpp
@@ -15,7 +15,7 @@ class StringConversionException : public std::runtime_error {
     using std::runtime_error::runtime_error;
 };
 
-enum class StringToLarge {
+enum class StringTooLarge {
     Throw,
     Truncate
 };
@@ -28,11 +28,11 @@ private:
 
 public:
     /// \brief Creates a string from the given \p data
-    String(const std::string& data, StringToLarge to_large = StringToLarge::Throw) {
+    explicit String(const std::string& data, StringTooLarge to_large = StringTooLarge::Throw) {
         this->set(data, to_large);
     }
 
-    String(const char* data, StringToLarge to_large = StringToLarge::Throw) {
+    explicit String(const char* data, StringTooLarge to_large = StringTooLarge::Throw) {
         this->set(data, to_large);
     }
 
@@ -46,10 +46,10 @@ public:
     }
 
     /// \brief Sets the content of the string to the given \p data
-    void set(const std::string& data, StringToLarge to_large = StringToLarge::Throw) {
+    void set(const std::string& data, StringTooLarge to_large = StringTooLarge::Throw) {
         std::string_view view = data;
         if (view.length() > this->length) {
-            if (to_large == StringToLarge::Throw) {
+            if (to_large == StringTooLarge::Throw) {
                 throw StringConversionException("String length (" + std::to_string(view.length()) +
                                                 ") exceeds permitted length (" + std::to_string(this->length) + ")");
             }
--- a/include/ocpp/common/string.hpp
+++ b/include/ocpp/common/string.hpp
@@ -15,25 +15,29 @@ class StringConversionException : public std::runtime_error {
     using std::runtime_error::runtime_error;
 };
 
+enum class StringToLarge {
+    Throw,
+    Truncate
+};
+
 /// \brief Contains a String impementation with a maximum length
 template <size_t L> class String {
 private:
     std::string data;
-    size_t length;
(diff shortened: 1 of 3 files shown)


lib/ocpp/v201/charge_point.cpp
```diff
@@ -1134,6 +1134,15 @@ void ChargePoint::message_callback(const std::string& message) {
         this->security->security_event_notification_req(CiString<50>(security_event), CiString<255>(message), true,
                                                         utils::is_critical(security_event));
         return;
+    } catch (const StringConversionException& e) {
+        this->logging->central_system("Unknown", message);
+        EVLOG_error << "JSON exception during reception of message: " << e.what();
+        this->message_dispatcher->dispatch_call_error(
+            CallError(MessageId("-1"), "RpcFrameworkError", e.what(), json({})));
+        const auto& security_event = ocpp::security_events::INVALIDMESSAGES;
+        this->security_event_notification_req(CiString<50>(security_event), CiString<255>(message), true,
```
marcemmers, line 1143:
```suggestion
        this->security->security_event_notification_req(CiString<50>(security_event), CiString<255>(message), true,
```

Pietfried:
I think we can still end up having issues here with the SecurityEventNotification with `CiString<255>(message)` when the ocpp `message` is > 255 characters.

marcemmers:
To be honest I did not check further, this was just an issue I found with some incoming message tests that we did.

I am not sure what you are referring to here? Isn't that CiString created by the application for the SecurityEventNotification? In that case there will be an error thrown that the application can/should catch right?

lib/ocpp/v201/charge_point.cpp
```diff
@@ -1134,6 +1134,15 @@ void ChargePoint::message_callback(const std::string& message) {
         this->security->security_event_notification_req(CiString<50>(security_event), CiString<255>(message), true,
                                                         utils::is_critical(security_event));
         return;
+    } catch (const StringConversionException& e) {
+        this->logging->central_system("Unknown", message);
+        EVLOG_error << "JSON exception during reception of message: " << e.what();
+        this->message_dispatcher->dispatch_call_error(
+            CallError(MessageId("-1"), "RpcFrameworkError", e.what(), json({})));
+        const auto& security_event = ocpp::security_events::INVALIDMESSAGES;
+        this->security->security_event_notification_req(CiString<50>(security_event), CiString<255>(message), true,
```
Pietfried, line 1143: This can still throw if `message` > 255 characters or `security_event` > 50 characters.

It might be an option to truncate CiStrings by default inside the class in case the number of characters exceeds the specified limit instead of throwing the exception. I dont have a complete overview but I think it can happen at quite a few places

marcemmers: I've added an argument that can optionally truncate if we want. I don't think truncating by default is an option because that might lead to strange behavior as well. We would silently accept incorrect messages from the BO for example.

include/ocpp/common/string.hpp
```diff
@@ -11,25 +11,33 @@
 
 namespace ocpp {
 
+class StringConversionException : public std::runtime_error {
+    using std::runtime_error::runtime_error;
+};
+
+enum class StringToLarge {
```
Pietfried, line 18: nit `StringTooLarge`

include/ocpp/common/string.hpp
```diff
@@ -11,25 +11,33 @@
...
 private:
     std::string data;
-    size_t length;
+    static constexpr size_t length = L;
 
 public:
     /// \brief Creates a string from the given \p data
-    String(const std::string& data) : length(L) {
-        this->set(data);
+    String(const std::string& data, StringToLarge to_large = StringToLarge::Throw) {
```
Pietfried, line 31:
```suggestion
    explicit String(const std::string& data, StringToLarge to_large = StringToLarge::Throw) {
```

include/ocpp/common/string.hpp
```diff
@@ -11,25 +11,33 @@
...
+        this->set(data, to_large);
     }
 
-    String(const char* data) : length(L) {
-        this->set(data);
+    String(const char* data, StringToLarge to_large = StringToLarge::Throw) {
```
Pietfried, line 35:
```suggestion
    explicit String(const char* data, StringToLarge to_large = StringToLarge::Throw) {
```

Pietfried (approved):
Looks good, just added a couple of minor remarks

PR #988, opened 2025-02-17
Support OCPP 2.1 variables

maaikez:
## Describe your changes

Add support for OCPP 2.1 variables and add check for required variables in device model.


tests/lib/ocpp/v201/test_device_model.cpp
```diff
@@ -11,29 +14,52 @@ namespace v201 {
...
-    }
-
-    void TearDown() override {
-        // reset the value to default
-        dm->set_value(cv.component, cv.variable.value(), ocpp::v201::AttributeEnum::Actual, "10", "test");
+    DeviceModelTest() : device_model_test_helper(), dm(device_model_test_helper.get_device_model()) {
     }
 };
 
+TEST_F(DeviceModelTest, test_allow_zero2) {
```
Pietfried, line 26: nit: `test_allow_zero2` defined before `test_allow_zero` looks a little bit odd. A name like `test_min_limit_and_allow_zero` is probably better

Pietfried (changes requested):
One general issue I see with this change is that you can now define device models that are valid according to the new integrity check but could lead to unhandled exceptions at runtime. Lets have a look at an example.

`AlignedDataCtrlr.AlignedDataTxEndedInterval` is only required if `AlignedDataCtrlr.Available` is true. So you can define a device model that doenst contain `AlignedDataCtrlr.AlignedDataTxEndedInterval`. In the code, we just access `AlignedDataCtrlr.AlignedDataTxEndedInterval` using `this->device_model.get_value<int>(ControllerComponentVariables::AlignedDataTxEndedInterval)`. This leads to an unhandled exception.

I see three options:
1) Define every variable **that is required only if the Available variable of the Controller is present and true** as a `ComponentVariable` and not a `RequiredComponentVariable`. This avoids the option to use `get_value` and forces us to use `get_optional_value` instead. 
This is the safest option, but it leads to quite some changes in code.

2) Keep defining these variables as `RequiredComponentVariable` and use `get_value` but make sure everywhere where it is called that the `Available` variable is checked.
Requires fewer changes than 1) but we can still crash at runtime
 
3) We could enforce that more controllers are required than in the specification. This is also how the code was written (if the spec says a variable is required we require it, no matter what value `Available` of the Controller has. Libocpp does support the `AlignedDataCtrlr` features, so it must also be present. Same is true for `SampledDataCtrlr` so we can enforce that they are present and available. `LocalAuthListCtrlr`, `SmartChargingCtrlr`, `TariffCostCtrlr`, `MonitoringCtrlr` and `DisplayMessageCtrlr` I would consider optional and we could then go with option 2).
Only few changes in existing code. We still need to protect code blocks where variables are accessed directly using `get_value`. The introduction of the functional blocks should make this easier though.

Interested in your thoughts @maaikez and @marcemmers .

include/ocpp/v201/ocpp_types.hpp
```diff
@@ -2235,11 +2236,37 @@ void to_json(json& j, const Firmware& k);
...
+                                                                                  OcppProtocolVersion::v21}) :
+        ComponentVariable(), required_for(required_for) {
+        this->component = component;
+        this->variable = variable;
+        this->customData = custom_data;
+    };
+
+    /// \brief For which ocpp protocol version(s) this component variable is required.
+    std::set<OcppProtocolVersion> required_for;
+};
```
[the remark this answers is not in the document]
maaikez: I would say 3 (in combination with 2). This is the way libocpp is written so we should just make the components available that we depend on. I also see no use in making them optional (AlignedDataCtrlr and SampledDataCtrlr).

maaikez:
> One general issue I see with this change is that you can now define device models that are valid according to the new integrity check but could lead to unhandled exceptions at runtime. Lets have a look at an example.
> 
> ...
> 
> I see three options:
> 
>  ...
> 
>     2. Keep defining these variables as `RequiredComponentVariable` and use `get_value` but make sure everywhere where it is called that the `Available` variable is checked.
>        Requires fewer changes than 1) but we can still crash at runtime
> 
>     3. We could enforce that more controllers are required than in the specification. This is also how the code was written (if the spec says a variable is required we require it, no matter what value `Available` of the Controller has. Libocpp does support the `AlignedDataCtrlr` features, so it must also be present. Same is true for `SampledDataCtrlr` so we can enforce that they are present and available. `LocalAuthListCtrlr`, `SmartChargingCtrlr`, `TariffCostCtrlr`, `MonitoringCtrlr` and `DisplayMessageCtrlr` I would consider optional and we could then go with option 2).
>        Only few changes in existing code. We still need to protect code blocks where variables are accessed directly using `get_value`. The introduction of the functional blocks should make this easier though.
> 
> 
> Interested in your thoughts @maaikez and @marcemmers .

I would say 3 (in combination with 2). This is the way libocpp is written so we should just make the components available that we depend on. I also see no use in making them optional (AlignedDataCtrlr and SampledDataCtrlr).

include/ocpp/v201/ocpp_types.hpp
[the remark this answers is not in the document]
maaikez: Hmm there will be a circular dependency then (DeviceModel <-> CtrlrComponentVariables).
Why not add it in CtrlrComponentVariables?

Pietfried (commented):
* clang format is still failing
* I'd suggest to catch `DeviceModelError` here for all incoming messages  from the CSMS https://github.com/EVerest/libocpp/blob/main/lib/ocpp/v201/charge_point.cpp#L772

Pietfried (approved):

doc/v201/ocpp_201_device_model_initialization.md
```diff
@@ -43,3 +43,16 @@ addition of a new one.
...
+
+
+## Required variables
+
+There are some required Variables, which can be found in the OCPP spec.  
+Some `Variables` are only required if the `Component` is `Available`, for example `Reservation` and `Smart Charging`. 
+There are  some Components that are always required because that is how libocpp works: `AlignedDataCtrlr` and 
+`SampledDataCtrlr`.  
+When libocpp is started and initialized, all required Variables will be checked and an DeviceModelError is thrown if 
+one of the required Variables is not there.
```
Pietfried, line 55:
```suggestion
one of the required Variables is not present.
```

doc/v201/ocpp_201_device_model_initialization.md
```diff
@@ -43,3 +43,16 @@ addition of a new one.
 Note: OCPP requires EVSE and Connector numbering starting from 1 counting upwards.
 
 Note: There should be no duplicate components or variables in the component config files.
+
+
+## Required variables
+
+There are some required Variables, which can be found in the OCPP spec.  
+Some `Variables` are only required if the `Component` is `Available`, for example `Reservation` and `Smart Charging`. 
+There are  some Components that are always required because that is how libocpp works: `AlignedDataCtrlr` and
```
Pietfried, line 52:
```suggestion
There are some Components that are always required because they are required by libocpp: `AlignedDataCtrlr` and 
```

doc/v201/ocpp_201_device_model_initialization.md
```diff
@@ -43,3 +43,16 @@ addition of a new one.
 Note: OCPP requires EVSE and Connector numbering starting from 1 counting upwards.
 
 Note: There should be no duplicate components or variables in the component config files.
+
+
+## Required variables
+
+There are some required Variables, which can be found in the OCPP spec.
```
Pietfried, line 50:
```suggestion
There are Variables that are defined as **required** in the OCPP specification. 
```

doc/v201/ocpp_201_device_model_initialization.md
```diff
@@ -43,3 +43,16 @@ addition of a new one.
 Note: OCPP requires EVSE and Connector numbering starting from 1 counting upwards.
 
 Note: There should be no duplicate components or variables in the component config files.
+
+
+## Required variables
+
+There are some required Variables, which can be found in the OCPP spec.  
+Some `Variables` are only required if the `Component` is `Available`, for example `Reservation` and `Smart Charging`.
```
Pietfried, line 51:
```suggestion
Some `Variables` are only required if the `Available` variable of the respective `Component` is present and configured to true. Examples for this are the `ReservationCtrlr` and the `SmartChargingCtrlr`.
```

Pietfried (commented):
Just noticed that this has been already merged, Im still submitting the comments so we could add them later

PR #995, opened 2025-02-21
Send CALLERROR only as response to CALL message

Pietfried:
## Describe your changes
Implemented a check of the enhanced_message to handle in message_callback so that only send CALLERROR on CALL messages.

Remark: We may want to restructure out custom exceptions a little better at some point in order to get rid of that repetitive code.


lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -1356,6 +1356,9 @@ void ChargePointImpl::message_callback(const std::string& message) {
         }
     } catch (json::exception& e) {
         EVLOG_error << "JSON exception during handling of message: " << e.what();
+        if (enhanced_message.messageTypeId != MessageTypeId::CALL) {
+            return; // CALLERROR shall only follow on a CALL message
```
maaikez, line 1360: Also no security event notification or log?

maaikez (approved):
Approved, I am only not sure if we shouldn't send a security event notification.

PR #998, opened 2025-02-26
Only change state to available if the current state is 'Reserved'

maaikez:
## Describe your changes

OCPP 1.6: When a connector is reserved and it is set to 'Inoperative', it should go to 'Unavailable' and the reservation should be cancelled. 
So when we set to 'Inoperative', status notification 'Unavailable' is sent. It seems the reservation is cancelled indeed, therefore 'ReservationEnd' is called, which causes the state to change to 'Available', but directly after to 'Unavailable' again. So that is three state changes where there should be only one.


maaikez:
See https://github.com/EVerest/everest-core/pull/1067#pullrequestreview-2647851233

Pietfried (approved):

PR #1015, opened 2025-03-21
fix(Ocpp16): Make StopTransactionOnEVSideDisconnect optional

### LINKED ISSUE #1012: TC_005_3_CS: Transaction Ends Instead of Suspending When EV Side Disconnects
### OCPP Version

OCPP1.6

### Describe the bug

I am conducting a test using the OCTT tool with Test Case Id: TC_005_3_CS. In this scenario, StopTransactionOnEVSideDisconnect is set to false, and UnlockConnectorOnEVSideDisconnect is also set to false. However, when the cable is disconnected at the EV side, the transaction ends immediately rather than being suspended. Could you please explain why this occurs?"

### To Reproduce

_No response_

### Anything else?

_No response_

mlitre:
This configuration key is now made optional due to newer erratas. Additionally, the value is not supported so it is now set to true by default.

- [ x ] I have performed a self-review of my code
- [ x ] If OCPP 2.0.1 or OCPP2.1: I have updated the [OCPP 2.x status document](https://github.com/EVerest/libocpp/tree/main/doc/ocpp_2x_status.md)
- [ x ] I read the [contribution documentation](https://github.com/EVerest/EVerest/blob/main/CONTRIBUTING.md) and made sure that my changes meet its requirements

Closes #1012

### DIFF
--- a/config/v16/profile_schemas/Core.json
+++ b/config/v16/profile_schemas/Core.json
@@ -133,7 +133,7 @@
         },
         "StopTransactionOnEVSideDisconnect": {
             "type": "boolean",
-            "readOnly": false
+            "readOnly": true
         },
         "StopTransactionOnInvalidId": {
             "type": "boolean",
--- a/config/v16/profile_schemas/Core.json
+++ b/config/v16/profile_schemas/Core.json
@@ -16,7 +16,6 @@
         "MeterValueSampleInterval",
         "NumberOfConnectors",
         "ResetRetries",
-        "StopTransactionOnEVSideDisconnect",
         "StopTransactionOnInvalidId",
         "StopTxnAlignedData",
         "StopTxnSampledData",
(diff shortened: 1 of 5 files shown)


lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -3651,8 +3656,8 @@ ConfigurationStatus ChargePointConfiguration::set(CiString<50> key, CiString<500
         }
     }
     if (key == "StopTransactionOnEVSideDisconnect") {
-        if (isBool(value.get())) {
-            this->setStopTransactionOnEVSideDisconnect(ocpp::conversions::string_to_bool(value.get()));
+        if (this->getStopTransactionOnEVSideDisconnect() == std::nullopt) {
+            return ConfigurationStatus::NotSupported;
         } else {
             return ConfigurationStatus::Rejected;
         }
```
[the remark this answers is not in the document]
mlitre: Ah got it thanks!

config/v16/profile_schemas/Core.json
Pietfried, line 1: L136 in this file should be

"readOnly": true

mlitre: Done

Pietfried (approved):

PR #1016, opened 2025-03-21
Respond with to ChangeAvailability with Scheduled when in Preparing state 

Pietfried:
## Describe your changes
This change changes the behavior of the ChangeAvailability handler. It will now also respond with Scheduled and schedule the change availability request in case the connector is in the status Preparing.

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -1645,19 +1645,23 @@ void ChargePointImpl::preprocess_change_availability_request(
     std::vector<int32_t>& accepted_connector_availability_changes) {
 
     // we can only change the connector availability if there is no active transaction on this
-    // connector. is that case this change must be scheduled and we should report an availability status
-    // of "Scheduled"
+    // connector or the connector is not in Preparing state. If this is the case this change must be scheduled and we
+    // should report an availability status of "Scheduled"
+    auto should_be_scheduled_func = [this](int32_t connector) {
+        return this->transaction_handler->transaction_active(connector) or
+               this->status->get_state(connector) == ChargePointStatus::Preparing;
+    };
 
     // check if connector exists
     if (request.connectorId <= this->configuration->getNumberOfConnectors() && request.connectorId >= 0) {
-        bool transaction_running = false;
+        bool should_be_scheduled = false;
 
         if (request.connectorId == 0) {
             accepted_connector_availability_changes.push_back(0);
             int32_t number_of_connectors = this->configuration->getNumberOfConnectors();
             for (int32_t connector = 1; connector <= number_of_connectors; connector++) {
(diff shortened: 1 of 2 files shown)


lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -1645,28 +1645,32 @@ void ChargePointImpl::preprocess_change_availability_request(
     std::vector<int32_t>& accepted_connector_availability_changes) {
 
     // we can only change the connector availability if there is no active transaction on this
-    // connector. is that case this change must be scheduled and we should report an availability status
-    // of "Scheduled"
+    // connector or the connector is not in Preparing state. If this is the case this change must be scheduled and we
+    // should report an availability status of "Scheduled"
+    auto should_be_scheduled_func = [this](int32_t connector) {
+        return this->transaction_handler->transaction_active(connector) or
+               this->status->get_state(connector) == ChargePointStatus::Preparing;
```
mlitre, line 1652: Shouldn't we also check if the status is reserved?

lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -2261,22 +2284,7 @@ void ChargePointImpl::handleStopTransactionResponse(const EnhancedMessage<v16::M
...
-            persist = this->change_availability_queue[connector].persist;
-            this->change_availability_queue.erase(connector);
-        }
-
-        if (change_queued) {
-            EVLOG_debug << "Queued availability change of connector " << connector << " to "
-                        << conversions::availability_type_to_string(connector_availability);
-            execute_connectors_availability_change({connector}, connector_availability, persist);
-        }
+        this->execute_queued_availability_change(connector);
```
mlitre, line 2287: Is it not better to wait for the EV to plug out before setting the connector to unavailable?

Pietfried: A transition from Finishing to Unavailable is possible (F8), but if you look at the description you are probably right.

"F8: All user actions completed and the connector is scheduled to become Unavailable"

If we only do it in `on_session_stopped`, the transitions would be `Finishing` -> `Available` -> `Unavailable` though, because the `disable_evse_callback` does cause the transition directly but will be executed asynchrously by the consumer calling `on_disabled`. For that reason, I would keep it like this.

mlitre: Understood, this makes sense to me then!

mlitre (approved):

lib/ocpp/v16/charge_point_impl.cpp
Pietfried: It looks like this change is not in line with this 1.6 spec requirement:

"... or (3) when the Charge Point or connector are set to Faulted or
Unavailable."

PR #1026, opened 2025-03-30
Add helpers to set charge point common data

mhei:
## Describe your changes

The idea is to overwrite this data later, for example if other data sources are available than those available at the time of initialization. This PR adds the required helper methods for the OCPP 1.6 implementation.

Note: since I currently only have a OCPP 1.6 test environment at hand and my knowledge about OCPP 2.x is still limited, help with OCPP 2.x part would be welcome.

### DIFF
--- a/lib/ocpp/v16/charge_point_configuration.cpp
+++ b/lib/ocpp/v16/charge_point_configuration.cpp
@@ -262,6 +262,18 @@ void ChargePointConfiguration::init_supported_measurands() {
     }
 }
 
+void ChargePointConfiguration::setChargepointInformation(const std::string& chargePointVendor,
+                                                         const std::string& chargePointModel,
+                                                         const std::string& chargePointSerialNumber,
+                                                         const std::optional<std::string>& firmwareVersion) {
+    this->config["Internal"]["ChargePointVendor"] = chargePointVendor;
+    this->config["Internal"]["ChargePointModel"] = chargePointModel;
+    this->config["Internal"]["ChargePointSerialNumber"] = chargePointSerialNumber;
+    if (firmwareVersion.has_value()) {
+        this->config["Internal"]["FirmwareVersion"] = firmwareVersion.value();
+    }
+}
+
 // Internal config options
 std::string ChargePointConfiguration::getChargePointId() {
     return this->config["Internal"]["ChargePointId"];
--- a/include/ocpp/v16/charge_point.hpp
+++ b/include/ocpp/v16/charge_point.hpp
@@ -81,6 +81,11 @@ public:
     /// @}  // End constructors 1.6 group
     /// @}  // End chargepoint constructors topic
 
+    /// \brief Allow to update the ChargePoint information which will be sent in BootNotification.req
+    void update_chargepoint_information(const std::string& chargepoint_vendor, const std::string& chargepoint_model,
+                                        const std::string& chargepoint_serialnumber,
+                                        const std::optional<std::string>& firmware_version);
+
     /// \brief Starts the ChargePoint, initializes and connects to the Websocket endpoint and initializes a
     /// BootNotification.req
     /// \param connector_status_map initial state of connectors including connector 0 with reduced set of states
--- a/include/ocpp/v16/charge_point_configuration.hpp
+++ b/include/ocpp/v16/charge_point_configuration.hpp
@@ -43,7 +43,9 @@ private:
 public:
     ChargePointConfiguration(const std::string& config, const fs::path& ocpp_main_path,
                              const fs::path& user_config_path);
-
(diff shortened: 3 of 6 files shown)


lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -262,6 +262,18 @@ void ChargePointConfiguration::init_supported_measurands() {
     }
 }
 
+void ChargePointConfiguration::setChargepointInformation(const std::string& chargePointVendor,
```
Pietfried, line 265: Please add other properties of the BootNotification as well:
* iccid
* imsi
* meterSerialNumber
* meterType
* chargeBoxSerialNumber

lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -262,6 +262,18 @@ void ChargePointConfiguration::init_supported_measurands() {
...
+                                                         const std::string& chargePointModel,
+                                                         const std::string& chargePointSerialNumber,
+                                                         const std::optional<std::string>& firmwareVersion) {
+    this->config["Internal"]["ChargePointVendor"] = chargePointVendor;
+    this->config["Internal"]["ChargePointModel"] = chargePointModel;
+    this->config["Internal"]["ChargePointSerialNumber"] = chargePointSerialNumber;
+    if (firmwareVersion.has_value()) {
+        this->config["Internal"]["FirmwareVersion"] = firmwareVersion.value();
+    }
```
Pietfried, line 274: In order to persist these changes they need to be written into the `user_config.json`. This can be done using the `set_in_user_config` function.

Pietfried (changes requested):
Hi @mhei , thanks for adding this extension! Please see the comments in line.

lib/ocpp/v16/charge_point_configuration.cpp
mhei: ok, will add as std::optional parameter

lib/ocpp/v16/charge_point_configuration.cpp
mhei: I guess you refer to `ChargePointConfiguration::setInUserConfig` ? Hm, I'm unsure about this. When I call this function for each parameter, then this would result in multiple writes to the same file on disk. So if you insist that this should be made persistent, then I'd merge this functionality into a combined write.

But on the other hand: in my target use-case, it is not necessary to make these changes persistent since this function would be called during each boot and thus the data is always refreshed. Actually, the idea behind is that there is some common place outside the EVerest environment where such common charger data is stored permanently. Thus storing it again in the OCPP configuration file would not make sense, and especially not for e.g. IMSI or meter serial number (components which could be replaced during e.g maintenance...).

On the other hand, it would be written anyway as part of yet another key change...

Pietfried: Yes, I was indeed refering to `setInUserConfig`.

I understand your use case, but libocpp itself is not aware of your target use case that calls the function on each boot, so I would like to treat the changes of these configuration keys like all the others. Otherwise this functionality should rather go into the `start` function or the `ChargePoint` constructor.

Of course you can merge the persisting functionality into a combined write. We are aware that the configuration handling for ocpp1.6 has some room for improvements in general.

Pietfried:
Hi @mhei , are you planning to revisit this? Feel free to let us know if you need help with finishing this off

mhei:
Hi @Pietfried , I just need more free time :smile: Actually, I rebased my working setup to latest main branch yesterday evening and I plan to work on it this weekend.
But one question came to my mind during rework: what would you prefer: a single `setChargepointInformation` method with all arguments, most of them as optional ones, or two/three methods... one for the common data, one for the meter related data and one for the modem related data? I tend to split into three, but I wondering what do you prefer?

Pietfried (changes requested):
Hi @mhei , thanks for picking this up again! I like your decision of spliting the update into three separate methods and the general structure of the PR looks good!

I would prefer that you use just `std::optional` instead of `NullOrOptionalString` because it is more in line with the existing interface and implementation. I know that this provides a little less flexibility because it doesnt allow the consumer to clear exsting keys, but the existing interface also doesnt allow this and afaics the [companion PR in everest-core](https://github.com/EVerest/everest-core/pull/1109) also can not make use of the clearing functionality anyways.

We are considering to refactor the whole configuration interface and implementation at some point and this is easier to do that if there aren't multiple approaches to updating config keys.

My suggestion is to:
* update the value if the optional is set
* keep the existing value if the optional is not set

Pietfried (approved):
Thanks for applying the changes, lgtm!

hikinggrass (approved):

PR #1031, opened 2025-04-07
V2.1 composite schedule

maaikez:
## Describe your changes

Add composite schedule changes for the three phases and different limits and setpoints.


include/ocpp/common/constants.hpp
```diff
@@ -21,6 +21,8 @@ constexpr std::int32_t DEFAULT_AND_MAX_NUMBER_PHASES = 3;
 constexpr float LOW_VOLTAGE = 230;
 
 constexpr float NO_LIMIT_SPECIFIED = -1.0;
+constexpr float NO_SETPOINT_SPECIFIED = std::numeric_limits<float>::max();
+constexpr float NO_DISCHARGE_LIMIT_SPECIFIED = -std::numeric_limits<float>::max();
```
mlitre, line 25: Shouldnt we make this a positive value since anyways it will be positive? or -1 like the rest

include/ocpp/v2/profile.hpp
```diff
@@ -1,15 +1,28 @@
...
 struct IntermediatePeriod {
     int32_t startPeriod;
-    float current_limit;
-    float power_limit;
+    PeriodLimit current_limit;
+    PeriodLimit power_limit;
+    PeriodLimit current_discharge_limit;
+    PeriodLimit power_discharge_limit;
+    PeriodLimit current_setpoint;
+    PeriodLimit power_setpoint;
```
mlitre, line 25: I am still thinking that we should make these optional, but it might be nice to have another opinion

lib/ocpp/v2/profile.cpp
```diff
@@ -490,65 +596,143 @@ IntermediateProfile merge_tx_profile_with_tx_default_profile(const IntermediateP
...
+    return combine_list_of_profiles(profiles, combinator, ocpp_version);
 }
 
-IntermediateProfile merge_profiles_by_lowest_limit(const std::vector<IntermediateProfile>& profiles) {
-    auto combinator = [](const period_pair_vector& periods) {
-        IntermediatePeriod period{};
-        period.current_limit = std::numeric_limits<float>::max();
-        period.power_limit = std::numeric_limits<float>::max();
-
+IntermediateProfile merge_profiles_by_lowest_limit(const std::vector<IntermediateProfile>& profiles,
```
mlitre, line 602: What should happen if the operation mode only makes sense for limit, here we are forcing some defaults when it might not be needed. We could also end up with a setpoint higher than the limit which seems strange.

mlitre:
I think it might be worth adding a diagram of how we merge profiles together to make things clear.

include/ocpp/v2/profile.hpp
maaikez: What is the advantage of making it optional?

I asked Piet about his opinion, he made the first version of this and I don't know why he did not use optional (I just continued on what there was).

lib/ocpp/v2/profile.cpp
maaikez: The setpoints are set with the functions get_min_limit and get_max_limit (which are called in this function).
And if a default was set, it's not used in the end.

include/ocpp/common/constants.hpp
mlitre: I thought for discharge it could only be positive since we already know that it is discharge?
Yes for setpoint that is not an issue.

include/ocpp/v2/profile.hpp
mlitre: None really except for coherence since we already use it for the phase information.

include/ocpp/common/constants.hpp
maaikez: No according to the spec, it is a negative number.

lib/ocpp/v2/profile.cpp
mlitre: Ah okay thanks for confirming

include/ocpp/common/constants.hpp
mlitre: Ok thanks for clarifying

mlitre (approved):

PR #1043, opened 2025-04-11
update OCPP 1.6 database migration to support earlier configuration

james-ctc:
## Describe your changes
Database migration was added in April 2024, however some databases that predate that can be successfully migrated.
This changes the initial upgrade script so that some earlier databases can be successfully migrated.

fix: database creation/migration scripts updated to support
      migration from earlier database versions (OCPP1.6)

Note: PRAGMA foreign_keys = ON; removed
The sqlite documentation states that it isn't used when in a transaction: 
[sqlite documentation](https://sqlite.org/pragma.html#pragma_foreign_keys )

> This pragma is a no-op within a transaction; foreign key constraint
> enforcement may only be enabled or disabled when there is no
> pending BEGIN or SAVEPOINT.

Database migration/creation is done within a BEGIN TRANSACTION.

PRAGMA foreign_keys = ON was removed because it gives the misleading impression that the feature is permanently enabled. Sqlite further states that PRAGMA foreign_keys = ON has to be enabled per database connection.


marcemmers:
During the initial implementation it was a conscious choice not to allow migrations from previous databases. The reason we had was that this whole system of migrations is built on the assumption that the exact sequence is followed.

If you apply the first migration on a database that already has table X it will be skipped. But if that table X is missing a field Y, this could lead to issues in the future when field Y would be transformed into Z because it does not exist.

We reasoned the safest solution was to not allow this migration and it could be done by hand if it was really needed. I can see what the use would be of allowing this but want you to consider this potential issue.

james-ctc:
The specific case I'm dealing with is devices on 2024-03 which was only just before database migration was added and has the same table structure.

marcemmers:
That still means that this change allows to also work on way older databases, there is no way to check anymore.

james-ctc:
If update were separate from create then the an update initial script could do the work.
Alternatively there could be code added to check for db existence, version 0 and run a specific one-off script.
The solution needs to be self contained within libocpp rather than relying on something else.
Ideally there needs to be a process difference between updates and creating a new database.

hikinggrass:
> Alternatively there could be code added to check for db existence, version 0 and run a specific one-off script. The solution needs to be self contained within libocpp rather than relying on something else. 

That sounds like a doable approach, maybe a custom `0_initialize_existing_db_to_version_1[...]` file that is not shipped by libocpp but can be added if it is required

Pietfried:
Closing this PR for now. We may come up with a seperate solution with an attempt to migrate existing databases by providing a custom one-off script to libocpp.

PR #1052, opened 2025-04-27
Fix handling and reporting of more invalid messages

hikinggrass:
## Describe your changes
Add missing catch of StringConversionException and report CALLERROR
    
Report CALLERROR when the message ID cannot be extracted from RPC message
This can happen when RPC messages are malformed
    
Respond with MessageTypeNotSupported CALLERROR when message type is unknown

Bump version to 0.26.2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.14)
 
 project(ocpp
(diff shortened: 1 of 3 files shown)


Pietfried (approved):
As a next step we should try to unify the exception handling between 1.6 and 2.0.1 a bit better

PR #1054, opened 2025-05-01
[OCPP 1.6] periodically check for expired profiles across all profile types

james-ctc:
## Describe your changes

OCPP 1.6 periodically checks for expired profiles
There is a periodic check where expired charging profiles are removed. 
Originally this was for charge point max profiles only. 

Now tx and tx default profiles are also checked.

Unit test created to ensure correct operation.

Note the original implementation has a subtle bug 
```
            this->stack_level_charge_point_max_profiles_map.erase(it++);
            this->database_handler->delete_charging_profile(it->second.chargingProfileId);
```
`delete_charging_profile()` is called with the profile ID of the next element since the iterator was incremented on the previous line. The recommended form for use of erase is `it = map.erase(it);` since erase() returns the updated iterator.

Expired profiles were not being removed from the database except on startup.

### DIFF
--- a/lib/ocpp/v16/smart_charging.cpp
+++ b/lib/ocpp/v16/smart_charging.cpp
@@ -4,6 +4,7 @@
 #include "ocpp/common/constants.hpp"
 #include "ocpp/common/types.hpp"
 #include "ocpp/v16/ocpp_enums.hpp"
+#include <chrono>
 #include <ocpp/v16/profile.hpp>
 #include <ocpp/v16/smart_charging.hpp>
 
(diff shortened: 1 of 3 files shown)


lib/ocpp/v16/smart_charging.cpp
```diff
@@ -11,6 +11,38 @@ using namespace std::chrono;
...
+        if (it->second.validTo) {
+            const auto validTo = it->second.validTo.value().to_time_point();
+            remove = validTo < now;
+        }
+
+        if (remove) {
+            // remove expired profile from the database and map
+            // the order of these matters!
+            db.delete_charging_profile(it->second.chargingProfileId);
+            it = map.erase(it);
```
TMilburnPodPoint, line 38: This is very confusing to me, better to increment `it`
like `for (auto it = map.cbegin(); it != map.cend(); it++)`

james-ctc: The pattern comes from here: https://en.cppreference.com/w/cpp/container/map/erase
When using erase() `References and iterators to the erased elements are invalidated.`
The return from erase() is `iterator following the last removed element` allowing the loop to continue.

e.g. 
```
for (auto it = map.cbegin(); it != map.cend(); it++) {
  if (it->first % 2 == 0) {
    map.erase(it);
    // "it" is now invalid and hence "it++" is invalid breaking the loop
  }
}
```

```
for (auto it = map.cbegin(); it != map.cend(); it++) {
  if (it->first % 2 == 0) {
    it = map.erase(it);
    // "it" is valid and points to the next item
    // the ++ in the loop means this next item is skipped
  }
}
```

TMilburnPodPoint: Ah yes I'd forgotten C++ has silly things like this
This is clearer to me
```
for (auto it = map.cbegin(); it != map.cend(); it = next_it) {
{
  ++next_it;
  if (remove)
  {
    map.erase(it);
  }
}
```

mlitre:
We just need to fix the DCO check and then the PR is good for me!

PR #1093, opened 2025-06-03
refactor(OCPP2.x): Adjustments to device model

Pietfried:
## Describe your changes
* Changed type of MessageTypesDiscardForQueueing since OCTT base report test case requires a valuesList property, but providing all MessageTypes exceeds the permitted length
* Connector and EVSE components are not installed by default anymore
* Added ISO15118EvseId as standardized variable

Companion PR in everest-core: https://github.com/EVerest/everest-core/pull/1194


config/v2/component_config/custom/Connector_1_1.json
```diff
@@ -1,3 +1,4 @@
+
```
Pietfried, line 1:
```suggestion
```

config/v2/component_config/custom/EVSE_1.json
```diff
@@ -1,3 +1,4 @@
+
```
Pietfried, line 1:
```suggestion
```

config/v2/component_config/custom/EVSE_2.json
```diff
@@ -1,3 +1,4 @@
+
```
Pietfried, line 1:
```suggestion
```

config/v2/component_config/standardized/OCPPCommCtrlr.json
```diff
@@ -213,12 +213,13 @@
       "variable_name": "MessageTypesDiscardForQueueing",
       "characteristics": {
         "supportsMonitoring": true,
-        "dataType": "SequenceList"
+        "dataType": "string"
```
hikinggrass, line 216: Unfortunate that this is necessary, there might be a slightly "hacky" workaround. According to the spec you're allowed to set a `valuesList` of a `SequenceList` to 
`* SequenceList: The (Actual) Variable value may be an
ordered (priority, etc) (sub-)set of the reported (CSV) valid
values.`
So since this variable is readonly we could just set the actual `value` and `valuesList` to the same, as long as `value` is not longer than 1000 characters...

hikinggrass (changes requested):

config/v2/component_config/standardized/OCPPCommCtrlr.json
Pietfried: It is unfortunate... The `valuesList`

> specifies the allowed values for the type.

If we set it to the same value, the CSMS is not really aware that it could include other message types as well. In addition, the device model interface currently doesnt allow to change the characteristics at runtime, so I would like to keep the `string` type here...

PR #1099, opened 2025-06-18
fix(v2): eventNotificationType made optional NotifyMonitoringReport

mlitre:
The code generator was amended to have an exception for eventNotificationType in NotifyMonitoringReportRequest. This was marked as new and required in OCPP 2.1, but breaks schema validation in OCPP 2.0.1 so the presense of this value will be checked in code and not as type definition to avoid issues.

### DIFF
--- a/include/ocpp/v2/ocpp_types.hpp
+++ b/include/ocpp/v2/ocpp_types.hpp
@@ -1424,7 +1424,7 @@ struct VariableMonitoring {
     float value;
     MonitorEnum type;
     int32_t severity;
-    EventNotificationEnum eventNotificationType;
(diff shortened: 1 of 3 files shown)


mlitre:
@Pietfried Update to the code generator is done, but I was wondering what was the best way forward for handling the `eventNotificationType` in code? For OCPP 2.0.1 we can just reset the optional to be sure it is not sent, but when it is missing in OCPP 2.1 should we throw an exception? skip sending the message? but then if the CSMS asked for the report they would wait indefinitely.

hikinggrass (approved):

PR #1107, opened 2025-07-11
Fix json exceptions being thrown during handling of invalid messages

### LINKED ISSUE #1104: Invalid JSON in IPC message causes uncaught exception and crash
**Describe the bug**
When EVerest is fed fuzzed output from the CSMS, after a few invalid messages the ocpp module crashes, as it receives a malformed JSON message through IPC, which causes an exception in the nlohmann::json library. As this exception is not caught, the ocpp module crashes and triggers the termination of the rest of the stack.

Exception occurs in [rpc.cpp] and more specifically in the json library call at [ipc.cpp] in the `everest-framework` repo.

This was encountered while fuzzing EVerest with our proprietary fuzzer as the CSMS.

Note that I have managed to create some coredump files with GDB right after the exception. I can share them if desired. The error is present regardless of compiling with debug symbols.

[rpc.cpp]: https://github.com/EVerest/everest-framework/blob/3826ca864480b17026ade643ba4883033e30668a/src/controller/rpc.cpp#L104 

[ipc.cpp]: https://github.com/EVerest/everest-framework/blob/3826ca864480b17026ade643ba4883033e30668a/src/controller/ipc.cpp#L59

**Runtime Information**
- Running with workspace obtained from `edm init main ... ` with `gcc 14.2.1`, running with script `/run-sil-ocpp201-pnc.sh`
- everest-mqtt and everest-nodered running in docker with host networking
- CentOS Stream 10 (VM)

Affiliation: Discovery Partners Institute, University of Illinois Chicago, under Professor Rigel Gjomemo

Image

hikinggrass:
When certain malformed messages that during parsing result in a json exceptions are reported to the CSMS it was observed that secondary parsing exceptions can occur when the error message is sent to the CSMS as part of a CallError as well as when the incoming message in re-serialized in the security event to be sent.

This fixes #1104 by catching the secondary exception thrown during error message generation, and by not sending a truncated version of the problematic message but the error message instead.

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -1302,7 +1302,15 @@ void ChargePointImpl::message_callback(const std::string& message) {
         return;
     } catch (const json::exception& e) {
         EVLOG_error << "JSON exception during reception of message: " << e.what();
-        this->message_dispatcher->dispatch_call_error(CallError(MessageId("-1"), "GenericError", e.what(), json({})));
+        std::string error_message;
+        try {
+            error_message = json(e.what()).dump();
+        } catch (const json::exception& ex) {
+            error_message = "JSON exception during reception of message: ";
+            error_message += ex.what();
+        }
+        this->message_dispatcher->dispatch_call_error(
+            CallError(MessageId("-1"), "GenericError", error_message, json({})));
         return;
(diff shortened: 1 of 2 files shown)


Pietfried (changes requested):
lgtm for ocpp2, it looks like we could have similar issues for ocpp1.6. Can you apply these changes to ocpp1.6, too?

PR #1112, opened 2025-07-22
Fix: Composite schedule calculation for connector 0

Pietfried:
## Describe your changes
Changed composite schedule calculation for connector 0:
* When generating the EVSE intermediate profiles, the session start time was not calculated and passed to the generate_evse_intermediates function. This caused an invalid calculation of the IntermediateProfile for EVSEs with an active transaction, because the session_start time was not considered
* For composite schedule for connector 0, the summing limits were incorrectly included, even when no limits for setpoints or discharge limits had been set by the CSMS. This has been fixed by excluding setpoint calculation for connector 0 in general.
* A limit of 0A was not applied for composite schedules for connector 0. This has been fixed by changing a `> 0` to a `>= 0` 


mlitre (commented):
This looks good to me, it's probably worth adding a test or two for the future.

hikinggrass (approved):

hikinggrass (approved):

PR #1114, opened 2025-07-23
fix(v21): Fix V2XChargingCtrlr variables

mlitre:
## Describe your changes

V2X variables were set globally when they were supposed to be evse specific.

### DIFF
--- a/lib/ocpp/v2/charge_point.cpp
+++ b/lib/ocpp/v2/charge_point.cpp
@@ -69,7 +69,7 @@ ChargePoint::ChargePoint(const std::map<int32_t, int32_t>& evse_connector_struct
     }
 
     // Make sure the received callback struct is completely filled early before we actually start running
-    if (!this->callbacks.all_callbacks_valid(this->device_model)) {
+    if (!this->callbacks.all_callbacks_valid(this->device_model, evse_connector_structure)) {
         EVLOG_AND_THROW(std::invalid_argument("All non-optional callbacks must be supplied"));
     }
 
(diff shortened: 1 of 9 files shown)


lib/ocpp/v2/charge_point.cpp
```diff
@@ -589,16 +589,19 @@ void ChargePoint::initialize(const std::map<int32_t, int32_t>& evse_connector_st
         this->callbacks.stop_transaction_callback, this->registration_status, this->upload_log_status,
         this->upload_log_status_id);
 
-    if (device_model->get_optional_value<bool>(ControllerComponentVariables::V2XChargingCtrlrAvailable)
-            .value_or(false)) {
+    const auto v2x_available_map =
+        device_model->get_all_optional_values_for_each_evse<bool>("V2XChargingCtrlr", V2xComponentVariables::Available);
```
Pietfried, line 593: To me it lools like you dont need `get_all_optional_values_for_each_evse` and could use something like this instead:

```cpp
bool any_v2x_available = std::any_of(
    evse_connector_structure.begin(),
    evse_connector_structure.end(),
    [this](const auto& entry) {
        const auto& [evse, connectors] = entry;
        return this->device_model
            ->get_optional_value<bool>(
                V2xComponentVariables::get_component_variable(evse, V2xComponentVariables::Available)
            )
            .value_or(false);
    }
);
``

File: lib/ocpp/v2/charge_point_callbacks.cpp [hunk 2]
```diff
@@ -87,8 +87,12 @@ bool Callbacks::all_callbacks_valid(std::shared_ptr<DeviceModel> device_model) c
...
         }
 
-        if (device_model->get_optional_value<bool>(ControllerComponentVariables::V2XChargingCtrlrAvailable)
-                .value_or(false) and
+        const auto v2x_available_map = device_model->get_all_optional_values_for_each_evse<bool>(
+            "V2XChargingCtrlr", V2xComponentVariables::Available);
+        const bool v2x_available = !v2x_available_map.empty() and
+                                   std::find_if(v2x_available_map.cbegin(), v2x_available_map.cend(),
+                                                [](const auto& it) { return it.second; }) != v2x_available_map.cend();
+        if (v2x_available and
```
Pietfried, line 95: same here, you might need to pass the `evse_connector_map` then as well

Pietfried (approved):
lgtm, see two comments in line, leaving it up to you to apply

PR #1148, opened 2025-10-17
Set ContractCertificateInstallationEnabled in example config

Pietfried:
## Describe your changes
Set ContractCertificateInstallationEnabled to true. This configuration is required to work with the EVerest Plug&Charge tutorial


Pietfried:
> It would be worth checking other variables that might be missing a default value

True, although we don't claim anywhere there is much to expect from the example device model. For now I just want it to work with the Plug&Charge tutorial we have

PR #1150, opened 2025-11-05
Extend v201 supported file

MatthiasContribute:
## Describe your changes
Add more information to the ocpp v201 status file, to extend the readability. The column requirement definition is added.


Pietfried:
I see, but it is supposed to be used alongside with the specification. We would need extra columns for **Precondition** and **Note** as well and I think this is just too much.

PR #1153, opened 2025-11-18
[OCPP 1.6] Add support for updating OCPP config via method calls

james-ctc:
## Describe your changes

In OCPP 1.6 previously only custom configuration could be updated via method calls.
This updated allows method calls to update OCPP configuration with the same processing that would occur had the request been via OCPP.

(similar to https://github.com/EVerest/libocpp/pull/1152 but for main)

### DIFF
--- a/include/ocpp/v16/charge_point_impl.hpp
+++ b/include/ocpp/v16/charge_point_impl.hpp
@@ -392,6 +392,16 @@ private:
     /// \param connector for which availability change shall be checked and executed
     void execute_queued_availability_change(const int32_t connector);
 
+    /// \brief Sets a configuration key (internal implementation)
+    /// \param key
+    /// \param value
+    /// \param uniqueId used when an OCPP response is sent
+    /// \return Indicates the result of the operation with an optional response message
+    /// \note the optional response message will be nullopt when a response has
+    ///       been sent by this method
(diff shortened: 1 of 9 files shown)


include/ocpp/v16/charge_point_impl.hpp
```diff
@@ -921,11 +921,13 @@ class ChargePointImpl : ocpp::ChargingStationBase {
...
     GetConfigurationResponse get_configuration_key(const GetConfigurationRequest& request);
 
-    /// \brief Sets a custom configuration key
+    /// \brief Sets a configuration key
     /// \param key
     /// \param value
     /// \return Indicates the result of the operation
-    ConfigurationStatus set_custom_configuration_key(CiString<50> key, CiString<500> value);
+    ConfigurationStatus set_configuration_key(CiString<50> key, CiString<500> value);
+    std::pair<ConfigurationStatus, std::optional<ChangeConfigurationResponse>>
```
Pietfried: looks like this function can be private

Pietfried (changes requested):

include/ocpp/v16/charge_point_impl.hpp
james-ctc: updated to make set_configuration_key_internal private

Pietfried (approved):

PR #1156, opened 2025-11-26
[OCPP 1.6] InvalidCentralSystemCertificate being logged too frequently

james-ctc:
## Describe your changes

fix: InvalidCentralSystemCertificate was logged on every failed connection

A CSMS certificate may be considered invalid when a charger has the incorrect time. This can be recovered by using NTP. However a security event is raised for each failed connection attempt until the correct time is set.

This change ensures that InvalidCentralSystemCertificate is logged only on the first occurance.

Replaces PR https://github.com/EVerest/libocpp/pull/1155

### DIFF
--- a/lib/ocpp/v16/charge_point_impl.cpp
+++ b/lib/ocpp/v16/charge_point_impl.cpp
@@ -44,6 +44,7 @@ ChargePointImpl::ChargePointImpl(const std::string& config, const fs::path& shar
     ocpp::ChargingStationBase(evse_security, security_configuration),
     bootreason(BootReasonEnum::PowerUp),
     initialized(false),
+    InvalidCSMSCertificate_logged(false),
     connection_state(ChargePointConnectionState::Disconnected),
     registration_status(RegistrationStatus::Pending),
     diagnostics_status(DiagnosticsStatus::Idle),
@@ -298,6 +299,10 @@ void ChargePointImpl::init_websocket() {
         this->message_queue->resume(this->message_queue_resume_delay);
         this->connected_callback();
 
+        // There has been a successful connection so a subsequent
+        // InvalidCSMSCertificate should be logged
+        InvalidCSMSCertificate_logged = false;
+
         // signal_set_charging_profiles_callback since composite schedule could have changed if
         // IgnoredProfilePurposesOffline are configured when becoming online
         if (this->signal_set_charging_profiles_callback != nullptr and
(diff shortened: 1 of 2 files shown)


lib/ocpp/v16/charge_point_impl.cpp
```diff
@@ -341,14 +346,20 @@ void ChargePointImpl::init_websocket() {
...
-            // This event is forced to accomodate for for TC_078_CS even if this event is not critical
-            this->securityEventNotification(CiString<50>(ocpp::security_events::INVALIDCENTRALSYSTEMCERTIFICATE),
-                                            std::nullopt, true, true);
+            if (InvalidCSMSCertificate_logged) {
+                EVLOG_warning << "Connection failed: InvalidCSMSCertificate";
+            } else {
+                // This event is forced to accommodate TC_078_CS despite this event being not critical
+                this->securityEventNotification(CiString<50>(ocpp::security_events::INVALIDCENTRALSYSTEMCERTIFICATE),
+                                                std::nullopt, true, true);
+                InvalidCSMSCertificate_logged = true;
```
mlitre, line 355: This is probably good enough, but I think it would be best if maybe we used the tech info to log what endpoint it tried and compare that if tries to log again. This would make it so that we would log only once per endpoint attempt (in case there is a change in endpoint attempts).

james-ctc: For OCPP 1.6 the endpoint is in  JSON file(s) and without the CSMS I'm not sure anything else can change it.
(perhaps the stable APIs?)
I did consider logging based on the returned certificate changing but that would require quite a few changes to the webscokets handling and interfaces to make the certificate information available.
Given the OCPP security white paper is clear that sending to the CSMS is not critical I didn't consider that level of change was justified.

mlitre: I see that makes sense, could we maybe log in the tech info the endpoint, could always be useful for a CSMS. Except for that, I think that's fine

james-ctc: I can see a case for potentially providing more information - but that information is hard to get at.
Fundamentally the useful information is why did the connection fail. The URI used (especially since it doesn't change) is not so useful. If OCPP 1.6 had multiple connection URIs then knowing which one failed could be useful.
With this failure it is a TLS setup problem: invalid CSMS certifiate, no longer valid, no verification path …
Possible causes: incorrect DNS, charger has the wrong time, wrong root certificate (in which case how does it successfully connect?). The URI doesn't help to fault find (and if it is wrong there would never be a successful connection to be able to send the OCPP message)
The information about the problem is in the charger logs (along with the connection URI).
Finally - since the original code doesn't log this information then changing the OCPP message should be a separate PR

mlitre (approved):

PR #1158, opened 2025-11-27
fix: Correct port when scheme is not provided

chrizap:
provided

Signed-off-by: Christofer Gilje Skjæveland <christofer.skjaeveland@zaptec.com>

## Describe your changes
This change ensures that URLs without an explicit scheme are handled correctly for security profiles 2 and 3.

Previously, when a URL like **test.uri.com/path/** was provided under security profile 2, it would incorrectly resolve to **wss://test.uri.com:80/path/cp0001**. With this fix, it now falls back to the correct default secure port, resulting in **wss://test.uri.com:443/path/cp0001**.


Pietfried (approved):
lgtm, thanks @chrizap

PR #1161, opened 2025-12-02
Added configuration key MeterPublicKey for OCPP1.6

Pietfried:
## Describe your changes
* Add support for the MeterPublicKey configuration key
* Added public function to set the public key of the powermeter per connector

The implementation for this configuration key follows the requirements of the [OCA Eichrecht Whitepaper](https://openchargealliance.org/ocpp-info-whitepapers/signed-meter-values-eichrecht-paper/)

The format the new configuration keys are reported is:
**MeterPublicKey[n]** where n is the id of the respective connector

Companion PR: https://github.com/EVerest/everest-core/pull/1599

### DIFF
--- a/lib/ocpp/v16/charge_point_configuration.cpp
+++ b/lib/ocpp/v16/charge_point_configuration.cpp
@@ -3172,6 +3172,79 @@ std::optional<KeyValue> ChargePointConfiguration::getWaitForSetUserPriceTimeoutK
     return result;
 }
 
+std::optional<KeyValue> ChargePointConfiguration::getPublicKeyKeyValue(const uint32_t connector_id) {
+    if (!this->config["Internal"].contains("MeterPublicKeys")) {
+        return std::nullopt;
+    }
+
+    const auto& meter_public_keys = this->config["Internal"].at("MeterPublicKeys");
+
+    if (!meter_public_keys.is_array()) {
+        return std::nullopt;
+    }
+
+    const auto& keys_array = meter_public_keys;
+    if (keys_array.size() < connector_id or connector_id < 1) {
+        return std::nullopt;
+    }
+
+    KeyValue kv;
+    kv.key = "MeterPublicKey[" + std::to_string(connector_id) + "]";
+    kv.readonly = true;
+    kv.value = keys_array.at(connector_id - 1).get<std::string>();
+    return kv;
+}
+
+std::optional<std::vector<KeyValue>> ChargePointConfiguration::getAllMeterPublicKeyKeyValues() {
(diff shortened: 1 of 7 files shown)


lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -3254,6 +3327,53 @@ void ChargePointConfiguration::setCentralSystemURI(std::string centralSystemUri)
     this->setInUserConfig("Internal", "CentralSystemURI", centralSystemUri);
 }
 
+namespace {
+template <typename T> std::optional<T> parse_meter_public_key_index(const std::string& input) {
```
james-ctc: is there any situation where parse_meter_public_key_index will be called where T != uint32?

lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -3502,6 +3622,24 @@ std::optional<KeyValue> ChargePointConfiguration::get(CiString<50> key) {
     if (key == "WebSocketPingInterval") {
         return this->getWebsocketPingIntervalKeyValue();
     }
+    if (key.get().rfind("MeterPublicKey[", 0) == 0) {
+        const std::string& s = key.get();
+        const auto connector_id_opt = parse_meter_public_key_index<uint32_t>(s);
+
+        if (connector_id_opt.has_value() == false) {
```
james-ctc: in other places `if (!…)` is used, which is preferable, rather than `if (… == false)`

lib/ocpp/v16/charge_point_configuration.cpp
```diff
@@ -3502,6 +3622,24 @@ std::optional<KeyValue> ChargePointConfiguration::get(CiString<50> key) {
...
+            EVLOG_error << "Invalid MeterPublicKey format for key '" << s << "'";
+            return std::nullopt;
+        }
+
+        const uint32_t connector_id = connector_id_opt.value();
```
james-ctc: safer to be `const auto` especially since `connector_id_opt` is auto.

james-ctc (commented):
**General**

especially for header files basic types as function arguments shouldn't be const.
e.g. `std::uint32_t connector_id` rather than `const std::uint32_t connector_id`

(it does have a meaning for the function implementation)

lib/ocpp/v16/charge_point_configuration.cpp
Pietfried: Removed the template

james-ctc (approved):

hikinggrass (approved):

PR #1171, opened 2026-01-20
fix: fetch readonly flag for AuthorizeRemoteTxRequests from Schema

mhungerford-cpi:
The OCPP 1.6 specs mention that the configuration key AuthorizeRemoteTxRequests can be ReadWrite or ReadOnly based on ChargingStation implementation. For our use case, we want it to be ReadOnly. With this fix the readonly flag for the concerned configuration key is fetched from Core schema profile.


mhungerford-cpi:
meant for our fork

